Issue #72, opened 2017-05-18
Fix frame extraction failure logic to mark specific frames


jrobble:
On two separate videos, we are getting the following errors:
“WfmProcessingException: Failed to extract all of the artifacts from Media #A! 4 != 5”
“WfmProcessingException: Failed to extract all of the artifacts from Media #B! 9 != 10”
 
In the detection.json, the detection job is completing without errors but all of the exemplars have an artifactExtractionStatus of FAILED. There are 5 exemplars in job A, and 10 in job B. Looking in the “/opt/mpf/share/artifacts/job..” folder all but 1 exemplar frame is present – 4 in job A, 9 in job B. In both cases it is the last exemplar that has a missing frame – yet all exemplar extractions are failing.

Looking at the code, it appears that the frame extractor attempts to extract all of the desired frames for a piece of media, and then the WFM does a post-check to make sure the proper number of frames were extracted. If that check fails then all the extractions are marked as failed for that piece of media.

We shouldn’t mark them all as failed if only one fails. We need to improve that logic.

PR #76, opened 2017-05-22
Fixed the way that extraction errors are reported.

jrobble:
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/76)


jrobble:
Fixes #72.

PR #221, opened 2017-08-18
Separated out dependencies.json.

jrobble:
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/221)


brosenberg42:
Reviewed 4 of 4 files at r1.

---

*[trunk/workflow-manager/src/main/webapp/resources/mytheme/js/controllers/AboutCtrl.js, line 30 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/221#-Ks5Mcf7iWC-8GJKZEqi:-Ks5Mcf7iWC-8GJKZEqj:bgnnj1b) ([raw file](https://github.com/openmpf/openmpf/blob/9c71871630bed3f410b034fdeee860cfe96fef5f/trunk/workflow-manager/src/main/webapp/resources/mytheme/js/controllers/AboutCtrl.js#L30)):*
> ```JavaScript
> /* global $, angular */
> 
> var AboutCtrl = function($scope, $location, $interpolate, depPromise, MetadataService) {
> ```

depPromise is not a promise, it is an http response object. The promise returned by `$http.get("resources/json/dependencies.json")` has already been resolved and unpacked. This function never has access  to the original promise.

---

*[trunk/workflow-manager/src/main/webapp/resources/mytheme/js/controllers/AboutCtrl.js, line 47 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/221#-Ks5MloECJIUJ9P-Yyky:-Ks5MloECJIUJ9P-Yykz:bvgtcxl) ([raw file](https://github.com/openmpf/openmpf/blob/9c71871630bed3f410b034fdeee860cfe96fef5f/trunk/workflow-manager/src/main/webapp/resources/mytheme/js/controllers/AboutCtrl.js#L47)):*
> ```JavaScript
> 
>     var loadDependencyGroups = function () {
>         $scope.depGroups = angular.fromJson(depPromise.data).groups;
> ```

depPromise.data is already a JavaScript object, you don't need to call `angular.fromJSON()`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/221)*

jrobble:

---

*[trunk/workflow-manager/src/main/webapp/resources/mytheme/js/controllers/AboutCtrl.js, line 30 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/221#-Ks5Mcf7iWC-8GJKZEqi:-Ks9jg7bSTT6PnHf7o4x:b-b6mhik) ([raw file](https://github.com/openmpf/openmpf/blob/9c71871630bed3f410b034fdeee860cfe96fef5f/trunk/workflow-manager/src/main/webapp/resources/mytheme/js/controllers/AboutCtrl.js#L30)):*

depPromise is not a promise, it is an http response object. The promise returned by `$http.get("resources/json/dependencies.json")` has already been resolved and unpacked. This function never has access  to the original promise.
</blockquote></details>

I based my code on the following example from https://github.com/angular-ui/ui-router/wiki:

```
promiseObj:  function($http){
            // $http returns a promise for the url data
            return $http({method: 'GET', url: '/someUrl'});
         },
```

Should I call it `depPromiseObj`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/221)*

brosenberg42:

---

*[trunk/workflow-manager/src/main/webapp/resources/mytheme/js/controllers/AboutCtrl.js, line 30 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/221#-Ks5Mcf7iWC-8GJKZEqi:-Ks9m_OwMjbJwvRb93ek:bsz2f68) ([raw file](https://github.com/openmpf/openmpf/blob/9c71871630bed3f410b034fdeee860cfe96fef5f/trunk/workflow-manager/src/main/webapp/resources/mytheme/js/controllers/AboutCtrl.js#L30)):*

I based my code on the following example from https://github.com/angular-ui/ui-router/wiki:

```
promiseObj:  function($http){
            // $http returns a promise for the url data
            return $http({method: 'GET', url: '/someUrl'});
         },
```

Should I call it `depPromiseObj`?
</blockquote></details>

No, that doesn't help. There is a difference between a promise and the value it resolves to. 
When I see a variable with promise in the name, that implies that it is a promise and I have to call `promise.then(function (result) {})` to get the http response. The value that is injected does not have a `.then()` method and is not a promise. 
It should be named as if you were just passing in a regular object.
There are at least two options:

1. You can call it something like `dependenciesResponse`.

or 

2. Change resolve to look like:
```
resolve: {
    dependencies: function($http) {
        return $http.get("resources/json/dependencies.json")
                .then(function(resp) {
                    return resp.data;
                });
    }
}
```
and then you don't access `.data` since you are now injecting the content of the http response (rather than the response itself) in to the controller:
`$scope.depGroups = dependencies.groups;`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/221)*

jrobble:

---

*[trunk/workflow-manager/src/main/webapp/resources/mytheme/js/controllers/AboutCtrl.js, line 30 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/221#-Ks5Mcf7iWC-8GJKZEqi:-KsA17pf88bEXHrUw5pQ:bt80l1n) ([raw file](https://github.com/openmpf/openmpf/blob/9c71871630bed3f410b034fdeee860cfe96fef5f/trunk/workflow-manager/src/main/webapp/resources/mytheme/js/controllers/AboutCtrl.js#L30)):*

No, that doesn't help. There is a difference between a promise and the value it resolves to. 
When I see a variable with promise in the name, that implies that it is a promise and I have to call `promise.then(function (result) {})` to get the http response. The value that is injected does not have a `.then()` method and is not a promise. 
It should be named as if you were just passing in a regular object.
There are at least two options:

1. You can call it something like `dependenciesResponse`.

or 

2. Change resolve to look like:
```
resolve: {
    dependencies: function($http) {
        return $http.get("resources/json/dependencies.json")
                .then(function(resp) {
                    return resp.data;
                });
    }
}
```
and then you don't access `.data` since you are now injecting the content of the http response (rather than the response itself) in to the controller:
`$scope.depGroups = dependencies.groups;`


</blockquote></details>

Looks like the example could use some correction. I'm going with your # 1.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/221)*

brosenberg42:
:lgtm:

---

Reviewed 2 of 2 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/221#-:-KsEB9_7w8zs-tivyovz:bnfp4nl)*

Issue #215, opened 2017-08-14
Update TimeUtils.java to support feed-forward batch processing


jrobble:
Update TimeUtils.java to support generating segments when feed-forward is enabled. Currently, `TimeUtils.createSegments()` does not take into consideration the value of FEED_FOWARD_TYPE nor FEED_FORWARD_TOP_CONFIDENCE_COUNT. Either update this method or create another method.

When generating feed-forward segments, chose the segment start frame and stop frame such that the greatest number of detections with the highest confidence (<= FEED_FORWARD_TOP_CONFIDENCE_COUNT) are captured within each segment. In some cases, it may make sense to generate segments where the value of (start frame - stop frame) exceeds the value of the segment size property if the top-confidence detections in that segment are spaced out. This is because only the frames associated with top-confidence detections will be processed, so only those frames count against the desired segment size.

Add unit tests to TestSegmenting.java to test the TimeUtils updates.

<br>**UPDATE 9/1/17:**

**When feed-forward is enabled, we should generate one segment for the whole feed-forward track.**

dpierce17:
While researching this task, found that the implementation of TimePair Comparable (method compareTo) needs to be fixed.  Added that fix to this task along with correcting TimePair hashCode method to use Objects.hash()

dpierce17:
While working this task, found a bug in DetectionSplitter:createTimePairsForTracks.  In this method, a 1 was added to track.getEndOffsetFrameInclusive() when forming the TimePair.  Research shows that track endOffsetFrame used to be exclusive (hence the +1), but the +1 is no longer required now that endOffsetFrame is inclusive.  Fixing this bug as part of this task.

dpierce17:
Also found that the Comparable for Detection was not fully implemented, only check was for other Detection equal to null (if so returned the other Detection as greater than this Detection), or for other Detection equal to this Detection.  Keeping original assumption that a null Detection is after this Detection (i.e. maybe the original intent was to mark end of data set?), and adding correct less than and greater then checks.

PR #228, opened 2017-08-24
Updated java.yml to use JDK and JRE 1.8.0_144.

jrobble:
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/228)

Issue #176, opened 2017-07-20
Investigate how the WFM will create sub-jobs for feed-forward batch processing


jrobble:
Let's define a sub-job as an MPFImageJob, MPFVideoJob, or MPFAudioJob. When the WFM gets a request to create a new batch job through the REST API, it creates Camel routes and starts the process of executing that job. An important part of the process involves splitting the job in sub-jobs. Each sub-job is passed to component services via a component-specific job request queue.

Generally, sub-jobs are created for two reasons:

1. There is at least one sub-job for piece of media. For example, there will be one MPFImageJob for every piece of image media in the batch job request.

2. There is one sub-job for each segment in a video. For example, for a 10,000 frame video the WFM will generate 10 MPFVideoJobs if the segment size is set to 1,000 frames.

The WFM is also responsible for discarding small tracks (configurable: MIN_TRACK_LENGTH) and merging tracks (configurable: MERGE_TRACKS) after every stage of the pipeline. Two tracks are merged if if the gap (number of frames) between two merged tracks is small enough (configurable: MIN_GAP_BETWEEN_TRACKS) and the detection region overlap % is above some threshold (configurable: MIN_OVERLAP).

For each of these merged tracks, the WFM generates segments for the next stage by first sorting the merged tracks by time and then combining / splitting them based on MIN_GAP_BETWEEN_SEGMENTS, TARGET_SEGMENT_LENGTH, and MIN_SEGMENT_LENGTH. Refer to the "Segmenting Media" section in the [User Guide](https://openmpf.github.io/docs/site/User-Guide/) for more information.

For feed-forward processing, the WFM needs to ensure that each new segment corresponds to a merged track generated in the previous pipeline stage. This will require by-passing some of the segmenting logic discussed in the "Segmenting Media" section. If the segment is larger than the desired segment size, then it's acceptable for the segment to be split into two smaller segments. This still ensures that each new segment corresponds to one merged track.

Ultimately we want to ensure that the WFM only needs to pass one track (or part of one track) generated in the previous stage to the sub-jobs in the next stage. 

Note that if the "FEED_FORWARD_TYPE" job/algorithm/media property is not set (or set to "NONE"), then the WFM creates sub-jobs normally (current behavior). This new behavior only takes effect if the "FEED_FORWARD_TYPE" property is set to something else.

Assess the feasibility of this approach.

jrobble:
[...]

PR #237, opened 2017-08-30
Wfm feed forward

dpierce17:
Pull request for Update TimeUtils.java to support feed-forward batch processing #215

---
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/237)


dpierce17:
github-with-pull-request Build 132 completed successfully, so this pull request is ready for review.  One question, I'd like to confirm that it is my wfm_feed_forward branch that was used for github-with-pull-request Build 132, but I'm not finding mention of my branch anywhere within the output of github-with-pull-request Build 132.  Is the branch tested listed somewhere in the Build output where I can find it?

brosenberg42:
No, it does not mention the branch. At the beginning of the build it will mention your pull request. For example:
```
Attempting to retrieve pull request #237 for the "openmpf" repo...
Retrieved pull request: "Wfm feed forward" (https://github.com/openmpf/openmpf/pull/237)
The pull request SHA is: 93a00d011a317fca76f65d6f07cc66a711af1299
```

jrobble:
Reviewed 1 of 6 files at r1.
Review status: 1 of 6 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/DetectionSplitter.java, line 64 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/237#-KsoNu94SIP3pw-Az4Lb:-KsoNu94SIP3pw-Az4Lc:b-53a6zn) ([raw file](https://github.com/openmpf/openmpf/blob/93a00d011a317fca76f65d6f07cc66a711af1299/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/DetectionSplitter.java#L64)):*
> ```Java
>   protected TimeUtils timeUtils;
> 
>   @Autowired
> ```

Looks like left indentation is a bit off for this autowire and the one above it.

Please check your left indentation for your other additions as well. It seems that you may have the indent set to 2 spaces instead of 4.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/transients/Detection.java, line 33 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/237#-KsoOdExi1ZV9WPhnZWI:-KsoOdExi1ZV9WPhnZWJ:b-ijvpxh) ([raw file](https://github.com/openmpf/openmpf/blob/93a00d011a317fca76f65d6f07cc66a711af1299/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/transients/Detection.java#L33)):*
> ```Java
> import com.google.common.base.Joiner;
> import java.util.Comparator;
> import java.util.Comparator;
> ```

This is a duplicate.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/transients/Detection.java, line 168 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/237#-KsoRZ699zaZ0zknJ56-:-KsoRZ6ATOZkmbiJ28I7:b-ueqmtf) ([raw file](https://github.com/openmpf/openmpf/blob/93a00d011a317fca76f65d6f07cc66a711af1299/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/transients/Detection.java#L168)):*
> ```Java
> 				(comparison = compareMap(detectionProperties,other.detectionProperties)) == 0) {
> 			comparison = 0;
>     } else if ( this.mediaOffsetFrame < other.mediaOffsetFrame ) {
> ```

I believe the mediaOffsetFrame comparison is already handled by the original comparison logic:

```
} else if((comparison = Integer.compare(mediaOffsetFrame, other.mediaOffsetFrame)) == 0 &&
				(comparison = Integer.compare(mediaOffsetTime, other.mediaOffsetTime)) == 0 &&
				(comparison = Integer.compare(x, other.x)) == 0 &&
				(comparison = Integer.compare(y, other.y)) == 0 &&
				(comparison = Integer.compare(width, other.width)) == 0 &&
				(comparison = Integer.compare(height, other.height)) == 0 &&
				(comparison = Float.compare(confidence, other.confidence)) == 0 &&
				(comparison = TextUtils.nullSafeCompare(artifactPath, other.artifactPath)) == 0 &&
				(comparison = compareMap(detectionProperties,other.detectionProperties)) == 0) {
		comparison = 0;
}
return comparison;
```

Specifically, this line:

`(comparison = Integer.compare(mediaOffsetFrame, other.mediaOffsetFrame)) == 0 &&`

does a few things. First, it uses `Integer.compare()`, which will return a -1, 0, or 1, and sets `comparison` to that value. If the value is not 0 then the `==` part of the clause evaluates to false. `false && anything` is false, so the next clause: 

`(comparison = Integer.compare(mediaOffsetTime, other.mediaOffsetTime)) == 0 &&`

is not evaluated. The whole`else if(comparison = ..., comparison = ..., comparison = ...)` case essentially becomes `else if(false)` so when `return comparison;` is invoked at the end of `compareTo` it still has the value of the initial comparison between the mediaOffsetFrame values.

<br>Now consider when:

`(comparison = Integer.compare(mediaOffsetFrame, other.mediaOffsetFrame)) == 0 &&`

sets `comparison` to 0. The `==` part of the clause evaluates to true. `true && anything` is that anything, so the next clause is evaluated:

`(comparison = Integer.compare(mediaOffsetTime, other.mediaOffsetTime)) == 0 &&`

That compares the mediaOffsetTime and sets the value of `comparison` to -1, 0, or 1. If the value is 0 then the whole`else if(comparison = ..., comparison = ..., comparison = ...)` case essentially becomes `else if(false)` so when `return comparison;` is invoked at the end of `compareTo` it still has the value of the initial comparison between the mediaOffsetTime values.


<br> The logic here is a bit hard to parse at first, but it is very concise if you know what it's doing. The key is understanding how Java evaluates the big conditional from left to right and ["short circuits"](https://stackoverflow.com/questions/8759868/java-logical-operator-short-circuiting) if necessary.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/237)*

jrobble:
Review status: 1 of 6 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/transients/Detection.java, line 168 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/237#-KsoRZ699zaZ0zknJ56-:-KsoYhvXFouJznHbSO--:b-42sdun) ([raw file](https://github.com/openmpf/openmpf/blob/93a00d011a317fca76f65d6f07cc66a711af1299/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/transients/Detection.java#L168)):*

I believe the mediaOffsetFrame comparison is already handled by the original comparison logic:

```
} else if((comparison = Integer.compare(mediaOffsetFrame, other.mediaOffsetFrame)) == 0 &&
				(comparison = Integer.compare(mediaOffsetTime, other.mediaOffsetTime)) == 0 &&
				(comparison = Integer.compare(x, other.x)) == 0 &&
				(comparison = Integer.compare(y, other.y)) == 0 &&
				(comparison = Integer.compare(width, other.width)) == 0 &&
				(comparison = Integer.compare(height, other.height)) == 0 &&
				(comparison = Float.compare(confidence, other.confidence)) == 0 &&
				(comparison = TextUtils.nullSafeCompare(artifactPath, other.artifactPath)) == 0 &&
				(comparison = compareMap(detectionProperties,other.detectionProperties)) == 0) {
		comparison = 0;
}
return comparison;
```

Specifically, this line:

`(comparison = Integer.compare(mediaOffsetFrame, other.mediaOffsetFrame)) == 0 &&`

does a few things. First, it uses `Integer.compare()`, which will return a -1, 0, or 1, and sets `comparison` to that value. If the value is not 0 then the `==` part of the clause evaluates to false. `false && anything` is false, so the next clause: 

`(comparison = Integer.compare(mediaOffsetTime, other.mediaOffsetTime)) == 0 &&`

is not evaluated. The whole`else if(comparison = ..., comparison = ..., comparison = ...)` case essentially becomes `else if(false)` so when `return comparison;` is invoked at the end of `compareTo` it still has the value of the initial comparison between the mediaOffsetFrame values.

<br>Now consider when:

`(comparison = Integer.compare(mediaOffsetFrame, other.mediaOffsetFrame)) == 0 &&`

sets `comparison` to 0. The `==` part of the clause evaluates to true. `true && anything` is that anything, so the next clause is evaluated:

`(comparison = Integer.compare(mediaOffsetTime, other.mediaOffsetTime)) == 0 &&`

That compares the mediaOffsetTime and sets the value of `comparison` to -1, 0, or 1. If the value is 0 then the whole`else if(comparison = ..., comparison = ..., comparison = ...)` case essentially becomes `else if(false)` so when `return comparison;` is invoked at the end of `compareTo` it still has the value of the initial comparison between the mediaOffsetTime values.


<br> The logic here is a bit hard to parse at first, but it is very concise if you know what it's doing. The key is understanding how Java evaluates the big conditional from left to right and ["short circuits"](https://stackoverflow.com/questions/8759868/java-logical-operator-short-circuiting) if necessary.
</blockquote></details>

Here's the implementation of `Integer.compare()`:

```
    public static int compare(int x, int y) {
        return (x < y) ? -1 : ((x == y) ? 0 : 1);
    }
```

I referenced that with:

`Integer.compare(mediaOffsetTime, other.mediaOffsetTime)`

So if `this` is less than `other` return -1. I believe the conditionals you added do the same thing.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/237)*

brosenberg42:
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/transients/Detection.java, line 168 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/237#-KsoRZ699zaZ0zknJ56-:-KsoYTUUfG-bAjL-PHrn:b2h1nke) ([raw file](https://github.com/openmpf/openmpf/blob/93a00d011a317fca76f65d6f07cc66a711af1299/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/transients/Detection.java#L168)):*

I believe the mediaOffsetFrame comparison is already handled by the original comparison logic:

```
} else if((comparison = Integer.compare(mediaOffsetFrame, other.mediaOffsetFrame)) == 0 &&
				(comparison = Integer.compare(mediaOffsetTime, other.mediaOffsetTime)) == 0 &&
				(comparison = Integer.compare(x, other.x)) == 0 &&
				(comparison = Integer.compare(y, other.y)) == 0 &&
				(comparison = Integer.compare(width, other.width)) == 0 &&
				(comparison = Integer.compare(height, other.height)) == 0 &&
				(comparison = Float.compare(confidence, other.confidence)) == 0 &&
				(comparison = TextUtils.nullSafeCompare(artifactPath, other.artifactPath)) == 0 &&
				(comparison = compareMap(detectionProperties,other.detectionProperties)) == 0) {
		comparison = 0;
}
return comparison;
```

Specifically, this line:

`(comparison = Integer.compare(mediaOffsetFrame, other.mediaOffsetFrame)) == 0 &&`

does a few things. First, it uses `Integer.compare()`, which will return a -1, 0, or 1, and sets `comparison` to that value. If the value is not 0 then the `==` part of the clause evaluates to false. `false && anything` is false, so the next clause: 

`(comparison = Integer.compare(mediaOffsetTime, other.mediaOffsetTime)) == 0 &&`

is not evaluated. The whole`else if(comparison = ..., comparison = ..., comparison = ...)` case essentially becomes `else if(false)` so when `return comparison;` is invoked at the end of `compareTo` it still has the value of the initial comparison between the mediaOffsetFrame values.

<br>Now consider when:

`(comparison = Integer.compare(mediaOffsetFrame, other.mediaOffsetFrame)) == 0 &&`

sets `comparison` to 0. The `==` part of the clause evaluates to true. `true && anything` is that anything, so the next clause is evaluated:

`(comparison = Integer.compare(mediaOffsetTime, other.mediaOffsetTime)) == 0 &&`

That compares the mediaOffsetTime and sets the value of `comparison` to -1, 0, or 1. If the value is 0 then the whole`else if(comparison = ..., comparison = ..., comparison = ...)` case essentially becomes `else if(false)` so when `return comparison;` is invoked at the end of `compareTo` it still has the value of the initial comparison between the mediaOffsetTime values.


<br> The logic here is a bit hard to parse at first, but it is very concise if you know what it's doing. The key is understanding how Java evaluates the big conditional from left to right and ["short circuits"](https://stackoverflow.com/questions/8759868/java-logical-operator-short-circuiting) if necessary.
</blockquote></details>

Jeff, I was looking at this too.
I think this is preferable:
```
 private static final Comparator<Detection> DEFAULT_COMPARATOR = Comparator
        .nullsFirst(Comparator
                .comparingInt(Detection::getMediaOffsetFrame)
                .thenComparingInt(Detection::getMediaOffsetTime)
                .thenComparingInt(Detection::getX)
                .thenComparingInt(Detection::getY)
                .thenComparingInt(Detection::getWidth)
                .thenComparingInt(Detection::getHeight)
                .thenComparingDouble(Detection::getConfidence)
                .thenComparing(Detection::getArtifactPath, TextUtils::nullSafeCompare)
                .thenComparing(Detection::getDetectionProperties, Detection::compareMap));

@Override
public int compareTo(Detection other) {
    return DEFAULT_COMPARATOR.compare(this, other);
}
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/237)*

jrobble:
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/transients/Detection.java, line 168 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/237#-KsoRZ699zaZ0zknJ56-:-Kso_9TmT8rrF20w_NHT:b-srv2b9) ([raw file](https://github.com/openmpf/openmpf/blob/93a00d011a317fca76f65d6f07cc66a711af1299/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/transients/Detection.java#L168)):*

Jeff, I was looking at this too.
I think this is preferable:
```
 private static final Comparator<Detection> DEFAULT_COMPARATOR = Comparator
        .nullsFirst(Comparator
                .comparingInt(Detection::getMediaOffsetFrame)
                .thenComparingInt(Detection::getMediaOffsetTime)
                .thenComparingInt(Detection::getX)
                .thenComparingInt(Detection::getY)
                .thenComparingInt(Detection::getWidth)
                .thenComparingInt(Detection::getHeight)
                .thenComparingDouble(Detection::getConfidence)
                .thenComparing(Detection::getArtifactPath, TextUtils::nullSafeCompare)
                .thenComparing(Detection::getDetectionProperties, Detection::compareMap));

@Override
public int compareTo(Detection other) {
    return DEFAULT_COMPARATOR.compare(this, other);
}
```

</blockquote></details>

Very nice! That does seem easier to mentally parse. Leave it to Brian to pull out the Java 8 features.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/237)*

brosenberg42:
Review status: 1 of 6 files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/transients/Detection.java, line 52 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/237#-KsojA2DAhP6KqnErEiw:-KsojA2DAhP6KqnErEix:b4ak67b) ([raw file](https://github.com/openmpf/openmpf/blob/93a00d011a317fca76f65d6f07cc66a711af1299/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/transients/Detection.java#L52)):*
> ```Java
>  * In the context of audio files, detections are not yet well-defined.
>  * */
> public class Detection implements Comparable<Detection>, Comparator<Detection> {
> ```

A class should not be its own Comparator, that is what Comparable is for. Detection should only implement Comparable.
When a class is Comparable that means it has a natural order. A Comparator is used when you want to sort objects based on an alternative ordering.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/transients/Detection.java, line 55 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/237#-KsojVNxtKAIp9xTmm57:-KsojVNxtKAIp9xTmm58:b-sc8iw3) ([raw file](https://github.com/openmpf/openmpf/blob/93a00d011a317fca76f65d6f07cc66a711af1299/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/transients/Detection.java#L55)):*
> ```Java
> 
>   /** Alternative Comparator developed for prioritizing Detections by (1) descending confidence, (2) ascending media frame offset, (3) ascending media time offset */
>   public static final Comparator<Detection> DetectionConfidenceComparator = new Comparator<Detection> () {
> ```

This isn't necessary if you use the getTopConfidenceDetections method in my other comment.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/transients/Detection.java, line 145 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/237#-Ksok1gHogVzNRe0MNdr:-Ksok1gHogVzNRe0MNds:b-5051a3) ([raw file](https://github.com/openmpf/openmpf/blob/93a00d011a317fca76f65d6f07cc66a711af1299/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/transients/Detection.java#L145)):*
> ```Java
>    */
>   @Override
>   public int compare(Detection d1, Detection d2) {
> ```

This method can just be removed when you remove implements Comparator from this class. 
The `compareTo(Detection other)` method is what actually defines the natural order.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/TimeUtils.java, line 94 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/237#-KsocHKxiwqSKzZpI2QO:-KsocHKxiwqSKzZpI2QP:b-20jry2) ([raw file](https://github.com/openmpf/openmpf/blob/93a00d011a317fca76f65d6f07cc66a711af1299/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/TimeUtils.java#L94)):*
> ```Java
>         SortedSet<Detection> detections = track.getDetections();
>         PriorityQueue <Detection> priorityQueue = new PriorityQueue(detections.size(), Detection.DetectionConfidenceComparator);
>         priorityQueue.addAll(detections);
> ```

If the priority queue has more than (topConfidenceCount + 1) elements at a time, you aren't going to get any better performance than just sorting the whole list.

For these examples, `n = tracks.size()` and `k = topConfidenceCount`.
If you sort the whole list it will take `O(n * lg(n))` steps.

Adding an item to a priority queue takes `O(lg(pq.size()))` steps, thus adding n items to a priority queue takes `O(n * lg(pq.size())` steps.
Since you are adding all the detections to the priority queue this will take `O(n * lg(n))` steps. The same as sorting.

Since you only need to get the k items you only need to store at most k + 1 items in the priority queue. If the queue is always of size k + 1 adding an item to the queue will take `O(lg(k))` steps to add each item. To this for the n items it will then take `O(n * lg(k)` steps. The code below gets the top confidence detections in `O(n * lg(k)` steps.

You can use the method below to get a sorted (high to low) list of the top detections. You can then create the time pairs in basically the same way you are now, except you don't have to deal with the priority queue.
```
private static List<Detection> getTopConfidenceDetections(Collection<Detection> allDetections, int topConfidenceCount) {

    PriorityQueue<Detection> priorityQueue = new PriorityQueue<>(
            topConfidenceCount,
            Comparator.comparingDouble(Detection::getConfidence)
                    .thenComparing(Comparator.naturalOrder()));

    for (Detection detection : allDetections) {
        priorityQueue.add(detection);
        if (priorityQueue.size() > topConfidenceCount) {
            // Remove lowest confidence detection.
            priorityQueue.poll();
        }
    }

    List<Detection> result = new ArrayList<>(priorityQueue.size());
    while (!priorityQueue.isEmpty()) {
        result.add(priorityQueue.poll());
    }
    Collections.reverse(result);
    return result;
}
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/237)*

jrobble:
Review status: 1 of 6 files reviewed at latest revision, 23 unresolved discussions, some commit checks failed.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/transients/Track.java, line 29 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/237#-KsoaBBdoUundIbqQLf_:-KsoaBBdoUundIbqQLfa:be3hin5) ([raw file](https://github.com/openmpf/openmpf/blob/93a00d011a317fca76f65d6f07cc66a711af1299/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/transients/Track.java#L29)):*
> ```Java
> package org.mitre.mpf.wfm.data.entities.transients;
> 
> import java.util.Comparator;
> ```

Comparator isn't used in this class. Please remove this import.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/transients/Track.java, line 39 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/237#-Ksoa7n6xHatVjPUXzCl:-Ksoa7n6xHatVjPUXzCm:bx4jlow) ([raw file](https://github.com/openmpf/openmpf/blob/93a00d011a317fca76f65d6f07cc66a711af1299/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/transients/Track.java#L39)):*
> ```Java
> import java.util.SortedSet;
> import java.util.TreeSet;
> import org.mitre.mpf.wfm.util.TimePair;
> ```

TimePair isn't used in this class. Please remove this import.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/transients/Track.java, line 143 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/237#-Ksoa14914tetNUa8vnD:-Ksoa14914tetNUa8vnE:bkx8hv9) ([raw file](https://github.com/openmpf/openmpf/blob/93a00d011a317fca76f65d6f07cc66a711af1299/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/transients/Track.java#L143)):*
> ```Java
> 	@Override
> 	public int hashCode() {
> 	  return Objects.hash(jobId,mediaId,stageIndex,actionIndex,startOffsetFrameInclusive,endOffsetFrameInclusive,
> ```

Please add spaces between commas.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/TimePair.java, line 94 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/237#-Ksoc5dTQDlNsD-DcBLJ:-Ksoc5dUb9fRBI3ioCz7:bd9w9ij) ([raw file](https://github.com/openmpf/openmpf/blob/93a00d011a317fca76f65d6f07cc66a711af1299/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/TimePair.java#L94)):*
> ```Java
>    * @return Returns true if this TimePair is unbounded (i.e. endInclusive has not been set)
>    */
> 	public boolean isUnboundedTimePair() { return endInclusive == -1; }
> ```

Please change this to `isUnbounded()`.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/TimePair.java, line 99 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/237#-KsocAj7Cy_2lIT7lwz6:-KsocAj7Cy_2lIT7lwz7:bjifvb7) ([raw file](https://github.com/openmpf/openmpf/blob/93a00d011a317fca76f65d6f07cc66a711af1299/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/TimePair.java#L99)):*
> ```Java
>    * @return Returns true if this TimePair is closed (i.e. it is not unbounded)
>    */
>   public boolean isClosedTimePair() { return !isUnboundedTimePair(); }
> ```

Please change this to `isClosed()`.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/TimePair.java, line 126 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/237#-KsocNWpeEire-u3EYcM:-KsocNWqzgR92rbOnA4Z:b-e6zkfh) ([raw file](https://github.com/openmpf/openmpf/blob/93a00d011a317fca76f65d6f07cc66a711af1299/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/TimePair.java#L126)):*
> ```Java
> 		  // TimePairs are equivalent if the start/end boundaries are equal
> 			result =  0;
> 		} else if ( this.startInclusive < other.startInclusive ) {
> ```

Similar to `Detection.compareTo()`, this does "short circuiting" so the cases you added should not be necessary. The original logic was sufficient. You can employ the Java 8 `thenComparingInt`approach here too.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/TimeUtils.java, line 87 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/237#-KsoecTxUSxaJRIAKuAT:-KsoecTxUSxaJRIAKuAU:blgp4t8) ([raw file](https://github.com/openmpf/openmpf/blob/93a00d011a317fca76f65d6f07cc66a711af1299/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/TimeUtils.java#L87)):*
> ```Java
>         // only use the exemplar to form the TimePair for the feed-forward track
>         Detection exemplarDetection = track.getExemplar();
>         int exemplarFrameOffsetInclusive = track.getStartOffsetFrameInclusive() + exemplarDetection.getMediaOffsetFrame();
> ```

I'm confused by this line. I noticed that you add the start offset frame to the media offset frame in a few places in your code.

Isn't `mediaOffsetFrame` the same as the frame index? For example, if it's 100 then it's the 101st frame in the video (accounting for frame index 0). Why are we adding the start offset frame to that?

I could understand of the `mediaOffsetFrame` is locally scoped to the segment range, but I think it's globally scoped to the whole video range.

Looking at `DetectionResponseProcessor.generateTrack()`, it seems to set `mediaOffsetFrame` to the frame number pulled from the detection response protobuf sent by the component. The components do the reverse transform to ensure what they return is relative to the whole video (coordinate region and frame index).

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/TimeUtils.java, line 166 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/237#-KsoihKEWKLpqWwZvJWZ:-KsoihKFDzY_l0NEofyy:b4lr6qk) ([raw file](https://github.com/openmpf/openmpf/blob/93a00d011a317fca76f65d6f07cc66a711af1299/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/TimeUtils.java#L166)):*
> ```Java
>    * @return returns the list of segment TimePairs.  The list may be empty if the input collection is empty or null.
>    */
>   public List<TimePair> createSegments(Integer topConfidenceCount, SortedSet<Track> trackSet, int targetSegmentLength, int minSegmentLength) {
> ```

My preference is to use a primitive `int` for `topConfidenceCount` rather than an `Integer`. That way we don't have to worry about NullPointerExceptions. If that property is not provided, we should just pass in a -1.

---

*[trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/TestSegmenting.java, line 68 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/237#-Ksp0ff8japEHk0xCsBF:-Ksp0ff8japEHk0xCsBG:b-4k1vcp) ([raw file](https://github.com/openmpf/openmpf/blob/93a00d011a317fca76f65d6f07cc66a711af1299/trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/TestSegmenting.java#L68)):*
> ```Java
>    */
>   @Test(timeout = 5 * MINUTES)
>   public void TestFeedForwardSegmentProcessing() throws Exception {
> ```

We should also add tests for:

- When the detection confidence values aren't increasing with the frame index and are more randomly distributed throughout the detections in the track. You have some tests where all of the top confidence values are the same, which is different.

- When track size is less than segment size.

- When segment size does not divide evenly into track size.

---

*[trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/TestSegmenting.java, line 95 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/237#-KsorEAM5jqK2AGEwEU4:-KsorEANLcHZrG_svZcB:bkxaa4j) ([raw file](https://github.com/openmpf/openmpf/blob/93a00d011a317fca76f65d6f07cc66a711af1299/trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/TestSegmenting.java#L95)):*
> ```Java
>         confidence = 60 + i; // high confidence, increasing with the number of detections for the first <targetSegmentLength> frames
>       } else {
>         confidence = i < 30 ? 30 - i : 0; // low confidence, decreasing with the number of detections for the remainder of the frames
> ```

This loop will never get beyond i = 29, so the `i < 30` part of this line seems unnecessary.

---

*[trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/TestSegmenting.java, line 110 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/237#-Ksp2VXLRwytpNWgJPxg:-Ksp2VXMaPGibc0RkZ2h:bol2be2) ([raw file](https://github.com/openmpf/openmpf/blob/93a00d011a317fca76f65d6f07cc66a711af1299/trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/TestSegmenting.java#L110)):*
> ```Java
> 
>     // run preliminary tests
>     Assert.assertTrue("FeedForward check Ex1 assertion failed, start frame offset is not "+startOffsetFrameInclusive,testTrack.getStartOffsetFrameInclusive()==startOffsetFrameInclusive);
> ```

Instead of using `Assert.assertTrue()` please use `Assert.equals()` instead (when you can).

---

*[trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/TestSegmenting.java, line 120 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/237#-KsosGQJDiGmJmgyl_-t:-KsosGQKFAWaANNZNuVv:b-3dovgi) ([raw file](https://github.com/openmpf/openmpf/blob/93a00d011a317fca76f65d6f07cc66a711af1299/trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/TestSegmenting.java#L120)):*
> ```Java
>     List<TimePair> resultsEx = timeUtils.createSegments(topConfidenceCount, testTrackSet, targetSegmentLength, minSegmentLength);
> 
>     int expectedNumSegments = 1; // for this test, the top <topConfidenceCount> == <targetSegmentLength}, we expect to only get 1 segment
> ```

Replace:

`<targetSegmentLength}`

with:

`<targetSegmentLength>`

---

*[trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/TestSegmenting.java, line 131 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/237#-KsowT4zSFchKApa7O82:-KsowT4zSFchKApa7O83:bqp1ygz) ([raw file](https://github.com/openmpf/openmpf/blob/93a00d011a317fca76f65d6f07cc66a711af1299/trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/TestSegmenting.java#L131)):*
> ```Java
>     }
> 
>     // Example 1a, test with topConfidenceCount=1 (use the exemplar), should still return 1 segment but of duration from the exemplar
> ```

Each of these sub-parts should be in its own test method.

---

*[trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/TestSegmenting.java, line 160 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/237#-Ksou_ARqXj1R3_JzcWN:-Ksou_ASRPkLqryh38RV:b-3zp0ol) ([raw file](https://github.com/openmpf/openmpf/blob/93a00d011a317fca76f65d6f07cc66a711af1299/trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/TestSegmenting.java#L160)):*
> ```Java
>     topConfidenceCount = 10; // specify that 10 of the top confidence detections should be used for feed-forward in this test
>     testDetections = new TreeSet<Detection>();
>     boolean isRemoved = testTrackSet.remove(testTrack); // remove the testTrack placed for the last set of tests so we can re-add it to run this test.
> ```

This is confusing. Please use a different set for this test.

---

*[trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/TestSegmenting.java, line 172 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/237#-KsovAjAVihFcHFhE1-j:-KsovAjAVihFcHFhE1-k:bor23e7) ([raw file](https://github.com/openmpf/openmpf/blob/93a00d011a317fca76f65d6f07cc66a711af1299/trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/TestSegmenting.java#L172)):*
Again:

> This loop will never get beyond i = 29, so the i < 30 part of this line seems unnecessary.

---

*[trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/TestSegmenting.java, line 220 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/237#-Ksp1spca0p3BTUY3tJJ:-Ksp1spca0p3BTUY3tJK:bdae9q5) ([raw file](https://github.com/openmpf/openmpf/blob/93a00d011a317fca76f65d6f07cc66a711af1299/trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/TestSegmenting.java#L220)):*
> ```Java
>     }
> 
>     // Example 2: segments to include frames that don't have missing detections,  Frames containing detections with top confidence are evenly dispersed within the track
> ```

Each example should be a separate test method.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/237)*

jrobble:
Note that this PR is just the beginning (but possibly the hardest part). All of the items in the comments on #176 need to be addressed for the WFM to fully support feed-forward.

---

Reviewed 5 of 6 files at r1.
Review status: all files reviewed at latest revision, 23 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/237#-:-Ksp3cemRwbn9FJ0d-AY:btclfe)*

brosenberg42:
Fixed some formatting and comparison methods. Issues in TimeUtils and tests haven't been addressed yet.

---

Review status: 0 of 6 files reviewed at latest revision, 23 unresolved discussions.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/DetectionSplitter.java, line 64 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/237#-KsoNu94SIP3pw-Az4Lb:-Ksts3M_y9vhYHXURwA3:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/93a00d011a317fca76f65d6f07cc66a711af1299/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/DetectionSplitter.java#L64)):*

Looks like left indentation is a bit off for this autowire and the one above it.

Please check your left indentation for your other additions as well. It seems that you may have the indent set to 2 spaces instead of 4.
</blockquote></details>

Done.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/transients/Detection.java, line 33 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/237#-KsoOdExi1ZV9WPhnZWI:-Ksts50Lcsci9Eg7jYRk:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/93a00d011a317fca76f65d6f07cc66a711af1299/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/transients/Detection.java#L33)):*

This is a duplicate.
</blockquote></details>

Done.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/transients/Track.java, line 29 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/237#-KsoaBBdoUundIbqQLf_:-KstsJ8jlDAhXfyjbvMX:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/93a00d011a317fca76f65d6f07cc66a711af1299/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/transients/Track.java#L29)):*

Comparator isn't used in this class. Please remove this import.
</blockquote></details>

Done.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/transients/Track.java, line 39 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/237#-Ksoa7n6xHatVjPUXzCl:-KstsK9IPKUhKRTIoLJM:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/93a00d011a317fca76f65d6f07cc66a711af1299/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/transients/Track.java#L39)):*

TimePair isn't used in this class. Please remove this import.
</blockquote></details>

Done.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/TimePair.java, line 94 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/237#-Ksoc5dTQDlNsD-DcBLJ:-KstsTOJVqQRmKn_TDyF:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/93a00d011a317fca76f65d6f07cc66a711af1299/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/TimePair.java#L94)):*

Please change this to `isUnbounded()`.
</blockquote></details>

Done.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/TimePair.java, line 99 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/237#-KsocAj7Cy_2lIT7lwz6:-KstsUEA9NLhBEOXBa-1:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/93a00d011a317fca76f65d6f07cc66a711af1299/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/TimePair.java#L99)):*

Please change this to `isClosed()`.
</blockquote></details>

Done.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/TimePair.java, line 126 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/237#-KsocNWpeEire-u3EYcM:-KstsW1rX1KRliV-L6gW:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/93a00d011a317fca76f65d6f07cc66a711af1299/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/TimePair.java#L126)):*

Similar to `Detection.compareTo()`, this does "short circuiting" so the cases you added should not be necessary. The original logic was sufficient. You can employ the Java 8 `thenComparingInt`approach here too.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/237#-:-KstuAfNFHqgWX18F1jv:b-mtri6t)*

jrobble:
Reviewed 6 of 6 files at r2.
Review status: all files reviewed at latest revision, 16 unresolved discussions.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/transients/Detection.java, line 118 at r2](https://reviewable.io:443/reviews/openmpf/openmpf/237#-Ksu-59R2LBh_svmdw9w:-Ksu-59R2LBh_svmdw9x:b-jwviuu) ([raw file](https://github.com/openmpf/openmpf/blob/32ef74d31d0392da6979c5c3b457c4f1e67309bb/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/transients/Detection.java#L118)):*
> ```Java
> 
> 
> 	private static final Comparator<Map.Entry<String, String>> ENTRY_COMPARATOR =
> ```

Let's call this MAP_ENTRY_COMPARATOR.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/237)*

jrobble:
Review status: all files reviewed at latest revision, 18 unresolved discussions.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/TimeUtils.java, line 121 at r2](https://reviewable.io:443/reviews/openmpf/openmpf/237#-KsyUOisaF9WVU-WdqND:-KsyUOisaF9WVU-WdqNE:b-rlihcs) ([raw file](https://github.com/openmpf/openmpf/blob/32ef74d31d0392da6979c5c3b457c4f1e67309bb/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/TimeUtils.java#L121)):*
> ```Java
>           }
>           // once the counter has reached the target segment length, do TimePair processing.
>           if (detectionCounter % targetSegmentLength == 0) {
> ```

I talked to the end users today and decided on a simplified approach: When feed-forward is enabled, we should generate one segment for the whole feed-forward track. We don't care what the segment size property is set to. The segment size should only be used to generate segments for the first stage of the pipeline when feed-forward is enabled.

Thus, the segment start frame should be set to the index of the high confidence detection we care about with the lowest frame index, and the segment stop frame should be set to the index of the high confidence detection we care about with the lowest frame index.

---

*[trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/TestSegmenting.java, line 200 at r2](https://reviewable.io:443/reviews/openmpf/openmpf/237#-KsyWsnamyDvQh0F4xbH:-KsyWsnbN0aMpbztgOZX:b-nlif71) ([raw file](https://github.com/openmpf/openmpf/blob/32ef74d31d0392da6979c5c3b457c4f1e67309bb/trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/TestSegmenting.java#L200)):*
> ```Java
>     }
> 
>     // Test Ex1d, where targetSegmentLength>1 and targetSegmentLength>topConfidenceCount
> ```

Test Ex1d is no longer necessary since we are now only going to be generating one segment per feed-forward track.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/237)*

jrobble:
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/TimeUtils.java, line 121 at r2](https://reviewable.io:443/reviews/openmpf/openmpf/237#-KsyUOisaF9WVU-WdqND:-Ksy_LAwWcR4ZU4Y1yEw:b-xjxpqs) ([raw file](https://github.com/openmpf/openmpf/blob/32ef74d31d0392da6979c5c3b457c4f1e67309bb/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/TimeUtils.java#L121)):*

I talked to the end users today and decided on a simplified approach: When feed-forward is enabled, we should generate one segment for the whole feed-forward track. We don't care what the segment size property is set to. The segment size should only be used to generate segments for the first stage of the pipeline when feed-forward is enabled.

Thus, the segment start frame should be set to the index of the high confidence detection we care about with the lowest frame index, and the segment stop frame should be set to the index of the high confidence detection we care about with the lowest frame index.
</blockquote></details>

Correction:

Thus, the segment start frame should be set to the index of the high confidence detection we care about with the lowest frame index, and the segment stop frame should be set to the index of the high confidence detection we care about with the **highest** frame index.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/237)*

jrobble:
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/TimeUtils.java, line 74 at r2](https://reviewable.io:443/reviews/openmpf/openmpf/237#-Ksyccoikp1b7TpVVa6L:-Ksyccojd6BxP72tLjf5:bn8m0sj) ([raw file](https://github.com/openmpf/openmpf/blob/32ef74d31d0392da6979c5c3b457c4f1e67309bb/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/TimeUtils.java#L74)):*
> ```Java
>    * @return sorted list of TimePairs.
>    */
>   public List<TimePair> createTimePairsForTracks(Integer topConfidenceCount, int targetSegmentLength, SortedSet<Track> tracks) {
> ```

Our approach has changed. This method should generate one TimePair per track. 

When we tackle #228, the WFM will need to remove all of the detections from the feed-forward track that do not have a high confidence. That modified track will be passed down to the component in the MPFVideoJob.

We're already doing all the work in this method to determine what the high confidence detections are. It may make sense to have this method remove the low confidence detections from each of the tracks in the passed-in `tracks` set.

Then, whatever part of the WFM calls this method can use those tracks to generate the job request messages.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/237)*

brosenberg42:
I was able to do some testing just by adding the feed forward properties to OcvFace.
```
        {
          "name": "FEED_FORWARD_TYPE",
          "description": "FEED_FORWARD_TYPE",
          "type": "STRING",
          "defaultValue": "NONE"
        },
        {
          "name": "FEED_FORWARD_TOP_CONFIDENCE_COUNT",
          "description": "FEED_FORWARD_TOP_CONFIDENCE_COUNT",
          "type": "INT",
          "defaultValue": "0"
        }
```

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/237#-:-KtXr7d0d2eT7KgjAxIZ:b-nvenn0)*

jrobble:
Which video did you use? Did it behave as expected?

---

Reviewed 4 of 15 files at r3.
Review status: 5 of 16 files reviewed at latest revision, 20 unresolved discussions.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/DetectionSplitter.java, line 146 at r3](https://reviewable.io:443/reviews/openmpf/openmpf/237#-KtYRkDAyu8RxvQ64EnT:-KtYRkDBH3VjAU8N113V:boolmw0) ([raw file](https://github.com/openmpf/openmpf/blob/9edafe8eeb9a2852ca2afb56472e35465590984c/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/DetectionSplitter.java#L146)):*
> ```Java
> 			// tracks. Note that the TimePairs created for these Tracks use the non-feed-forward version of timeUtils.createTimePairsForTracks
> 			// TODO look here for any modifications required to be made to support feed-forward
> 			SortedSet<Track> previousTrack;
> ```

This should be `previousTracks`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/237#-:-KtYPlNdWS0kW2SXH_FT:b-enwy5p)*

jrobble:
Does this PR also cover #240? I left a comment on that issue about a SPHINX test to make sure the audio segmenter is working properly with FF.

---

Reviewed 11 of 15 files at r3, 5 of 5 files at r4.
Review status: all files reviewed at latest revision, 39 unresolved discussions.

---

*[trunk/mpf-system-tests/src/main/java/org/mitre/mpf/mst/TestSystem.java, line 248 at r4](https://reviewable.io:443/reviews/openmpf/openmpf/237#-KtrNNlo77LGadQ0cRec:-KtrNNlpOAhuv_SNnLaU:bw9abw5) ([raw file](https://github.com/openmpf/openmpf/blob/eb1bd30db41393f63702cb100c9b75558dc4b5eb/trunk/mpf-system-tests/src/main/java/org/mitre/mpf/mst/TestSystem.java#L248)):*
> runPipelineOnMedia(String pipelineName, List<JsonMediaInputObject> media, boolean buildOutput, int priority) throws Exception {

This is currently used in TestSystemStress in the other repo. This change will require updating the other repo too.

---

*[trunk/mpf-system-tests/src/main/java/org/mitre/mpf/mst/TestSystem.java, line 314 at r4](https://reviewable.io:443/reviews/openmpf/openmpf/237#-KtrR8xIi_Idip86hLbl:-KtrR8xIi_Idip86hLbm:b5pg6sy) ([raw file](https://github.com/openmpf/openmpf/blob/eb1bd30db41393f63702cb100c9b75558dc4b5eb/trunk/mpf-system-tests/src/main/java/org/mitre/mpf/mst/TestSystem.java#L314)):*
> ```Java
> 
> 
> 	protected JsonOutputObject getJobOutputObject(long jobId) throws IOException {
> ```

Other places that use OBJECT_MAPPER could be updated to use this method.

---

*[trunk/mpf-system-tests/src/test/java/org/mitre/mpf/mst/TestSystemOnDiff.java, line 162 at r4](https://reviewable.io:443/reviews/openmpf/openmpf/237#-KtrRhnomjWA_6NeV3QS:-KtrRhnomjWA_6NeV3QT:b-8xye32) ([raw file](https://github.com/openmpf/openmpf/blob/eb1bd30db41393f63702cb100c9b75558dc4b5eb/trunk/mpf-system-tests/src/test/java/org/mitre/mpf/mst/TestSystemOnDiff.java#L162)):*
> ```Java
> 
> 		assertFalse(detections.isEmpty());
> 		int firstMotionFrame = 31;
> ```

Please drop a comment explaining this number. The first 31 frames of the video are the same, so there's no motion.

Please explain in the test below too.

---

*[trunk/mpf-system-tests/src/test/java/org/mitre/mpf/mst/TestSystemOnDiff.java, line 166 at r4](https://reviewable.io:443/reviews/openmpf/openmpf/237#-KtrRuINXTARtxvesFsK:-KtrRuINXTARtxvesFsL:b-1azg5o) ([raw file](https://github.com/openmpf/openmpf/blob/eb1bd30db41393f63702cb100c9b75558dc4b5eb/trunk/mpf-system-tests/src/test/java/org/mitre/mpf/mst/TestSystemOnDiff.java#L166)):*
> ```Java
> 				           .allMatch(d -> d.getOffsetFrame() >= firstMotionFrame));
> 
> 		int maxXMotion = 320;
> ```

Please drop a comment explaining this number.

Please explain in the test below too.

---

*[trunk/mpf-system-tests/src/test/resources/samples/face/NOTICE, line 23 at r4](https://reviewable.io:443/reviews/openmpf/openmpf/237#-KtrTL_dsrabd6ZQp2k9:-KtrTL_dsrabd6ZQp2kA:b-cv4dqz) ([raw file](https://github.com/openmpf/openmpf/blob/eb1bd30db41393f63702cb100c9b75558dc4b5eb/trunk/mpf-system-tests/src/test/resources/samples/face/NOTICE#L23)):*
> ```
> 
> # ff-region-motion-face.avi
> Based on new_face_video.avi
> ```

Add period at end.

---

*[trunk/mpf-system-tests/src/test/resources/samples/face/NOTICE, line 27 at r4](https://reviewable.io:443/reviews/openmpf/openmpf/237#-KtrTN4Aa2JsDIQ33Qx4:-KtrTN4Aa2JsDIQ33Qx5:b-cv4dqz) ([raw file](https://github.com/openmpf/openmpf/blob/eb1bd30db41393f63702cb100c9b75558dc4b5eb/trunk/mpf-system-tests/src/test/resources/samples/face/NOTICE#L27)):*
> ```
> Creative Commons Attribution 2.0 Generic License: https://creativecommons.org/licenses/by/2.0/
> https://www.flickr.com/photos/hahatango/7484448042/in/photolist-cpnKDC-9PQvQE-cehi5m-eX6cjM-7XHcXs-8HwEEG-97vd4h-cpo5cS-6fDWiQ-9phRV3-bAPcDX-cpo5ky-7DrHhu-6UNght-8GPWGS-cehi4J-BUk6xN-7DrqHQ-7DrzRy-7eLZPm-6UN5pn-8Hwx5d-88XgZj-xNoj9R-HHwC5c-oiWXVh-oAbTv3-86XZiL-7urFMR-6UShyC-7DnKEP-7DnQU8-7Dryxb-6TU3nb-oNw7jC-6teNT2-6bZoZj-7Drxb5-cpouDu-6UN6Wv-cpokHj-9jEhVN-7DrQ1q-cpouub-7DrPNN-7DoeZ6-8km6iE-7Do4St-8kZ2Ni-83HeW4
> Extracted two faces. Left face is between column 0 and 319 and right face is between column 320 and 639
> ```

Add period at end.

---

*[trunk/mpf-system-tests/src/test/resources/samples/face/NOTICE, line 29 at r4](https://reviewable.io:443/reviews/openmpf/openmpf/237#-KtrTO_MqwLJZugsjHO4:-KtrTO_NQX1pvuVJqVTz:b-cv4dqz) ([raw file](https://github.com/openmpf/openmpf/blob/eb1bd30db41393f63702cb100c9b75558dc4b5eb/trunk/mpf-system-tests/src/test/resources/samples/face/NOTICE#L29)):*
> ```
> Extracted two faces. Left face is between column 0 and 319 and right face is between column 320 and 639
> No motion in first 30 frames.
> Starting at frame 31 face on left moves
> ```

Add period at end.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/segmenting/MediaSegmenter.java, line 81 at r4](https://reviewable.io:443/reviews/openmpf/openmpf/237#-KtrWgvHMy03uT7n2kiu:-KtrWgvHMy03uT7n2kiv:bvw34wt) ([raw file](https://github.com/openmpf/openmpf/blob/eb1bd30db41393f63702cb100c9b75558dc4b5eb/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/segmenting/MediaSegmenter.java#L81)):*
> ```Java
> 		if (context.isFirstDetectionStage()) {
> 			return context.getAlgorithmProperties().stream()
> 					.filter(ap -> !ap.getPropertyName().equalsIgnoreCase("FEED_FORWARD_TYPE"))
> ```

I think we should we also filter out FEED_FORWARD_TOP_CONFIDENCE_COUNT.

---

*[trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/segmenting/TestAudioMediaSegmenter.java, line 8 at r4](https://reviewable.io:443/reviews/openmpf/openmpf/237#-KtrZPn21mnu4j3USiQD:-KtrZPn3mMz4Cwi-z2Gr:b-k09zl5) ([raw file](https://github.com/openmpf/openmpf/blob/eb1bd30db41393f63702cb100c9b75558dc4b5eb/trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/segmenting/TestAudioMediaSegmenter.java#L8)):*
> ```Java
>  * 52.227-14, Alt. IV (DEC 2007).                                             *
>  *                                                                            *
>  * Copyright 2016 The MITRE Corporation. All Rights Reserved.                 *
> ```

Update to 2017.

---

*[trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/segmenting/TestAudioMediaSegmenter.java, line 12 at r4](https://reviewable.io:443/reviews/openmpf/openmpf/237#-KtrZRVQAlBAf13LtN-m:-KtrZRVRM0cARrRn7Lxt:b-k09zl5) ([raw file](https://github.com/openmpf/openmpf/blob/eb1bd30db41393f63702cb100c9b75558dc4b5eb/trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/segmenting/TestAudioMediaSegmenter.java#L12)):*
> ```Java
> 
> /******************************************************************************
>  * Copyright 2016 The MITRE Corporation                                       *
> ```

Update to 2017.

---

*[trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/segmenting/TestAudioMediaSegmenter.java, line 157 at r4](https://reviewable.io:443/reviews/openmpf/openmpf/237#-KtrkoziYUYBS7TLeHcb:-Ktrkozj77YabJ-lBktx:b-bnsu5a) ([raw file](https://github.com/openmpf/openmpf/blob/eb1bd30db41393f63702cb100c9b75558dc4b5eb/trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/segmenting/TestAudioMediaSegmenter.java#L157)):*
> ```Java
> 	private static Set<Track> createTestTracks() {
> 		Detection detection1 = createDetection(5, 5);
> 		Track track1 = new Track(1, 1, 0, 0, 0,
> ```

If you look [here](https://github.com/openmpf/openmpf/blob/master/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/DetectionResponseProcessor.java#L222), the start and stop offset frames are calculated by multiplying the time offset by the FPS. We should have tests with tracks that start and stop at different frames. 

The audio segmenter will only generate one segment regardless of how many previous tracks there are, and the frame indexes, but I would feel better if we tested with non-overlapping frame indexes.

---

*[trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/segmenting/TestImageMediaSegmenter.java, line 8 at r4](https://reviewable.io:443/reviews/openmpf/openmpf/237#-KtrZUqaH-XnE1DJmcIj:-KtrZUqaH-XnE1DJmcIk:b-k09zl5) ([raw file](https://github.com/openmpf/openmpf/blob/eb1bd30db41393f63702cb100c9b75558dc4b5eb/trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/segmenting/TestImageMediaSegmenter.java#L8)):*
> ```Java
>  * 52.227-14, Alt. IV (DEC 2007).                                             *
>  *                                                                            *
>  * Copyright 2016 The MITRE Corporation. All Rights Reserved.                 *
> ```

Update to 2017.

---

*[trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/segmenting/TestImageMediaSegmenter.java, line 12 at r4](https://reviewable.io:443/reviews/openmpf/openmpf/237#-KtrZVNdTjy9FckXuz3I:-KtrZVNdTjy9FckXuz3J:b-k09zl5) ([raw file](https://github.com/openmpf/openmpf/blob/eb1bd30db41393f63702cb100c9b75558dc4b5eb/trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/segmenting/TestImageMediaSegmenter.java#L12)):*
> ```Java
> 
> /******************************************************************************
>  * Copyright 2016 The MITRE Corporation                                       *
> ```

Update to 2017.

---

*[trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/segmenting/TestMediaSegmenter.java, line 8 at r4](https://reviewable.io:443/reviews/openmpf/openmpf/237#-KtrZWhO04m3_gR7FMbY:-KtrZWhPEtpEpbYgDDMB:b-k09zl5) ([raw file](https://github.com/openmpf/openmpf/blob/eb1bd30db41393f63702cb100c9b75558dc4b5eb/trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/segmenting/TestMediaSegmenter.java#L8)):*
> ```Java
>  * 52.227-14, Alt. IV (DEC 2007).                                             *
>  *                                                                            *
>  * Copyright 2016 The MITRE Corporation. All Rights Reserved.                 *
> ```

Update to 2017.

---

*[trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/segmenting/TestMediaSegmenter.java, line 12 at r4](https://reviewable.io:443/reviews/openmpf/openmpf/237#-KtrZXbNhLClgnoFbif0:-KtrZXbNhLClgnoFbif1:b-k09zl5) ([raw file](https://github.com/openmpf/openmpf/blob/eb1bd30db41393f63702cb100c9b75558dc4b5eb/trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/segmenting/TestMediaSegmenter.java#L12)):*
> ```Java
> 
> /******************************************************************************
>  * Copyright 2016 The MITRE Corporation                                       *
> ```

Update to 2017.

---

*[trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/segmenting/TestVideoMediaSegmenter.java, line 8 at r4](https://reviewable.io:443/reviews/openmpf/openmpf/237#-KtrZYzUokIyV3I6isKP:-KtrZYzUokIyV3I6isKQ:b-k09zl5) ([raw file](https://github.com/openmpf/openmpf/blob/eb1bd30db41393f63702cb100c9b75558dc4b5eb/trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/segmenting/TestVideoMediaSegmenter.java#L8)):*
> ```Java
>  * 52.227-14, Alt. IV (DEC 2007).                                             *
>  *                                                                            *
>  * Copyright 2016 The MITRE Corporation. All Rights Reserved.                 *
> ```

Update to 2017.

---

*[trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/segmenting/TestVideoMediaSegmenter.java, line 12 at r4](https://reviewable.io:443/reviews/openmpf/openmpf/237#-KtrZZa9RsIv9jSbbzOI:-KtrZZa9RsIv9jSbbzOJ:b-k09zl5) ([raw file](https://github.com/openmpf/openmpf/blob/eb1bd30db41393f63702cb100c9b75558dc4b5eb/trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/segmenting/TestVideoMediaSegmenter.java#L12)):*
> ```Java
> 
> /******************************************************************************
>  * Copyright 2016 The MITRE Corporation                                       *
> ```

Update to 2017.

---

*[trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/segmenting/TestVideoMediaSegmenter.java, line 122 at r4](https://reviewable.io:443/reviews/openmpf/openmpf/237#-KtraxcjvXvKNOmx2SP3:-KtraxcjvXvKNOmx2SP4:bwkmliv) ([raw file](https://github.com/openmpf/openmpf/blob/eb1bd30db41393f63702cb100c9b75558dc4b5eb/trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/segmenting/TestVideoMediaSegmenter.java#L122)):*
> ```Java
> 		DetectionProtobuf.VideoTrack shortTrack;
> 		DetectionProtobuf.VideoTrack longTrack;
> 		if (detectionRequests.get(0).getVideoRequest().getFeedForwardTrack().getFrameLocationsCount() == 1) {
> ```

Drop a comment explaining that the track order in the JSON is random. Drop it here and wherever you have this kind of logic. At first glance it's odd.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/237#-:-KtrVJo0nERG9p-LHlPu:b-1vcgkw)*

brosenberg42:
No. I will add that test in the pull request that enables feed forward in Sphinx.

---

Review status: all files reviewed at latest revision, 38 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/237#-:-KtvoyKOuQHO8xCVbxdv:b-kqjcss)*

jrobble:
Reviewed 12 of 12 files at r5.
Review status: all files reviewed at latest revision, 28 unresolved discussions.

---

*[trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/segmenting/TestAudioMediaSegmenter.java, line 157 at r4](https://reviewable.io:443/reviews/openmpf/openmpf/237#-KtrkoziYUYBS7TLeHcb:-KtvzlhtcDc1dTye6gln:b-2zc3is) ([raw file](https://github.com/openmpf/openmpf/blob/eb1bd30db41393f63702cb100c9b75558dc4b5eb/trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/segmenting/TestAudioMediaSegmenter.java#L157)):*

If you look [here](https://github.com/openmpf/openmpf/blob/master/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/DetectionResponseProcessor.java#L222), the start and stop offset frames are calculated by multiplying the time offset by the FPS. We should have tests with tracks that start and stop at different frames. 

The audio segmenter will only generate one segment regardless of how many previous tracks there are, and the frame indexes, but I would feel better if we tested with non-overlapping frame indexes.


</blockquote></details>

I see that you changed the stop time and are now asserting the start and stop time values, which is good. 

I think we should have two segmenting tests (especially for non-FF). One that represents an audio file, and one that represents a video file. For the audio file, the start offset frame and stop offset frame will both be 0 (see link above). For the video file, the start offset frame and stop offset frame will be calculated using the FPS.

We should also be using representative time values. Where you have 5 and 10 those are 5 and 10 milliseconds, respectively. That's technically possible, but maybe 5000 and 10000 (5 and 10 seconds, respectively) are more reasonable.

Also, 5 and 10 don't work well for a test that represents a video file. Assuming a 60 FPS video, each frame represents 16 milliseconds, so 5 and 10 milliseconds get floored to frame 0. On the other hand, 5 and 10 seconds equate to frame 300 and 600, which are better values to use since they are different.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/237)*

brosenberg42:
Review status: all files reviewed at latest revision, 28 unresolved discussions.

---

*[trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/segmenting/TestAudioMediaSegmenter.java, line 157 at r4](https://reviewable.io:443/reviews/openmpf/openmpf/237#-KtrkoziYUYBS7TLeHcb:-Ku-pxxW_CO38VtFT6Wa:bsrzt6c) ([raw file](https://github.com/openmpf/openmpf/blob/eb1bd30db41393f63702cb100c9b75558dc4b5eb/trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/segmenting/TestAudioMediaSegmenter.java#L157)):*

I see that you changed the stop time and are now asserting the start and stop time values, which is good. 

I think we should have two segmenting tests (especially for non-FF). One that represents an audio file, and one that represents a video file. For the audio file, the start offset frame and stop offset frame will both be 0 (see link above). For the video file, the start offset frame and stop offset frame will be calculated using the FPS.

We should also be using representative time values. Where you have 5 and 10 those are 5 and 10 milliseconds, respectively. That's technically possible, but maybe 5000 and 10000 (5 and 10 seconds, respectively) are more reasonable.

Also, 5 and 10 don't work well for a test that represents a video file. Assuming a 60 FPS video, each frame represents 16 milliseconds, so 5 and 10 milliseconds get floored to frame 0. On the other hand, 5 and 10 seconds equate to frame 300 and 600, which are better values to use since they are different.
</blockquote></details>

I don't understand what test you are asking me to add. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/237)*

jrobble:
Reviewed 1 of 1 files at r6.
Review status: all files reviewed at latest revision, 28 unresolved discussions.

---

*[trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/segmenting/TestAudioMediaSegmenter.java, line 157 at r4](https://reviewable.io:443/reviews/openmpf/openmpf/237#-KtrkoziYUYBS7TLeHcb:-Ku5h6MNC2SBlmCHQg9o:b-ygwitd) ([raw file](https://github.com/openmpf/openmpf/blob/eb1bd30db41393f63702cb100c9b75558dc4b5eb/trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/segmenting/TestAudioMediaSegmenter.java#L157)):*

I don't understand what test you are asking me to add. 
</blockquote></details>

Scratch this. There was some confusion about using FPS to calculate frame offsets with audio files. Your recent update removes that confusing code.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/237)*

Issue #306, opened 2017-10-25
Add dependency injection to Node Manager


jrobble:
Add dependency injection to Node Manager to enable testing streaming video support.

PR #317, opened 2017-10-27
Make Node Manager use dependency injection

brosenberg42:
Issue #306

---
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/317)


jrobble:
The following fields in NodeManagerConstants.java are no longer used:

```
    public static final String AMQ_HOSTNAME_PARAM = "activeMqHostname";
    public static final String JGROUPS_CONFIG_PARAM = "jgroupsConfig";

    public static final String DEFAULT_CHANNEL = "MPF_Channel";
```

---

Reviewed 14 of 16 files at r1.
Review status: 12 of 14 files reviewed at latest revision, 7 unresolved discussions.

---

*[trunk/node-manager/src/main/java/org/mitre/mpf/nms/ChannelReceiver.java, line 342 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/317#-Kxo3sNHM9YDFDvasw9B:-Kxo3sNHM9YDFDvasw9C:bqr0szi) ([raw file](https://github.com/openmpf/openmpf/blob/c12f9be99700fe20daa8c4b0dd9405d596233f96/trunk/node-manager/src/main/java/org/mitre/mpf/nms/ChannelReceiver.java#L342)):*
> ```Java
> 
>         LOG.debug("{} starting up", fqn);
>         msgChannel = new ChannelNode();
> ```

Was the msgChannel created in the old constructor ever used?

---

*[trunk/node-manager/src/main/java/org/mitre/mpf/nms/ChildNodeStateManager.java, line 269 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/317#-KxoA9RY75UfWh3s0Pca:-KxoA9RY75UfWh3s0Pcb:b-skjpeq) ([raw file](https://github.com/openmpf/openmpf/blob/c12f9be99700fe20daa8c4b0dd9405d596233f96/trunk/node-manager/src/main/java/org/mitre/mpf/nms/ChildNodeStateManager.java#L269)):*
> ```Java
>     public static int getMinServiceTimeUpMillis() {
>         return minServiceTimeUpMillis;
>     }
> ```

In NodeDescriptor.java these lines don't seem to be used:

```
        private Address jAddr = null;
```

```
	public int getMinServiceTimeupMillis() {
		return minServiceTimeupMillis;
	}

	public void setMinServiceTimeupMillis(int minServiceTimeupMillis) {
		this.minServiceTimeupMillis = minServiceTimeupMillis;
	}
```

---

*[trunk/node-manager/src/main/java/org/mitre/mpf/nms/MasterNode.java, line 55 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/317#-KxoAXw6JR2-Ea2-EPIp:-KxoAXw6JR2-Ea2-EPIq:b-7xiqr) ([raw file](https://github.com/openmpf/openmpf/blob/f797492998d818e0337536e41666751051c3a929/trunk/node-manager/src/main/java/org/mitre/mpf/nms/MasterNode.java#L55)):*
> ```Java
>     public static MasterNode create() {
>         try (ClassPathXmlApplicationContext context
>                      = new ClassPathXmlApplicationContext("applicationContext.xml", MasterNode.class))  {
> ```

Can you use "classpath:applicationContext.xml" here if you change the location of that file?

---

*[trunk/node-manager/src/main/java/org/mitre/mpf/nms/NodeStatusHttpServer.java, line 126 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/317#-KxoEErQHOZvj20ivFQ3:-KxoEErR4hN4gfJ9ohM5:bm1lcv8) ([raw file](https://github.com/openmpf/openmpf/blob/f797492998d818e0337536e41666751051c3a929/trunk/node-manager/src/main/java/org/mitre/mpf/nms/NodeStatusHttpServer.java#L126)):*
> ```Java
> 		COMPILED_TEMPLATE.execute(writer, viewContext);
> 	}
> }
> ```

This does deserve its own class.

---

*[trunk/node-manager/src/main/resources/org/mitre/mpf/nms/applicationContext.xml, line 42 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/317#-Kxo6mTXPA_Oz08qQ1Jn:-Kxo6mTXPA_Oz08qQ1Jo:b-x7jqjj) ([raw file](https://github.com/openmpf/openmpf/blob/f797492998d818e0337536e41666751051c3a929/trunk/node-manager/src/main/resources/org/mitre/mpf/nms/applicationContext.xml#L42)):*
> ```XML
>     <tx:annotation-driven/>
> 
> 	<context:property-placeholder location="classpath:org/mitre/mpf/nms/nm.properties"/>
> ```

Could applicationContext.xml be located in openmpf/trunk/node-manager/src/main/resources/applicationContext.xml? That would better match the location of openmpf/trunk/workflow-manager/src/main/resources/applicationContext.xml.

Could nm.properties be located in openmpf/trunk/node-manager/src/main/resources/properties/nm.properties? That would better match the location of openmpf/trunk/workflow-manager/src/main/resources/properties/mpf.properties.

---

*[trunk/node-manager/src/main/resources/org/mitre/mpf/nms/nm.properties, line 33 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/317#-Kxo8CoMJ6PSXjJPonMy:-Kxo8CoNfCY4cVAHl59d:b-lc3jz0) ([raw file](https://github.com/openmpf/openmpf/blob/f797492998d818e0337536e41666751051c3a929/trunk/node-manager/src/main/resources/org/mitre/mpf/nms/nm.properties#L33)):*
> ```INI
> mpf.jgroups.channel.name=MPF_Channel
> 
> mpf.this.node=${THIS_MPF_NODE}
> ```

This is a cleaner approach.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/nodeManager/NodeManagerStatus.java, line 66 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/317#-KxoCHmjeUCyjzNdbzk9:-KxoCHmkagvpk6SLcqve:bs8gecw) ([raw file](https://github.com/openmpf/openmpf/blob/f797492998d818e0337536e41666751051c3a929/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/nodeManager/NodeManagerStatus.java#L66)):*
> ```Java
> 	private MasterNode masterNode;
> 
> 	private volatile boolean isRunning = false;
> ```

I don't see why this needs to be `volatile` since it's only ever accessed through the getter, do you?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/317#-:-Kxo2Ktlfbi_qaUPNOkq:bhjh2k5)*

jrobble:
Reviewed 2 of 16 files at r1.
Review status: all files reviewed at latest revision, 11 unresolved discussions.

---

*[trunk/node-manager/src/main/java/org/mitre/mpf/nms/NodeManager.java, line 54 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/317#-KxsBQc0vViYcAN8gz_r:-KxsBQc2Qu_4gVYRO6gG:b3lvkfi) ([raw file](https://github.com/openmpf/openmpf/blob/f797492998d818e0337536e41666751051c3a929/trunk/node-manager/src/main/java/org/mitre/mpf/nms/NodeManager.java#L54)):*
> ```Java
>     public void run() {
>         nodeStateManager.startReceiving(ChannelReceiver.NodeTypes.NodeManager, "NodeManager");
> 	    nodeStatusHttpServer.start();
> ```

Let's conditionalize starting the HTTP server based on a property in the nm.properties file. By default, set it to false. Generally speaking, it's not a good idea to allow access to the status info without some kind of authentication.

We may want to consider an nm-private.properties file (and nm-private.properties example file) that can be used for development purposes where that field is true, much like the respective mpf properties files.

---

*[trunk/node-manager/src/main/java/org/mitre/mpf/nms/NodeManager.java, line 63 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/317#-KxrvawOmiXQLz1c8Kqz:-KxrvawOmiXQLz1c8Kr-:b9yu2sw) ([raw file](https://github.com/openmpf/openmpf/blob/f797492998d818e0337536e41666751051c3a929/trunk/node-manager/src/main/java/org/mitre/mpf/nms/NodeManager.java#L63)):*
> ```Java
> 
>     public static void main(String[] args) {
>         LOG.info("NodeManager Started");
> ```

Say "NodeManager started".

---

*[trunk/node-manager/src/main/java/org/mitre/mpf/nms/NodeManager.java, line 67 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/317#-KxrvLnxTJeMo2HrmgFX:-KxrvHkmZUBurespM4jX:bck4dwm) ([raw file](https://github.com/openmpf/openmpf/blob/f797492998d818e0337536e41666751051c3a929/trunk/node-manager/src/main/java/org/mitre/mpf/nms/NodeManager.java#L67)):*
> ```Java
>         // Log that we are being shutdown, but more hooks are found during process launches in BaseNodeLauncher
>         Runtime.getRuntime().addShutdownHook(new Thread(
>                 () -> LOG.info("Service shutdown")));
> ```

Would it be better to say "NodeManager shutdown" here?

---

*[trunk/node-manager/src/main/java/org/mitre/mpf/nms/NodeStatusHttpServer.java, line 85 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/317#-Kxrxcqq0FFV8C9fyMYP:-KxrxcqrrqrICSBRv_ur:b-1y3bqi) ([raw file](https://github.com/openmpf/openmpf/blob/f797492998d818e0337536e41666751051c3a929/trunk/node-manager/src/main/java/org/mitre/mpf/nms/NodeStatusHttpServer.java#L85)):*
> ```Java
> 			handleHttpShutdownRequest(exchange);
> 			return;
> 		}
> ```

I think we should remove the above "special case" block and the `handleHttpShutdownRequest` method below. It allows users to type this kind of request into their browser:

http://localhost:8008/shutdown?id=localhost.localdomain%3ACaffeDetection%3A1

First, no one's been using that functionality because no one was aware of it.

Second, someone not logged into the system can stop a service, which is bad.

Third, this may be a remnant from when there were start and stop buttons on the status page. "Use back button to return to main page" doesn't make sense anymore.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/317)*

brosenberg42:
Review status: all files reviewed at latest revision, 11 unresolved discussions.

---

*[trunk/node-manager/src/main/java/org/mitre/mpf/nms/ChannelReceiver.java, line 342 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/317#-Kxo3sNHM9YDFDvasw9B:-KxseWScjqWCi27CodaK:b1qqlu5) ([raw file](https://github.com/openmpf/openmpf/blob/c12f9be99700fe20daa8c4b0dd9405d596233f96/trunk/node-manager/src/main/java/org/mitre/mpf/nms/ChannelReceiver.java#L342)):*

Was the msgChannel created in the old constructor ever used?
</blockquote></details>

It doesn't look like it.

---

*[trunk/node-manager/src/main/java/org/mitre/mpf/nms/MasterNode.java, line 55 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/317#-KxoAXw6JR2-Ea2-EPIp:-Kxsch5GAvyzbEHrU9YT:b-lpr1ba) ([raw file](https://github.com/openmpf/openmpf/blob/f797492998d818e0337536e41666751051c3a929/trunk/node-manager/src/main/java/org/mitre/mpf/nms/MasterNode.java#L55)):*

Can you use "classpath:applicationContext.xml" here if you change the location of that file?
</blockquote></details>

If I do that, then this will load WorkflowManager's applicationContext.xml.

---

*[trunk/node-manager/src/main/resources/org/mitre/mpf/nms/applicationContext.xml, line 42 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/317#-Kxo6mTXPA_Oz08qQ1Jn:-Kxsd4IJJ_H3Ba49Syx2:bpx2b1z) ([raw file](https://github.com/openmpf/openmpf/blob/f797492998d818e0337536e41666751051c3a929/trunk/node-manager/src/main/resources/org/mitre/mpf/nms/applicationContext.xml#L42)):*

Could applicationContext.xml be located in openmpf/trunk/node-manager/src/main/resources/applicationContext.xml? That would better match the location of openmpf/trunk/workflow-manager/src/main/resources/applicationContext.xml.

Could nm.properties be located in openmpf/trunk/node-manager/src/main/resources/properties/nm.properties? That would better match the location of openmpf/trunk/workflow-manager/src/main/resources/properties/mpf.properties.
</blockquote></details>

If we want to put applicationContext.xml in the src/main/resources directory, we will need to name the file something different, otherwise MasterNode will load WorkflowManager's applicationContext.xml.

nm.properties can be moved to src/main/resources/properties, but I thought it was better to keep nm.properties and NodeManager's applicationContext.xml together. 

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/nodeManager/NodeManagerStatus.java, line 66 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/317#-KxoCHmjeUCyjzNdbzk9:-KxsefCGeNmwkFt_IpEa:b-uww8qr) ([raw file](https://github.com/openmpf/openmpf/blob/f797492998d818e0337536e41666751051c3a929/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/nodeManager/NodeManagerStatus.java#L66)):*

I don't see why this needs to be `volatile` since it's only ever accessed through the getter, do you?
</blockquote></details>

I'm not exactly sure what you mean.  I don't understand the connection between `volatile` and accessing the field through a getter.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/317)*

jrobble:
Review status: all files reviewed at latest revision, 9 unresolved discussions.

---

*[trunk/node-manager/src/main/resources/org/mitre/mpf/nms/applicationContext.xml, line 42 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/317#-Kxo6mTXPA_Oz08qQ1Jn:-KxspP6He_-HvYgsTmVo:bc41f1x) ([raw file](https://github.com/openmpf/openmpf/blob/f797492998d818e0337536e41666751051c3a929/trunk/node-manager/src/main/resources/org/mitre/mpf/nms/applicationContext.xml#L42)):*
My preference would be to rename the file to applicationContext-nm.xml and move it to openmpf/trunk/node-manager/src/main/resources.

> but I thought it was better to keep nm.properties and NodeManager's applicationContext.xml together

My preference would be to put nm.properties in openmpf/trunk/node-manager/src/main/resources/properties, which is one directory below where applicationContext-nm.xml will be.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/nodeManager/NodeManagerStatus.java, line 66 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/317#-KxoCHmjeUCyjzNdbzk9:-KxsqKZZQeAebNJq1gNi:b55j5ti) ([raw file](https://github.com/openmpf/openmpf/blob/f797492998d818e0337536e41666751051c3a929/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/nodeManager/NodeManagerStatus.java#L66)):*

I'm not exactly sure what you mean.  I don't understand the connection between `volatile` and accessing the field through a getter.
</blockquote></details>

Generally, `volatile` would be used in a situation where you have multiple threads accessing the same variable to ensure that they access the same memory and are always getting the newest value. If `volatile` is not used, then there's the chance that two threads are accessing copies of the variable in their own memory spaces, which would have different values. Refer to [this page](http://tutorials.jenkov.com/java-concurrency/volatile.html).

First, this variable is not accessed in a threaded way. Second, it's only ever accessed through the getter, so the variable is not accessed directly. I don't see any of the typical concerns here that would warrant the need for `volatile`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/317)*

brosenberg42:
Review status: 8 of 16 files reviewed at latest revision, 9 unresolved discussions.

---

*[trunk/node-manager/src/main/java/org/mitre/mpf/nms/ChildNodeStateManager.java, line 269 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/317#-KxoA9RY75UfWh3s0Pca:-KxtEp47WkHPoQYgwqGl:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/c12f9be99700fe20daa8c4b0dd9405d596233f96/trunk/node-manager/src/main/java/org/mitre/mpf/nms/ChildNodeStateManager.java#L269)):*

In NodeDescriptor.java these lines don't seem to be used:

```
        private Address jAddr = null;
```

```
	public int getMinServiceTimeupMillis() {
		return minServiceTimeupMillis;
	}

	public void setMinServiceTimeupMillis(int minServiceTimeupMillis) {
		this.minServiceTimeupMillis = minServiceTimeupMillis;
	}
```
</blockquote></details>

Done.

---

*[trunk/node-manager/src/main/java/org/mitre/mpf/nms/NodeManager.java, line 54 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/317#-KxsBQc0vViYcAN8gz_r:-KxtEqqtwgqgfSnsUuwH:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/f797492998d818e0337536e41666751051c3a929/trunk/node-manager/src/main/java/org/mitre/mpf/nms/NodeManager.java#L54)):*

Let's conditionalize starting the HTTP server based on a property in the nm.properties file. By default, set it to false. Generally speaking, it's not a good idea to allow access to the status info without some kind of authentication.

We may want to consider an nm-private.properties file (and nm-private.properties example file) that can be used for development purposes where that field is true, much like the respective mpf properties files.
</blockquote></details>

Done.

---

*[trunk/node-manager/src/main/java/org/mitre/mpf/nms/NodeManager.java, line 63 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/317#-KxrvawOmiXQLz1c8Kqz:-KxtEXvOeB_ZThTds5xK:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/f797492998d818e0337536e41666751051c3a929/trunk/node-manager/src/main/java/org/mitre/mpf/nms/NodeManager.java#L63)):*

Say "NodeManager started".
</blockquote></details>

Done.

---

*[trunk/node-manager/src/main/java/org/mitre/mpf/nms/NodeManager.java, line 67 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/317#-KxrvLnxTJeMo2HrmgFX:-KxtEb6o5XxcaDOTqOJW:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/f797492998d818e0337536e41666751051c3a929/trunk/node-manager/src/main/java/org/mitre/mpf/nms/NodeManager.java#L67)):*

Would it be better to say "NodeManager shutdown" here?
</blockquote></details>

Done.

---

*[trunk/node-manager/src/main/java/org/mitre/mpf/nms/NodeStatusHttpServer.java, line 85 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/317#-Kxrxcqq0FFV8C9fyMYP:-KxtEd7pxWLEQQ2Ydkv-:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/f797492998d818e0337536e41666751051c3a929/trunk/node-manager/src/main/java/org/mitre/mpf/nms/NodeStatusHttpServer.java#L85)):*

I think we should remove the above "special case" block and the `handleHttpShutdownRequest` method below. It allows users to type this kind of request into their browser:

http://localhost:8008/shutdown?id=localhost.localdomain%3ACaffeDetection%3A1

First, no one's been using that functionality because no one was aware of it.

Second, someone not logged into the system can stop a service, which is bad.

Third, this may be a remnant from when there were start and stop buttons on the status page. "Use back button to return to main page" doesn't make sense anymore.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/317)*

jrobble:
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/nodeManager/NodeManagerStatus.java, line 66 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/317#-KxoCHmjeUCyjzNdbzk9:-KxtNHQS87wisJ0k3lkW:biku6n9) ([raw file](https://github.com/openmpf/openmpf/blob/f797492998d818e0337536e41666751051c3a929/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/nodeManager/NodeManagerStatus.java#L66)):*

Generally, `volatile` would be used in a situation where you have multiple threads accessing the same variable to ensure that they access the same memory and are always getting the newest value. If `volatile` is not used, then there's the chance that two threads are accessing copies of the variable in their own memory spaces, which would have different values. Refer to [this page](http://tutorials.jenkov.com/java-concurrency/volatile.html).

First, this variable is not accessed in a threaded way. Second, it's only ever accessed through the getter, so the variable is not accessed directly. I don't see any of the typical concerns here that would warrant the need for `volatile`.
</blockquote></details>

Do you think `volatile` is necessary here?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/317)*

brosenberg42:
Review status: 8 of 16 files reviewed at latest revision, 1 unresolved discussion.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/nodeManager/NodeManagerStatus.java, line 66 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/317#-KxoCHmjeUCyjzNdbzk9:-KxtVywvmTG8iYx-3gSf:b-4gqb9f) ([raw file](https://github.com/openmpf/openmpf/blob/f797492998d818e0337536e41666751051c3a929/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/nodeManager/NodeManagerStatus.java#L66)):*

Do you think `volatile` is necessary here?
</blockquote></details>

I'm not sure. The way you described volatile in the first paragraph is consistent with my understanding of volatile. The part I don't understand is how the fact that a getter is used means that you don't need volatile. How would a getter prevent the thread from using a cached value?

Based on my understanding of volatile, the fact that you are using a getter has no impact on the program's behavior. The isRunning method is only called by the org.mitre.mpf.wfm.nodeManager.StartUp class.  Startup extends Spring's SmartLifecycle class. I don't know how Spring uses that method, so I'm not sure if volatile is necessary.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/317)*

jrobble:
Reviewed 1 of 17 files at r1, 10 of 10 files at r2.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/nodeManager/NodeManagerStatus.java, line 66 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/317#-KxoCHmjeUCyjzNdbzk9:-KxxBQ-5CSKpx4HBMsQw:b2bzyix) ([raw file](https://github.com/openmpf/openmpf/blob/f797492998d818e0337536e41666751051c3a929/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/nodeManager/NodeManagerStatus.java#L66)):*

I'm not sure. The way you described volatile in the first paragraph is consistent with my understanding of volatile. The part I don't understand is how the fact that a getter is used means that you don't need volatile. How would a getter prevent the thread from using a cached value?

Based on my understanding of volatile, the fact that you are using a getter has no impact on the program's behavior. The isRunning method is only called by the org.mitre.mpf.wfm.nodeManager.StartUp class.  Startup extends Spring's SmartLifecycle class. I don't know how Spring uses that method, so I'm not sure if volatile is necessary.
</blockquote></details>

Generally, in my experience, `volatile` is used as a quick and dirty solution (like using `synchronized`) without really thinking about if it's necessary or not. 

In some cases where it was necessary there have been multiple places in the same class that access the data member directly via callbacks, threads classes, or other means. I don't see any of that here.

That's just been my experience and technically the getter does not make `volatile` unnecessary.

It seems that the use of `volatile` in this case may stem from [this example](https://stackoverflow.com/questions/6258774/i-cant-seem-to-get-an-implementation-of-springs-lifecycle-working), or [this one](https://github.com/spring-projects/spring-integration-samples/blob/master/basic/sftp/src/test/java/org/springframework/integration/samples/sftp/EmbeddedSftpServer.java). Let's leave it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/317)*

PR #325, opened 2017-11-06
 Added support for generic detections.

jrobble:
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/325)


brosenberg42:
Reviewed 16 of 23 files at r1.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/SplitCompletedPredicate.java, line 36 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/325#-KyMPVHA9Q6mVlaLd0pz:-KyMPVHA9Q6mVlaLd0q-:b6zi7m) ([raw file](https://github.com/openmpf/openmpf/blob/760f227627e551e3d1645de416e9373cf6509651/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/SplitCompletedPredicate.java#L36)):*
> ```Java
> 
> /**
>  * This predicate matches when the target exchange's incoming message's {@link MpfHeaders#c} header
> ```

Was this change intentional?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/325)*

jrobble:

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/SplitCompletedPredicate.java, line 36 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/325#-KyMPVHA9Q6mVlaLd0pz:-KyMSRLJbDZ-f9i-wzxw:b9pxlng) ([raw file](https://github.com/openmpf/openmpf/blob/760f227627e551e3d1645de416e9373cf6509651/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/SplitCompletedPredicate.java#L36)):*

Was this change intentional?
</blockquote></details>

No. I will revert it. Good catch.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/325)*

jrobble:
> I think some changes need to be made to org.mitre.mpf.wfm.util.MediaTypeUtils.  If we are now allowing any file type, maybe we don't even need a white list anymore.

I've been thinking about this. We still need the whitelist to map MIME types to data types. For example:

```
whitelist.image/gif=VIDEO
```

However, we may no longer need to use the whitelist for filtering out what the user can and cannot upload through the UI. I've been testing this set of PRs using the REST API, which allows me to select any file type.

I see two approaches:

1. The user has to whitelist each MIME type that they care about. For example:

```
whitelist.application/pdf=UNKNOWN
```

2. Every file is allowed and set to UKNOWN if the MIME type doesn't correspond to a video, image, or audio file.

The first approach has the advantage of preventing some file type uploads, but if the user has something like a virus checker component, then they probably want to upload every file type, so the first approach is very inconvenient.  For that reason I'm leaning towards the second approach.

Do you have a preference?

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/325#-:-KyMTFJz8D4QHKS1xNJL:b-2algvk)*

brosenberg42:
I think the second approach is better.  
With the first approach, when a user adds a Generic component they would need to know all the types it processes and manually change mediaTypes.properties.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/325#-:-KyMYVnnRdZdpyCWdlkY:b-sysurb)*

jrobble:
I will create a separate issue for this.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/325#-:-KyM_Nc6vJNO4D5s9NbV:b-s6sxju)*

brosenberg42:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/325#-:-KyMgkk2aRw0j3VYeVrA:bnfp4nl)*

PR #332, opened 2017-11-08
Add serviceName field to AlgorithmDefinition

brosenberg42:
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/332)

### DIFF
--- a/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/pipeline/xml/AlgorithmDefinition.java
+++ b/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/pipeline/xml/AlgorithmDefinition.java
@@ -45,6 +45,10 @@ public class AlgorithmDefinition {
     private String name;
 	public String getName() { return name; }
 
+	@XStreamAsAttribute
+	private String serviceName;
+	public String getServiceName() { return serviceName; }
+
 	/** The REQUIRED description of this algorithm. */
     private String description;
 	public String getDescription() { return description; }
@@ -79,10 +83,11 @@ public class AlgorithmDefinition {
 	 * @param name The REQUIRED non-null name of the algorithm. This is preprocessed using {@link org.mitre.mpf.wfm.util.TextUtils#trimAndUpper(String)}.
 	 * @param description The REQUIRED non-null description of the algorithm. This is preprocessed using {@link org.mitre.mpf.wfm.util.TextUtils#trim(String)}.
 	 */
-    public AlgorithmDefinition(ActionType actionType, String name, String description,
+    public AlgorithmDefinition(ActionType actionType, String algorithmName, String serviceName, String description,
                                boolean supportsBatchProcessing, boolean supportsStreamProcessing) {
         this.actionType = actionType;
-	    this.name = TextUtils.trimAndUpper(name);
+	    this.name = TextUtils.trimAndUpper(algorithmName);
+	    this.serviceName = TextUtils.trim(serviceName);
         this.description = TextUtils.trim(description);
 	    this.providesCollection = new ProvidesCollection();
 	    this.requiresCollection = new RequiresCollection();
--- a/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/service/component/AddComponentServiceImpl.java
+++ b/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/service/component/AddComponentServiceImpl.java
@@ -181,15 +181,20 @@ public class AddComponentServiceImpl implements AddComponentService {
         customPipelineValidator.validate(descriptor);
 
         AlgorithmDefinition algorithmDef = null;
-        String algoName = null;
         if (descriptor.algorithm != null) {
             algorithmDef = convertJsonAlgo(descriptor);
-            algoName = saveAlgorithm(algorithmDef);
-            model.setAlgorithmName(algoName);
+            String serviceName = saveService(descriptor, algorithmDef);
+            model.setServiceName(serviceName);
         }
+
         model.setComponentName(descriptor.componentName);
 
         try {
+        	if (algorithmDef != null) {
+                String algoName = saveAlgorithm(algorithmDef);
+                model.setAlgorithmName(algoName);
+            }
+
             Set<String> savedActions = saveActions(descriptor, algorithmDef);
             model.getActions().addAll(savedActions);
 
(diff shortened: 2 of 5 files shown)


jrobble:
Please explain the motivation for this update.

It seems that you're still using the component name as the service name. As far as the node manager is concerned, it uses a [FQN](https://github.com/openmpf/openmpf/blob/0992ddaca607ccddad499f8db8c3420279ba0b89/trunk/node-manager/src/main/java/org/mitre/mpf/nms/ServiceDescriptor.java#L75) to keep track of the service descriptors. 

Do you need to do something different for streaming?

---

Review status: 0 of 5 files reviewed at latest revision, 2 unresolved discussions.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/pipeline/xml/AlgorithmDefinition.java, line 86 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/332#-KySWPVkU-XFl4eXgRw6:-KySWPVkU-XFl4eXgRw7:bxsd7zz) ([raw file](https://github.com/openmpf/openmpf/blob/d79976372de9d53fb8764ccc3bcaa5c4abfe2287/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/pipeline/xml/AlgorithmDefinition.java#L86)):*
> ```Java
> 	 * @param description The REQUIRED non-null description of the algorithm. This is preprocessed using {@link org.mitre.mpf.wfm.util.TextUtils#trim(String)}.
> 	 */
>     public AlgorithmDefinition(ActionType actionType, String algorithmName, String serviceName, String description,
> ```

Update method comment above.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/service/component/AddComponentServiceImpl.java, line 186 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/332#-KySX5S__RtGQcCuhLNH:-KySX5S__RtGQcCuhLNI:bjmfmnx) ([raw file](https://github.com/openmpf/openmpf/blob/d79976372de9d53fb8764ccc3bcaa5c4abfe2287/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/service/component/AddComponentServiceImpl.java#L186)):*
> ```Java
>         if (descriptor.algorithm != null) {
>             algorithmDef = convertJsonAlgo(descriptor);
>             String serviceName = saveService(descriptor, algorithmDef);
> ```

Shouldn't we wait to add the service until we know that the algorithms, actions, tasks, and pipelines were all saved successfully?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/332#-:-KySXi_rzQquI7TVjva0:b-pr6fzf)*

brosenberg42:
Right now AFAIK, there is no way to determine which service runs a particular algorithm, or vice-versa. If there is some way to do that now, then I don't need to make this change.
For streaming, given a TransientStreamingJob, I need to tell the node manager which services to start.

---

Review status: 0 of 5 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/332#-:-KySaChoPElJYoopsV6V:b-lim0lf)*

brosenberg42:
Review status: 0 of 5 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/service/component/AddComponentServiceImpl.java, line 186 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/332#-KySX5S__RtGQcCuhLNH:-KySd4av0L69SdcHvtAd:b-97tb19) ([raw file](https://github.com/openmpf/openmpf/blob/d79976372de9d53fb8764ccc3bcaa5c4abfe2287/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/service/component/AddComponentServiceImpl.java#L186)):*

Shouldn't we wait to add the service until we know that the algorithms, actions, tasks, and pipelines were all saved successfully?
</blockquote></details>

Why would that order be better? The order of the saving algorithms, actions, tasks, and pipelines is based on which objects reference each other. I save the algorithm before the action because the action references the algorithm. Now, the algorithm references the service, so I save the service before the algorithm. If saving anything fails, the service will also be removed by the call to `removeComponentService.deleteCustomPipelines(model, true);`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/332)*

jrobble:
I see. For batch processing the node manager doesn't need to know what each services does, but for streaming processing the node manager does need to know.

Looking at `TransientAction`, I see that all we have to work with is the name of the algorithm with no way to get the service name.

The service names are listed in nodeManagerPalette.json, but that file doesn't list the associated algorithms. Rather than add the service name to Algorithms.xml, I think it's better to list the name of the algorithm in nodeManagerPalette.json for two reasons:

<br/>

1. Conceptually, the service executes the algorithm, so the relationship is service -> algorithm. The algorithm is part of a service, but can conceptually stand on its own as a simpler entity.

<br/>

2. If we make a change to Algorithms.xml, we will need to apply that change to the existing Algorithms.xml file when we do an OpenMPF upgrade. While the user will most likely select the option to re-register the components in the deployment package, if there are pre-existing third-party components then those entries in the Algorithms.xml file will need to be updated with the service name field. I'd prefer we avoid that.

I see the following in NodeController.java:

```
private Map<String, ServiceModel> nodeManagerPaletteMap = null;
```

That data member maps the service name to the `ServiceModel`. If we take the second approach I described above, the `ServiceModel` will contain the algorithm name. You can either create another map that maps the algorithm name to the `ServiceModel` (which contains the service name), or create a method that iterates over all the `ServiceModel` objects looking for one with the algorithm name.

---

Review status: 0 of 5 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/service/component/AddComponentServiceImpl.java, line 186 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/332#-KySX5S__RtGQcCuhLNH:-KyTc6UuDkWP4cDj9707:b-p1l49q) ([raw file](https://github.com/openmpf/openmpf/blob/d79976372de9d53fb8764ccc3bcaa5c4abfe2287/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/service/component/AddComponentServiceImpl.java#L186)):*
> Now, the algorithm references the service, so I save the service before the algorithm.

That makes sense based on your new code. Conceptually, I think the service should reference the algorithm.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/332#-:-KySegtAQHbjMS2YcsEO:b-m2vhkb)*

PR #365, opened 2017-12-05
Frame extraction

jrobble:
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/365)


jrobble:
Overcome by https://github.com/openmpf/openmpf/pull/387.

PR #370, opened 2017-12-07
Streaming protobufs

clnowacki:
Added the segment summary response message to the detection.proto file. Updated the streaming MessageUtils to pack and unpack the protobufs in the segment summary message. Added a test to the MessageUtils gtests.

---
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/370)


clnowacki:
Compared with the wrong branch

PR #373, opened 2017-12-08
Updated node-manager-only-pom.xml.

jrobble:
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/373)

### DIFF
--- a/trunk/mpf-install/node-manager-only-pom.xml
+++ b/trunk/mpf-install/node-manager-only-pom.xml
@@ -39,6 +39,11 @@
     <packaging>pom</packaging>
     <name>MPF Install (Node Manager Only)</name>
 
+    <!-- poms in the following directories -->
+    <modules>
+        <module>../node-manager</module>
+    </modules>
+
     <!-- our jar modules and their dependencies: the one found first wins even over a newer version -->
     <dependencies>
         <dependency>
--- a/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/transients/TransientJob.java
+++ b/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/transients/TransientJob.java
@@ -77,13 +77,13 @@ public class TransientJob {
 	private String callbackMethod;
 	public String getCallbackMethod() { return callbackMethod; }
 
-	public TransientJob(@JsonProperty("id") long id,
-						@JsonProperty("externalId") String externalId,
-						@JsonProperty("pipeline") TransientPipeline pipeline,
-						@JsonProperty("currentStage") int currentStage,
-						@JsonProperty("priority") int priority,
-						@JsonProperty("outputEnabled") boolean outputEnabled,
-						@JsonProperty("cancelled") boolean cancelled) {
+	public TransientJob(long id,
+						String externalId,
+						TransientPipeline pipeline,
+						int currentStage,
+						int priority,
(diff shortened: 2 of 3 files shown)


jrobble:
Review status: 0 of 3 files reviewed at latest revision, 3 unresolved discussions.

---

*[trunk/mpf-install/node-manager-only-pom.xml, line 44 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/373#-L-tGGSUF5jXpAhNFFPY:-L-tGGSUF5jXpAhNFFPZ:b2g60ha) ([raw file](https://github.com/openmpf/openmpf/blob/df19d19649e0cf02a7930ed83712c12a8acde82d/trunk/mpf-install/node-manager-only-pom.xml#L44)):*
> ```XML
>     
>     <modules>
>         <module>../node-manager</module>
> ```

During the first part of the Jenkins build, where only the node-manager is built and installed, maven was not able to find the node-manager pom. This solves that problem.

Note that the problem only happened once I completely blew away the .m2 directory on Jenkins 1.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/transients/TransientJob.java, line 80 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/373#-L-tGkuH7KuLozjamzog:-L-tGkuH7KuLozjamzoh:b86iu25) ([raw file](https://github.com/openmpf/openmpf/blob/9a9061aeea57a6d96d55e2501e598d4b714a8f86/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/transients/TransientJob.java#L80)):*
> ```Java
> 	public String getCallbackMethod() { return callbackMethod; }
> 
> 	public TransientJob(@JsonProperty("id") long id,
> ```

These `@JsonProperty` annotations were why some builds were failing on Jenkins 1. Don't ask me why the builds passed on Jenkins 2 and Jenkins 3. I don't know.

Specifically, the exception was:

```
Caused by: com.fasterxml.jackson.databind.JsonMappingException: Could not find creator property with name 'callbackURL' (in class org.mitre.mpf.wfm.data.entities.transients.TransientJob)
```

It seems that Jackson got confused and tried to use this constructor to initialize a `TransientJob` instead of the constructor below, which should have been used since it's the one with `@JsonCreator`. Once I removed these `@JsonProperty` annotations the build passed on Jenkins 1. See openmpf-master-github-with-pull-request build # 32.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/JsonUtils.java, line 81 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/373#-L-tGanW_jRr4cLTPohG:-L-tGanW_jRr4cLTPohH:b4mtun5) ([raw file](https://github.com/openmpf/openmpf/blob/df19d19649e0cf02a7930ed83712c12a8acde82d/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/JsonUtils.java#L81)):*
> ```Java
> 			return smileObjectMapper.readValue(json, targetClass);
> 		} catch(IOException ioe) {
> 			throw new WfmProcessingException(String.format("Failed to deserialize instance of '%s': %s", targetClass.getSimpleName(), ioe.getMessage()), ioe);
> ```

The `WfmProcessingException` was hiding the real exception, so I made these updates.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/373)*

Issue #391, opened 2017-12-22
Use long for frame id instead of int


jrobble:
It's possible that a user may connect to a streaming video that's been running for over a year. If the frame ids are obtained from the stream, or calculated based on the stream start time, then they will be very large numbers. Thus, 'long' is a more suitable data type than `int`.

This task requires updating both the Java and C++ code to use `long`. This includes the APIs, components, WFM, protobufs, etc.

PR #401, opened 2018-01-10
Streaming component executor

brosenberg42:
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/401)


jrobble:
Reviewed 40 of 46 files at r1, 7 of 8 files at r2.
Review status: 41 of 47 files reviewed at latest revision, 28 unresolved discussions.

---

*[trunk/detection/executor/cpp/streaming/BasicAmqMessageSender.cpp, line 83 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/401#-L2a2Ujim3ucssGtvf_C:-L2a2UjjNizeDu7c2ZmY:b1j6bad) ([raw file](https://github.com/openmpf/openmpf/blob/b0190b5aa91721935a8da2dbd1c0cf6021a44ba9/trunk/detection/executor/cpp/streaming/BasicAmqMessageSender.cpp#L83)):*
> ```cpp
>         message->setLongProperty("JOB_ID", job_id_);
>         message->setIntProperty("FRAME_NUMBER", frame_number);
>         message->setLongProperty("ACTIVITY_DETECT_TIME", GetTimestampMillis());
> ```

Let's call this ACTIVITY_DETECTION_TIMESTAMP.

Please make the same change in MPFAMQMessage.h.

---

*[trunk/detection/executor/cpp/streaming/BasicAmqMessageSender.cpp, line 97 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/401#-L2a3SJMAQcbXywivA8m:-L2a3SJNZmsvtgVmL712:bux97rb) ([raw file](https://github.com/openmpf/openmpf/blob/b0190b5aa91721935a8da2dbd1c0cf6021a44ba9/trunk/detection/executor/cpp/streaming/BasicAmqMessageSender.cpp#L97)):*
> ```cpp
>                                                 const std::string &error_message,
>                                                 const std::string &detection_type) {
>         SendSegmentReport(frame_number, detection_type, {}, error_message);
> ```

When an error occurs, should we also pass along any tracks that have been generated so far?

---

*[trunk/detection/executor/cpp/streaming/BasicAmqMessageSender.cpp, line 101 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/401#-L2a0wPozvMrO3bFhIdR:-L2a0wPpIjiIrUgeXGAh:b-hncsq9) ([raw file](https://github.com/openmpf/openmpf/blob/b0190b5aa91721935a8da2dbd1c0cf6021a44ba9/trunk/detection/executor/cpp/streaming/BasicAmqMessageSender.cpp#L101)):*
> ```cpp
> 
> 
>     void BasicAmqMessageSender::SendSegmentReport(int frame_number,
> ```

Can we also call this `SendSummaryReport` and use function overloading?

---

*[trunk/detection/executor/cpp/streaming/BasicAmqMessageSender.h, line 45 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/401#-L2a8-y96LcSXaQHkfC4:-L2a8-yAZLpu5YRGXtTF:b-50iuqp) ([raw file](https://github.com/openmpf/openmpf/blob/b0190b5aa91721935a8da2dbd1c0cf6021a44ba9/trunk/detection/executor/cpp/streaming/BasicAmqMessageSender.h#L45)):*
> ```objc
> namespace MPF { namespace COMPONENT {
> 
>     // TODO: Combine with AMQMessenger when adding support for multistage pipelines.
> ```

It seems like some of the converters in MPFAMQMessage.h can be used in the reduced scope implementation. Have you considered using them?

---

*[trunk/detection/executor/cpp/streaming/BasicAmqMessageSender.h, line 46 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/401#-L2a60PvsDSHJS-EEQQI:-L2a60PvsDSHJS-EEQQJ:b-cgth9k) ([raw file](https://github.com/openmpf/openmpf/blob/b0190b5aa91721935a8da2dbd1c0cf6021a44ba9/trunk/detection/executor/cpp/streaming/BasicAmqMessageSender.h#L46)):*
> ```objc
> 
>     // TODO: Combine with AMQMessenger when adding support for multistage pipelines.
>     class BasicAmqMessageSender {
> ```

I'm comparing the name of this class against MPFAMQMessage.h. There, "AMQ" is capitalized. I think the "Amq" is easier to read. How about we rename the existing classes with "AMQ" in them to "Amq"? That will avoid the issue where "MPFAMQ" looks like one jumble of letters.

---

*[trunk/detection/executor/cpp/streaming/ExecutorErrors.h, line 28 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/401#-L2a8oSZan5P22vJa1zD:-L2a8oSZan5P22vJa1zE:b7fvyjc) ([raw file](https://github.com/openmpf/openmpf/blob/b0190b5aa91721935a8da2dbd1c0cf6021a44ba9/trunk/detection/executor/cpp/streaming/ExecutorErrors.h#L28)):*
> ```objc
> 
> 
> #ifndef MPF_EXECUTORERRORS_H
> ```

My preference with these defines is to put an underscore between all the words. How do you feel about that?

---

*[trunk/detection/executor/cpp/streaming/ExecutorErrors.h, line 35 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/401#-L2a9el3ZgVMtSDjjQBt:-L2a9el3ZgVMtSDjjQBu:b234ee7) ([raw file](https://github.com/openmpf/openmpf/blob/b0190b5aa91721935a8da2dbd1c0cf6021a44ba9/trunk/detection/executor/cpp/streaming/ExecutorErrors.h#L35)):*
> ```objc
> namespace MPF { namespace COMPONENT {
> 
>     enum class ExitCode {
> ```

In other classes the enums are in all caps. Referring to the [style guide](https://google.github.io/styleguide/cppguide.html#Enumerator_Names), they should be named like constants. Since we aren't using the "k" prefix for that, we should capitalize our enums.

---

*[trunk/detection/executor/cpp/streaming/main.cpp, line 99 at r2](https://reviewable.io:443/reviews/openmpf/openmpf/401#-L2bEUkYRCNTWa5aFgzR:-L2bEUkZ09paCXGm9nlU:b-sunczx) ([raw file](https://github.com/openmpf/openmpf/blob/132c2e7d16a6274794def58f4d3c9cc06a7bf0aa/trunk/detection/executor/cpp/streaming/main.cpp#L99)):*
> ```cpp
> 
> 
> ExitCode run_job(const JobSettings &settings, const std::string &app_dir, log4cxx::LoggerPtr &logger) {
> ```

Let's pass the logger first to be consistent with other functions in the C++ code.

---

*[trunk/detection/executor/cpp/streaming/main.cpp, line 125 at r2](https://reviewable.io:443/reviews/openmpf/openmpf/401#-L2bG62eHsWRXUYHG7Tk:-L2bG62eHsWRXUYHG7Tl:bippfvl) ([raw file](https://github.com/openmpf/openmpf/blob/132c2e7d16a6274794def58f4d3c9cc06a7bf0aa/trunk/detection/executor/cpp/streaming/main.cpp#L125)):*
> ```cpp
>         bool read_failed = false;
> 
>         while (!quit_watcher->IsTimeToQuit()) {
> ```

Should we also check read_failed in this conditional?

---

*[trunk/detection/executor/cpp/streaming/main.cpp, line 128 at r2](https://reviewable.io:443/reviews/openmpf/openmpf/401#-L2bFODb4PmwELnO3YU6:-L2bFODb4PmwELnO3YU7:b-6d7z25) ([raw file](https://github.com/openmpf/openmpf/blob/132c2e7d16a6274794def58f4d3c9cc06a7bf0aa/trunk/detection/executor/cpp/streaming/main.cpp#L128)):*
> ```cpp
>             cv::Mat frame;
>             if (!video_capture.read(frame)) {
>                 // TODO: Detect and report stalls.
> ```

I created #405 for this.

---

*[trunk/detection/executor/cpp/streaming/main.cpp, line 185 at r2](https://reviewable.io:443/reviews/openmpf/openmpf/401#-L2bBfKRCk2lmcaLz9Tt:-L2bBfKTejzZHN1YJP3s:bfupwbw) ([raw file](https://github.com/openmpf/openmpf/blob/132c2e7d16a6274794def58f4d3c9cc06a7bf0aa/trunk/detection/executor/cpp/streaming/main.cpp#L185)):*
> ```cpp
>     std::string exe_path(buffer);
>     size_t last_slash_pos = exe_path.rfind('/');
>     if (last_slash_pos != std::string::npos) {
> ```

Is there ever a case where the path won't at least start with a leading slash (corresponding to the root directory)?

---

*[trunk/detection/executor/cpp/streaming/main.cpp, line 197 at r2](https://reviewable.io:443/reviews/openmpf/openmpf/401#-L2bGcSGVu1-MX8Ye-bx:-L2bGcSGVu1-MX8Ye-by:b6002ek) ([raw file](https://github.com/openmpf/openmpf/blob/132c2e7d16a6274794def58f4d3c9cc06a7bf0aa/trunk/detection/executor/cpp/streaming/main.cpp#L197)):*
> ```cpp
>     if (logger->getAllAppenders().empty()) {
>         log4cxx::BasicConfigurator::configure();
>         LOG4CXX_WARN(logger, "Unable to locate StreamingExecutorLog4cxxConfig.xml. Logging to standard out instead.")
> ```

Please create a variable for "StreamingExecutorLog4cxxConfig.xml" so you don't need to hardcode the name twice. 

---

*[trunk/detection/executor/cpp/streaming/QuitWatcher.cpp, line 63 at r2](https://reviewable.io:443/reviews/openmpf/openmpf/401#-L2bHn_OKWx5Of6ZWX_z:-L2bHn_P5t6Z9CQDVelW:b-q1xtju) ([raw file](https://github.com/openmpf/openmpf/blob/132c2e7d16a6274794def58f4d3c9cc06a7bf0aa/trunk/detection/executor/cpp/streaming/QuitWatcher.cpp#L63)):*
> ```cpp
>             std::string line;
>             while (std::getline(std::cin, line)) {
>                 if (line == "quit") {
> ```

Thinking about the future, how will this class be used when we support the "pause" and "resume" commands? Will it be rolled into another class? It may be worth putting a TODO in here explaining what you have in mind.

---

*[trunk/detection/executor/cpp/streaming/StreamingExecutorLog4cxxConfig.xml, line 55 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/401#-L2awlgL0hav7v04TNRB:-L2awlgL0hav7v04TNRC:bnmao41) ([raw file](https://github.com/openmpf/openmpf/blob/b0190b5aa91721935a8da2dbd1c0cf6021a44ba9/trunk/detection/executor/cpp/streaming/StreamingExecutorLog4cxxConfig.xml#L55)):*
> ```XML
>     <param name="DatePattern" value="'.'yyyy-MM-dd" />
>     <layout class="org.apache.log4j.PatternLayout">
>       <param name="ConversionPattern" value="%d %p %c{36}:%L - %m%n" />
> ```

In the batch version of this file, this line is:

`<param name="ConversionPattern" value="%d %p [%t] %c{36}:%L - %m%n" />`

Did  you intentionally remove `[%t]`?

---

*[trunk/detection/executor/cpp/streaming/FrameReader/main.cpp, line 193 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/401#-L2a7EvTTgNfvIuMLmp0:-L2a7EvTTgNfvIuMLmp1:b-1y27qv) ([raw file](https://github.com/openmpf/openmpf/blob/b0190b5aa91721935a8da2dbd1c0cf6021a44ba9/trunk/detection/executor/cpp/streaming/FrameReader/main.cpp#L193)):*
> ```cpp
> 
>     try {
>         MPFJobStatusMessage msg(job_name,job_id, "IN_PROGRESS", 1234);
> ```

You provided a random value for the status change time. Shouldn't that be the system time?

---

*[trunk/detection/executor/cpp/streaming/MessageUtils/MPFAMQMessenger.h, line 43 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/401#-L2a6cMtZqrli1f1iTc1:-L2a6cMugqpn3JuP9XPw:b-gh6438) ([raw file](https://github.com/openmpf/openmpf/blob/b0190b5aa91721935a8da2dbd1c0cf6021a44ba9/trunk/detection/executor/cpp/streaming/MessageUtils/MPFAMQMessenger.h#L43)):*
> ```objc
> 
> // TODO: For future use.
> // TODO: Refactor to use RAII
> ```

What is RAII? Please define in comment.

---

*[trunk/detection/test-components/cpp/CplusplusHelloWorld/plugin-files/descriptor/descriptor.json, line 10 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/401#-L2av2X9vlhWVYggA3uZ:-L2av2XA2dbdbrjAzt9x:b-8pv4gu) ([raw file](https://github.com/openmpf/openmpf/blob/b0190b5aa91721935a8da2dbd1c0cf6021a44ba9/trunk/detection/test-components/cpp/CplusplusHelloWorld/plugin-files/descriptor/descriptor.json#L10)):*
> ```JSON
>   "environmentVariables" : [{
>       "name":"LD_LIBRARY_PATH",
>       "value":"${MPF_HOME}/plugins/CplusplusHelloWorld/lib:${LD_LIBRARY_PATH}"
> ```

Is this a required change or personal preference?

---

*[trunk/node-manager/src/main/java/org/mitre/mpf/nms/EnvironmentVariableExpander.java, line 37 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/401#-L2ahqDCUBiCglJADgix:-L2ahqDCUBiCglJADgiy:b65hnez) ([raw file](https://github.com/openmpf/openmpf/blob/b0190b5aa91721935a8da2dbd1c0cf6021a44ba9/trunk/node-manager/src/main/java/org/mitre/mpf/nms/EnvironmentVariableExpander.java#L37)):*
> ```Java
> import static java.util.stream.Collectors.toMap;
> 
> public class EnvironmentVariableExpander {
> ```

I feel like this should go in a "util" directory.

---

*[trunk/node-manager/src/main/java/org/mitre/mpf/nms/streaming/IniManager.java, line 119 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/401#-L2aie0FVCnlaLfAXgJ5:-L2aie0FVCnlaLfAXgJ6:b70kbe4) ([raw file](https://github.com/openmpf/openmpf/blob/b0190b5aa91721935a8da2dbd1c0cf6021a44ba9/trunk/node-manager/src/main/java/org/mitre/mpf/nms/streaming/IniManager.java#L119)):*
> ```Java
> 
> 
> 	private static Ini newIni() {
> ```

This tiny method seems a little overkill because it's currently used in the one method above, but I can see how it may be more useful in the future in the methods below. Please update those comments to use this method so we don't forget later on.

---

*[trunk/node-manager/src/main/java/org/mitre/mpf/nms/streaming/StreamingJob.java, line 74 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/401#-L2ahNKdw1Tr7srhRCOe:-L2ahNKdw1Tr7srhRCOf:b-gsueky) ([raw file](https://github.com/openmpf/openmpf/blob/b0190b5aa91721935a8da2dbd1c0cf6021a44ba9/trunk/node-manager/src/main/java/org/mitre/mpf/nms/streaming/StreamingJob.java#L74)):*
> ```Java
> 	}
> 
> 	private void onAllProcessesExit(Throwable error) {
> ```

Shouldn't we at least log this error?

---

*[trunk/node-manager/src/main/java/org/mitre/mpf/nms/streaming/StreamingProcess.java, line 69 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/401#-L2agmX_k4v5nTaeTFwS:-L2agmX_k4v5nTaeTFwT:b5fabs1) ([raw file](https://github.com/openmpf/openmpf/blob/b0190b5aa91721935a8da2dbd1c0cf6021a44ba9/trunk/node-manager/src/main/java/org/mitre/mpf/nms/streaming/StreamingProcess.java#L69)):*
> ```Java
> 	private void restartUntilDone(Process originalProcess) {
> 		int exitCode = awaitExit(originalProcess);
> 		if (exitCode == 0) {
> ```

Please add a comment here explaining that this method will return normally if the process was cancelled by the user. Or, just check against `StreamingProcessExitReason.CANCELLED`.

---

*[trunk/node-manager/src/main/java/org/mitre/mpf/nms/streaming/StreamingProcess.java, line 83 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/401#-L2ah6xaj9EjkRj4Qp-B:-L2ah6xbPjLhyMfYY0_G:b5fabs1) ([raw file](https://github.com/openmpf/openmpf/blob/b0190b5aa91721935a8da2dbd1c0cf6021a44ba9/trunk/node-manager/src/main/java/org/mitre/mpf/nms/streaming/StreamingProcess.java#L83)):*
> ```Java
> 			}
> 			exitCode = awaitExit(restartedProcess);
> 			if (exitCode == 0) {
> ```

Please add a comment here explaining that this method will return normally if the process was cancelled by the user. Or, just check against `StreamingProcessExitReason.CANCELLED`.

---

*[trunk/node-manager/src/main/java/org/mitre/mpf/nms/streaming/StreamingProcess.java, line 127 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/401#-L2agabzmA_mhoMV-LaP:-L2agac-huQd8Gr1mMRN:b-ie4eyf) ([raw file](https://github.com/openmpf/openmpf/blob/b0190b5aa91721935a8da2dbd1c0cf6021a44ba9/trunk/node-manager/src/main/java/org/mitre/mpf/nms/streaming/StreamingProcess.java#L127)):*
> ```Java
> 
> 			LOG.info("Process: {} exited with exit code {}", _executable, exitCode);
> 			if (exitCode == StreamingProcessExitReason.STREAM_STALLED.exitCode) {
> ```

Please add a comment here stating that you're throwing an exception when the process terminates due to a stall to avoid trying to restart the process.

---

*[trunk/node-manager/src/main/resources/properties/nm.properties, line 38 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/401#-L2a_UHU7n6bwKST_GJa:-L2a_UHU7n6bwKST_GJb:buarun1) ([raw file](https://github.com/openmpf/openmpf/blob/b0190b5aa91721935a8da2dbd1c0cf6021a44ba9/trunk/node-manager/src/main/resources/properties/nm.properties#L38)):*
> ```INI
> 
> mpf.streaming.job.ini.dir=${MPF_HOME}/tmp/streaming-ini-files
> mpf.streaming.process.max.restarts=0
> ```

These properties start with "mpf.streaming" while the ones below start with just "streaming". Let's remove "mpf." from these.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/businessrules/StreamingJobRequestBo.java, line 89 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/401#-L2YS9LAc1kyLIKzDRKL:-L2YS9LBTy5mLbiSDbTx:b-4zfcvu) ([raw file](https://github.com/openmpf/openmpf/blob/b0190b5aa91721935a8da2dbd1c0cf6021a44ba9/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/businessrules/StreamingJobRequestBo.java#L89)):*
> ```Java
> 	void handleJobStatusChange(long jobId, JobStatus status, long timestamp);
> 
> 	void handleNewActivityAlert(long jobId, int frameId, long timestamp);
> ```

Is the `int` here just temporary until you address #391?

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/routes/StreamingJobRoutesBuilder.java, line 106 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/401#-L2aVqqTJZZyxpV8ZMr_:-L2aVqqUj6vSP9-5pJyb:b9k010z) ([raw file](https://github.com/openmpf/openmpf/blob/b0190b5aa91721935a8da2dbd1c0cf6021a44ba9/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/routes/StreamingJobRoutesBuilder.java#L106)):*
> ```Java
> 
> 
> 	private static SegmentSummaryReport convertProtobufResponse(
> ```

We could put the following methods in a *Processor.java class, but I guess it's not necessary.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/RedisImpl.java, line 804 at r2](https://reviewable.io:443/reviews/openmpf/openmpf/401#-L2bLSXDwbew1xd4hXlK:-L2bLSXFEccJ5h2eBDhy:buerlbj) ([raw file](https://github.com/openmpf/openmpf/blob/132c2e7d16a6274794def58f4d3c9cc06a7bf0aa/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/RedisImpl.java#L804)):*
> ```Java
>      * @exception WfmProcessingException is thrown if one of the streaming jobs listed in jobIds isn't in REDIS (i.e. it is not an active job).
>      */
>     @Override
> ```

A lot of methods in here should have `@Override` shouldn't they?

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/nodeManager/NodeManagerStatus.java, line 311 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/401#-L2aLoRfhwdefB_jLMqM:-L2aLoRh3GRpOoT0MI5E:b-nza3ac) ([raw file](https://github.com/openmpf/openmpf/blob/b0190b5aa91721935a8da2dbd1c0cf6021a44ba9/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/nodeManager/NodeManagerStatus.java#L311)):*
> ```Java
> 				break;
> 			default:
> 				status = JobStatus.ERROR;
> ```

I'd prefer that we explicitly check for ERROR. The default should also set the status to JobStatus.ERROR and log a message about unexpected exit reason.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/401)*

jrobble:
I still need to verify that the tests pass on Jenkins and run this code locally.

---

Reviewed 5 of 46 files at r1, 1 of 8 files at r2.
Review status: all files reviewed at latest revision, 27 unresolved discussions.

---

*[trunk/detection/executor/cpp/streaming/BasicAmqMessageSender.cpp, line 97 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/401#-L2a3SJMAQcbXywivA8m:-L2bQFEFBh7iVaOVUtWx:b-3fvibo) ([raw file](https://github.com/openmpf/openmpf/blob/b0190b5aa91721935a8da2dbd1c0cf6021a44ba9/trunk/detection/executor/cpp/streaming/BasicAmqMessageSender.cpp#L97)):*

When an error occurs, should we also pass along any tracks that have been generated so far?
</blockquote></details>

If we're not passing along the tracks, would it make more sense to use `SendJobStatus()` over `SendSegmentReport()` here? One disadvantage is that we won't be passing along the segment number; however we could log it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/401#-:-L2bPgAGHZVEolLz0TQR:b-95eq28)*

jrobble:
Review status: all files reviewed at latest revision, 28 unresolved discussions.

---

*[trunk/detection/executor/cpp/streaming/main.cpp, line 145 at r2](https://reviewable.io:443/reviews/openmpf/openmpf/401#-L2cosEIN4RN4adaaluh:-L2cosEJ08po5mTwdOhu:b-owfww6) ([raw file](https://github.com/openmpf/openmpf/blob/132c2e7d16a6274794def58f4d3c9cc06a7bf0aa/trunk/detection/executor/cpp/streaming/main.cpp#L145)):*
> ```cpp
>             if (activity_found) {
>                 LOG4CXX_DEBUG(logger, log_prefix << "Sending new activity alert for frame: " << frame_number)
>                 sender.SendActivityAlert(frame_number);
> ```

A new activity alert should only be sent once per segment.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/401)*

jrobble:
Reviewed 29 of 29 files at r3.
Review status: all files reviewed at latest revision, 14 unresolved discussions.

---

*[trunk/detection/executor/cpp/streaming/BasicAmqMessageSender.h, line 46 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/401#-L2a60PvsDSHJS-EEQQI:-L2g9JOqahsbIwlQjBq3:bvq9cxa) ([raw file](https://github.com/openmpf/openmpf/blob/b0190b5aa91721935a8da2dbd1c0cf6021a44ba9/trunk/detection/executor/cpp/streaming/BasicAmqMessageSender.h#L46)):*

I agree that MPFAMQMessage looks like a jumble or letters. It's one thing if two different classes use slightly different naming conventions, but it seems weird to use two different naming conventions in the same name.
</blockquote></details>

Is that something you're willing to do as part of this PR?

---

*[trunk/detection/executor/cpp/streaming/ExecutorErrors.h, line 28 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/401#-L2a8oSZan5P22vJa1zD:-L2g9V7sMU2jmCF0m9u8:b-s8bb24) ([raw file](https://github.com/openmpf/openmpf/blob/b0190b5aa91721935a8da2dbd1c0cf6021a44ba9/trunk/detection/executor/cpp/streaming/ExecutorErrors.h#L28)):*

I don't manually add those, CLion does it for me. `PROJECTNAME_NAMEOFFILE_H` is the default pattern. PROJECTNAME is taken from the CMake function: `project(MPF)`. All the CMake projects except for this one have dashes in their name, for example `project(openmpf-components)`. The dashes get converted to underscores so you will see include guards like, `OPENMPF_COMPONENTS_CAFFEDETECTION_H`.

I can change my CLion settings to use a different pattern when it creates new source files, but the real problem is that the pattern is used in just about every C++ header file. The include guards are really just noise, so I don't really think it is worth it to go back and change all those files.
</blockquote></details>

Ok. If that's what CLion is doing by default then let's just go with that.

---

*[trunk/detection/executor/cpp/streaming/main.cpp, line 125 at r2](https://reviewable.io:443/reviews/openmpf/openmpf/401#-L2bG62eHsWRXUYHG7Tk:-L2gL4rHb0AJq2bt-gSw:b-tah54x) ([raw file](https://github.com/openmpf/openmpf/blob/132c2e7d16a6274794def58f4d3c9cc06a7bf0aa/trunk/detection/executor/cpp/streaming/main.cpp#L125)):*

No, there is `break;` right after `read_failed = true;`.  If I were to check `read_failed` here, the rest of the body of the loop would execute.
</blockquote></details>

I didn't see the `break` before. It's fine.

---

*[trunk/detection/executor/cpp/streaming/QuitWatcher.cpp, line 63 at r2](https://reviewable.io:443/reviews/openmpf/openmpf/401#-L2bHn_OKWx5Of6ZWX_z:-L2gB2MBjxXI0CXuM4GE:b-cz65pw) ([raw file](https://github.com/openmpf/openmpf/blob/132c2e7d16a6274794def58f4d3c9cc06a7bf0aa/trunk/detection/executor/cpp/streaming/QuitWatcher.cpp#L63)):*

My intent is to also check for the other commands here. I would also add flags and methods to check them for the other commands. I now notice that the naming of the class and methods wouldn't make sense for those commands. I plan to use this same class, but I need to re-think the naming.
</blockquote></details>

StandardInWatcher? CommandLineWatcher?

---

*[trunk/detection/executor/cpp/streaming/StreamingExecutorLog4cxxConfig.xml, line 55 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/401#-L2awlgL0hav7v04TNRB:-L2gBviVEPz-qz-ixom9:bpeykrq) ([raw file](https://github.com/openmpf/openmpf/blob/b0190b5aa91721935a8da2dbd1c0cf6021a44ba9/trunk/detection/executor/cpp/streaming/StreamingExecutorLog4cxxConfig.xml#L55)):*

Yes, I don't think a hex representation of the thread id is useful.
</blockquote></details>

Mostly it's there for parity with the Java logs, which use a more meaningful string representation for the thread. I've found it useful on the Java side (mostly WFM) to better help understand the log timestamps. Knowing that instructions are executed in different threads helps.

Really, we  only need the `%d %p` part to get `<TIMESTAMP><LEVEL>`, which is required by `FilteredMpfLogTailerListener` to parse the log content correctly.

I'm fine with removing `[%t]`, but it should be removed from all the C++ logs, which should be a separate task. Do you think it makes sense to remove it from all the C++ logs?

---

*[trunk/detection/test-components/cpp/CplusplusHelloWorld/plugin-files/descriptor/descriptor.json, line 10 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/401#-L2av2X9vlhWVYggA3uZ:-L2gUWUqlJOZb64z0OZ4:b-jeslk) ([raw file](https://github.com/openmpf/openmpf/blob/b0190b5aa91721935a8da2dbd1c0cf6021a44ba9/trunk/detection/test-components/cpp/CplusplusHelloWorld/plugin-files/descriptor/descriptor.json#L10)):*

It's not required now that we don't set LD_LIBRARY_PATH.
</blockquote></details>

That's true.

---

*[trunk/node-manager/src/main/java/org/mitre/mpf/nms/BaseServiceLauncher.java, line 129 at r3](https://reviewable.io:443/reviews/openmpf/openmpf/401#-L2gUk2EpogNUk54uqhd:-L2gUk2EpogNUk54uqhe:b-oxgb06) ([raw file](https://github.com/openmpf/openmpf/blob/e274a2c93814b74bfa3679be1ed9f4dc07a0d86a/trunk/node-manager/src/main/java/org/mitre/mpf/nms/BaseServiceLauncher.java#L129)):*
> ```Java
>     	
>         this.mServiceDesc = desc;
>         this.mStrSub = new StrSubstitutor(new StrLookup<String>() {
> ```

Good find!

---

*[trunk/node-manager/src/main/java/org/mitre/mpf/nms/streaming/IniManager.java, line 119 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/401#-L2aie0FVCnlaLfAXgJ5:-L2gRUgs2DIblGahCeOp:b-1hjx54) ([raw file](https://github.com/openmpf/openmpf/blob/b0190b5aa91721935a8da2dbd1c0cf6021a44ba9/trunk/node-manager/src/main/java/org/mitre/mpf/nms/streaming/IniManager.java#L119)):*

I wasn't sure whether or not there would be other configuration options that needed to be changed. This makes sure that all the ini files use the same configuration.
I made the changes to the commented out  methods, but reviewable is hiding them (from me at least) because they are just whitespace changes.
</blockquote></details>

Looking closer, I see the changes now.

---

*[trunk/node-manager/src/main/java/org/mitre/mpf/nms/streaming/StreamingProcess.java, line 128 at r3](https://reviewable.io:443/reviews/openmpf/openmpf/401#-L2gRraIJEGVB0TZNNgX:-L2gRraJ_rAuV1KWBseT:bf7vwj3) ([raw file](https://github.com/openmpf/openmpf/blob/31325a7d0e3334ac0df22a64d8f34d5571055c87/trunk/node-manager/src/main/java/org/mitre/mpf/nms/streaming/StreamingProcess.java#L128)):*
> ```Java
> 			LOG.info("Process: {} exited with exit code {}", _executable, exitCode);
> 			if (exitCode == StreamingProcessExitReason.STREAM_STALLED.exitCode) {
> 				// Throw exception to prevent the restart process when process exits due to a stall.
> ```

Let's reword to: "prevent restarting when".

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/nodeManager/NodeManagerStatus.java, line 311 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/401#-L2aLoRfhwdefB_jLMqM:-L2gS_4MRSxfkNhKTLQY:b-859oyx) ([raw file](https://github.com/openmpf/openmpf/blob/b0190b5aa91721935a8da2dbd1c0cf6021a44ba9/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/nodeManager/NodeManagerStatus.java#L311)):*

The StreamingProcessExitReason enum has been changed to be more specific about the exit reason. There is no longer a StreamingProcessExitReason.ERROR. I can log a message with the exit reason.
</blockquote></details>

I understand now. It's probably best not to have one long list of error cases that all map to JobStatus.ERROR. Better to just have the default to make it easier to add more exit reason errors later on without needing to modify this code.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/401)*

brosenberg42:
Review status: 44 of 51 files reviewed at latest revision, 8 unresolved discussions.

---

*[trunk/detection/executor/cpp/streaming/BasicAmqMessageSender.cpp, line 97 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/401#-L2a3SJMAQcbXywivA8m:-L3-dIjz77-48BdvUvXq:b6stzcb) ([raw file](https://github.com/openmpf/openmpf/blob/b0190b5aa91721935a8da2dbd1c0cf6021a44ba9/trunk/detection/executor/cpp/streaming/BasicAmqMessageSender.cpp#L97)):*
> When an error occurs, should we also pass along any tracks that have been generated so far?

The streaming component executor now does this.

---

*[trunk/detection/executor/cpp/streaming/BasicAmqMessageSender.cpp, line 101 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/401#-L2a0wPozvMrO3bFhIdR:-L3-c0sYbq9KxrNbkYzD:b-8lulcr) ([raw file](https://github.com/openmpf/openmpf/blob/b0190b5aa91721935a8da2dbd1c0cf6021a44ba9/trunk/detection/executor/cpp/streaming/BasicAmqMessageSender.cpp#L101)):*

Can we also call this `SendSummaryReport` and use function overloading?
</blockquote></details>

This is no longer relevant. Now, there is only one SendSummaryReport method that uses a default argument.

---

*[trunk/detection/executor/cpp/streaming/BasicAmqMessageSender.h, line 45 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/401#-L2a8-y96LcSXaQHkfC4:-L3-cQOpVFMDbL28113b:bh5915j) ([raw file](https://github.com/openmpf/openmpf/blob/b0190b5aa91721935a8da2dbd1c0cf6021a44ba9/trunk/detection/executor/cpp/streaming/BasicAmqMessageSender.h#L45)):*

It seems like some of the converters in MPFAMQMessage.h can be used in the reduced scope implementation. Have you considered using them?
</blockquote></details>

It will require too many changes. I don't want to make changes to make it work with the "reduced scope" implementation only to change it again when we add support for multistage pipelines.  

---

*[trunk/detection/executor/cpp/streaming/BasicAmqMessageSender.h, line 46 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/401#-L2a60PvsDSHJS-EEQQI:-L3-ctUgfOP-_ga03Voz:bd21ch6) ([raw file](https://github.com/openmpf/openmpf/blob/b0190b5aa91721935a8da2dbd1c0cf6021a44ba9/trunk/detection/executor/cpp/streaming/BasicAmqMessageSender.h#L46)):*

Is that something you're willing to do as part of this PR?
</blockquote></details>

No, it will result in a cascade of changes that don't need to be in this pull request.

---

*[trunk/detection/executor/cpp/streaming/main.cpp, line 145 at r2](https://reviewable.io:443/reviews/openmpf/openmpf/401#-L2cosEIN4RN4adaaluh:-L3-d0LQTUZIqIRTk53p:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/132c2e7d16a6274794def58f4d3c9cc06a7bf0aa/trunk/detection/executor/cpp/streaming/main.cpp#L145)):*

A new activity alert should only be sent once per segment.
</blockquote></details>

Done.

---

*[trunk/detection/executor/cpp/streaming/StandardInWatcher.cpp, line 63 at r2](https://reviewable.io:443/reviews/openmpf/openmpf/401#-L2bHn_OKWx5Of6ZWX_z:-L3-d3MQnDWW2oJaUo9o:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/132c2e7d16a6274794def58f4d3c9cc06a7bf0aa/trunk/detection/executor/cpp/streaming/StandardInWatcher.cpp#L63)):*

StandardInWatcher? CommandLineWatcher?
</blockquote></details>

Done.

---

*[trunk/detection/executor/cpp/streaming/StreamingExecutorLog4cxxConfig.xml, line 55 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/401#-L2awlgL0hav7v04TNRB:-L2zCSHScqUZkdx9akVR:bf4smp0) ([raw file](https://github.com/openmpf/openmpf/blob/b0190b5aa91721935a8da2dbd1c0cf6021a44ba9/trunk/detection/executor/cpp/streaming/StreamingExecutorLog4cxxConfig.xml#L55)):*

Mostly it's there for parity with the Java logs, which use a more meaningful string representation for the thread. I've found it useful on the Java side (mostly WFM) to better help understand the log timestamps. Knowing that instructions are executed in different threads helps.

Really, we  only need the `%d %p` part to get `<TIMESTAMP><LEVEL>`, which is required by `FilteredMpfLogTailerListener` to parse the log content correctly.

I'm fine with removing `[%t]`, but it should be removed from all the C++ logs, which should be a separate task. Do you think it makes sense to remove it from all the C++ logs?
</blockquote></details>

I think it makes sense to remove it from all the C++ logs.

---

*[trunk/node-manager/src/main/java/org/mitre/mpf/nms/streaming/StreamingProcess.java, line 128 at r3](https://reviewable.io:443/reviews/openmpf/openmpf/401#-L2gRraIJEGVB0TZNNgX:-L3-byO-YGJ7JAwwzQ42:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/31325a7d0e3334ac0df22a64d8f34d5571055c87/trunk/node-manager/src/main/java/org/mitre/mpf/nms/streaming/StreamingProcess.java#L128)):*

Let's reword to: "prevent restarting when".
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/401)*

jrobble:
Reviewed 9 of 9 files at r4.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[trunk/detection/executor/cpp/streaming/BasicAmqMessageSender.h, line 46 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/401#-L2a60PvsDSHJS-EEQQI:-L3-iT240n3k62ah_p8W:bxluwkf) ([raw file](https://github.com/openmpf/openmpf/blob/b0190b5aa91721935a8da2dbd1c0cf6021a44ba9/trunk/detection/executor/cpp/streaming/BasicAmqMessageSender.h#L46)):*

No, it will result in a cascade of changes that don't need to be in this pull request.
</blockquote></details>

Ok. I created #407 for later.

---

*[trunk/detection/executor/cpp/streaming/StandardInWatcher.h, line 53 at r4](https://reviewable.io:443/reviews/openmpf/openmpf/401#-L3-m2hyKusP2r9eUKak:-L3-m2hyKusP2r9eUKal:bjbnmjn) ([raw file](https://github.com/openmpf/openmpf/blob/c06ba850dc9732d40d1bfe695186a0f3e36e222f/trunk/detection/executor/cpp/streaming/StandardInWatcher.h#L53)):*
> ```objc
>         static std::string error_message_;
> 
>         static void WatchForStandardIn();
> ```

Please rename this to `WatchStandardIn()`, or just `Watch()`.

---

*[trunk/detection/executor/cpp/streaming/StreamingExecutorLog4cxxConfig.xml, line 55 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/401#-L2awlgL0hav7v04TNRB:-L3-mv0iKJ3trugczYgJ:b-zaegjm) ([raw file](https://github.com/openmpf/openmpf/blob/b0190b5aa91721935a8da2dbd1c0cf6021a44ba9/trunk/detection/executor/cpp/streaming/StreamingExecutorLog4cxxConfig.xml#L55)):*

I think it makes sense to remove it from all the C++ logs.
</blockquote></details>

Ok. I created #408 for later.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/401)*

brosenberg42:
Review status: 48 of 51 files reviewed at latest revision, 1 unresolved discussion.

---

*[trunk/detection/executor/cpp/streaming/StandardInWatcher.h, line 53 at r4](https://reviewable.io:443/reviews/openmpf/openmpf/401#-L3-m2hyKusP2r9eUKak:-L3-qPt490RD9IBQsluh:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/c06ba850dc9732d40d1bfe695186a0f3e36e222f/trunk/detection/executor/cpp/streaming/StandardInWatcher.h#L53)):*

Please rename this to `WatchStandardIn()`, or just `Watch()`.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/401)*

jrobble:
Reviewed 2 of 9 files at r4, 3 of 3 files at r5.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/401)*

jrobble:
Reviewed 2 of 2 files at r6.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/401)*

jrobble:

---

*[trunk/detection/test-components/cpp/CplusplusHelloWorld/plugin-files/descriptor/descriptor.json, line 10 at r6](https://reviewable.io:443/reviews/openmpf/openmpf/401#-L33ytMKoXJWI5FKvTKp:-L33ytMLevokp_I2wqkD:bs1r95n) ([raw file](https://github.com/openmpf/openmpf/blob/2e4fbd16ed61f94138933e9545994cd48b3e5e36/trunk/detection/test-components/cpp/CplusplusHelloWorld/plugin-files/descriptor/descriptor.json#L10)):*
> ```JSON
>   "environmentVariables" : [{
>       "name":"LD_LIBRARY_PATH",
>       "value":"${MPF_HOME}/plugins/CplusplusHelloWorld/lib"
> ```

I was able to successfully register this component, but there was a problem when starting it on the Nodes page. Here's the error:

```
[mpf@localhost CplusplusHelloWorld]$ LD_LIBRARY_PATH=${MPF_HOME}/plugins/CplusplusHelloWorld/lib ${MPF_HOME}/bin/amq_detection_component ${MPF_HOME}/plugins/CplusplusHelloWorld/lib/libmpfCplusplusHelloWorld.so MPF.DETECTION_CPLUSPLUSHELLOWORLD_REQUEST
/home/mpf/git/openmpf-projects/openmpf/trunk/install/bin/amq_detection_component: error while loading shared libraries: libactivemq-cpp.so.19: cannot open shared object file: No such file or directory

[mpf@localhost CplusplusHelloWorld]$ echo $?
127
```

To resolve this, I had to add `:${LD_LIBRARY_PATH}` to the above line as follows:

```
  "environmentVariables" : [{
      "name":"LD_LIBRARY_PATH",
      "value":"${MPF_HOME}/plugins/CplusplusHelloWorld/lib:${LD_LIBRARY_PATH}"
  }],
```

Note that the sample hello world component in the C++ SDK repo does the same.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/401)*

jrobble:
Reviewed 4 of 4 files at r7.

---

*[trunk/detection/test-components/cpp/CplusplusHelloWorld/HelloWorld.cpp, line 50 at r7](https://reviewable.io:443/reviews/openmpf/openmpf/401#-L34kgeQyvr-qS4Fdyhg:-L34kgeRqum6QufdxUUy:b-dnu1g0) ([raw file](https://github.com/openmpf/openmpf/blob/e274a2c93814b74bfa3679be1ed9f4dc07a0d86a/trunk/detection/test-components/cpp/CplusplusHelloWorld/HelloWorld.cpp#L50)):*
> ```cpp
> 
>     // Configure logger
>     log4cxx::xml::DOMConfigurator::configure(run_dir + "/CplusplusHelloWorld/config/Log4cxxConfig.xml");
> ```

I would prefer that the batch version of both the test and sample hello world components use a log file to match the streaming version.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/401)*

brosenberg42:
Review status: 52 of 57 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[trunk/detection/test-components/cpp/CplusplusHelloWorld/HelloWorld.cpp, line 50 at r7](https://reviewable.io:443/reviews/openmpf/openmpf/401#-L34kgeQyvr-qS4Fdyhg:-L356vVcOHFK0DxDG0dq:b-xqoljk) ([raw file](https://github.com/openmpf/openmpf/blob/e274a2c93814b74bfa3679be1ed9f4dc07a0d86a/trunk/detection/test-components/cpp/CplusplusHelloWorld/HelloWorld.cpp#L50)):*

I would prefer that the batch version of both the test and sample hello world components use a log file to match the streaming version.
</blockquote></details>

Yea, I agree. I just blindly copied and pasted.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/401)*

jrobble:
Reviewed 5 of 5 files at r8.

---

*[trunk/detection/test-components/cpp/CplusplusHelloWorld/plugin-files/descriptor/descriptor.json, line 24 at r8](https://reviewable.io:443/reviews/openmpf/openmpf/401#-L35ooTIOu4MOWU_Otn_:-L35ooTIOu4MOWU_Otna:b-hoalz6) ([raw file](https://github.com/openmpf/openmpf/blob/c371bd824e964a78cd4ef11a7e7e77848f0fc40e/trunk/detection/test-components/cpp/CplusplusHelloWorld/plugin-files/descriptor/descriptor.json#L24)):*
> ```JSON
>         "state2"
>       ],
>       "properties" : [
> ```

This is missing:

```
      "properties" : [
        {
          "name" : "FRAME_INTERVAL",
          "description" : "This is a test property with a default value of 1.",
          "type" : "INT",
          "defaultValue" : "1"
        }
      ]
```

Right now I get the following exception:

```
2018-01-17 20:06:21,512 ERROR [0x7feb59eb97c0] org.mitre.mpf.detection:471 - Standard Exception caught in main.cpp:  map::at
```

I believe it's because of this line:

```
    LOG4CXX_INFO(hw_logger_, "[" << job.job_name << "] Job properties contains FRAME_INTERVAL with a value of "
                                 << job.job_properties.at("FRAME_INTERVAL") << ".");
```

Either add the FRAME_INTERVAL here, or update that log line to print out one or more of the properties below. I'm leaning towards the latter since we already have tests that use the 3 properties below.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/401)*

brosenberg42:
Review status: 56 of 57 files reviewed at latest revision, 1 unresolved discussion.

---

*[trunk/detection/test-components/cpp/CplusplusHelloWorld/plugin-files/descriptor/descriptor.json, line 24 at r8](https://reviewable.io:443/reviews/openmpf/openmpf/401#-L35ooTIOu4MOWU_Otn_:-L39-pycZyQgQ5r7Ac87:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/c371bd824e964a78cd4ef11a7e7e77848f0fc40e/trunk/detection/test-components/cpp/CplusplusHelloWorld/plugin-files/descriptor/descriptor.json#L24)):*

This is missing:

```
      "properties" : [
        {
          "name" : "FRAME_INTERVAL",
          "description" : "This is a test property with a default value of 1.",
          "type" : "INT",
          "defaultValue" : "1"
        }
      ]
```

Right now I get the following exception:

```
2018-01-17 20:06:21,512 ERROR [0x7feb59eb97c0] org.mitre.mpf.detection:471 - Standard Exception caught in main.cpp:  map::at
```

I believe it's because of this line:

```
    LOG4CXX_INFO(hw_logger_, "[" << job.job_name << "] Job properties contains FRAME_INTERVAL with a value of "
                                 << job.job_properties.at("FRAME_INTERVAL") << ".");
```

Either add the FRAME_INTERVAL here, or update that log line to print out one or more of the properties below. I'm leaning towards the latter since we already have tests that use the 3 properties below.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/401)*

jrobble:
Reviewed 1 of 2 files at r9, 1 of 1 files at r10.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/401)*

jrobble:
Reviewed 2 of 2 files at r11.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/401)*

jrobble:
Reviewed 1 of 1 files at r12.

---

*[trunk/workflow-manager/pom.xml, line 453 at r12](https://reviewable.io:443/reviews/openmpf/openmpf/401#-L3Z2vAxoYVgyZ1FSb6j:-L3Z2vAxoYVgyZ1FSb6k:b-v60fmo) ([raw file](https://github.com/openmpf/openmpf/blob/76888600b7cc946228440aba2162903352c71d50/trunk/workflow-manager/pom.xml#L453)):*
> ```XML
>             <artifactId>hibernate-infinispan</artifactId>
>             <version>${hibernate.framework.version}</version>
>             <exclusions>
> ```

Did this make any difference? In either case, it makes sense if we're including our own version.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/401)*

brosenberg42:
*[trunk/workflow-manager/pom.xml, line 453 at r12](https://reviewable.io:443/reviews/openmpf/openmpf/401#-L3Z2vAxoYVgyZ1FSb6j:-L3ZGY-nH4tEnqmCRhyT:b-sug82z) ([raw file](https://github.com/openmpf/openmpf/blob/76888600b7cc946228440aba2162903352c71d50/trunk/workflow-manager/pom.xml#L453)):*

Did this make any difference? In either case, it makes sense if we're including our own version.
</blockquote></details>

If you run `mvn dependency:tree -Dverbose -Dincludes=org.jgroups` in the openmpf directory you will see:
```
[INFO] ------------------------------------------------------------------------
[INFO] Building MPF Workflow Manager 1.0.0
[INFO] ------------------------------------------------------------------------
[INFO] 
[INFO] --- maven-dependency-plugin:2.8:tree (default-cli) @ mpf-workflowManager ---
[INFO] org.mitre.mpf:mpf-workflowManager:war:1.0.0
[INFO] +- org.mitre.mpf:mpf-nodemanager:jar:1.0.0:compile
[INFO] |  \- org.jgroups:jgroups:jar:3.6.4.Final:compile
[INFO] \- org.hibernate:hibernate-infinispan:jar:5.0.8.Final:compile
[INFO]    \- org.infinispan:infinispan-core:jar:7.2.1.Final:compile
[INFO]       \- (org.jgroups:jgroups:jar:3.6.2.Final:compile - omitted for conflict with 3.6.4.Final)
[INFO]                                                                         
```
After adding the exclusion there is no longer a conflict. It really shouldn't be doing anything different since it says omitted.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/401)*

jrobble:
Reviewed 7 of 7 files at r13.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/401)*

PR #406, opened 2018-01-12
added isActive parameter to [GET] /rest/streaming/jobs

dpierce17:
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/406)

### DIFF
--- a/trunk/workflow-manager/src/main/java/org/mitre/mpf/mvc/controller/StreamingJobController.java
+++ b/trunk/workflow-manager/src/main/java/org/mitre/mpf/mvc/controller/StreamingJobController.java
@@ -148,15 +148,16 @@ public class StreamingJobController {
      */
     //EXTERNAL
     @RequestMapping(value = "/rest/streaming/jobs", method = RequestMethod.GET)
-    @ApiOperation(value = "Gets a list of job ids for all streaming jobs.",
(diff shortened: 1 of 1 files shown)


jrobble:
Reviewed 1 of 1 files at r1.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/mvc/controller/StreamingJobController.java, line 151 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/406#-L2gblUs1XfaBmQvlDbu:-L2gblUs1XfaBmQvlDbv:bokzxcf) ([raw file](https://github.com/openmpf/openmpf/blob/ee687066b9e484ad5f375f63793ef8e7a8ef0b2e/trunk/workflow-manager/src/main/java/org/mitre/mpf/mvc/controller/StreamingJobController.java#L151)):*
Works as expected.

When setting "isActive" to false, it occurred to me that users might think that should return only streaming jobs that are cancelled / terminated (those that aren't active).

Update the comment above to say:

> "Gets a list of job ids for streaming jobs. If isActive is true, don't include streaming jobs that are terminated or cancelled. If false, return all streaming jobs."

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/406)*

jrobble:
Reviewed 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/406)*

PR #417, opened 2018-01-23
Extra nm logging

brosenberg42:
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/417)


jrobble:
Would any of this logging be useful to include long-term?

PR #422, opened 2018-01-25
Fixed frame number checks in C++ streaming component executor.

jrobble:
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/422)

### DIFF
--- a/trunk/detection/executor/cpp/streaming/main.cpp
+++ b/trunk/detection/executor/cpp/streaming/main.cpp
@@ -172,6 +172,7 @@ private:
             StandardInWatcher *std_in_watcher = StandardInWatcher::GetInstance();
 
             bool segment_activity_alert_sent = false;
+            int segment_end = -1;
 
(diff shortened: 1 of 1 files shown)


jrobble:
The current executor code has two issues:

1. If the segment size is 100, it checks to see if the first segment ends at frame 100 when it should be checking that the segment ends at frame 99. This is because the segment range is [0, 99], not [0, 100].

2. After the first segment, the code will run `BeginSegment()`, then `ProcessFrame()` once, then `EndSegment()`, regardless of the segment size. This is because `frame_ number` is incremented before `BeginSegment()`, which has the following check:

```
if (frame_number % settings_.segment_size == 0) {
```

and the `frame_number` does not change when this check is executed before `EndSegment()`:

```
if (frame_number != 0 && (frame_number % settings_.segment_size == 0)) {
```

The modulo part of both checks essentially check for the same thing because `frame_ number` doesn't change.

Issue #400, opened 2018-01-10
Create StreamingJobStatus class


jrobble:
Create a StreamingJobStatus class that's similar to the existing JobStatus class, but is only for stream processing. The enum entries in the class should be limited to those that are only useful for streaming. The TERMINATED, PAUSED, and STALLED entries can be removed from the JobStatus class since they are not relevant to batch processing.

Add a detail (String) field to StreamingJobStatus so that the ERROR status can have additional information. When health reports are generated, they should get the status value from this field if it's populated. Part of this task is to ensure the StreamingJobStatus information can be stored in mySQL and Redis. This is slightly different from storing the JobStatus information because of this additional field.

Also, ensure that the proper StreamingJobStatus is created in `NodeManagerStatus.streamingJobExited()`.

When the WFM receives the segment summary report, if the error string is populated, set the StreamingJobStatus to ERROR, set the detail to that string’s value, and update mySQL and Redis accordingly.

dpierce17:
Where I find deeper links into the legacy code (i.e. AtmosphereController.broadcast) I plan to just send the string equivalent of the streaming job status without the detail field.

dpierce17:
if the stream status detail string is to be added to the persistent storage (mysql), need to alter the mpf streaming_job_request table (for existing jobs) by doing the following:

 mysql --user=root -p -e "alter table mpf.streaming_job_request add column status_detail varchar(255) after status;"

Some variant of this statement will have to be added to the deployment scripts (TODO)

dpierce17:
we should add the mysql alter table statement to mysql.yml. I'll create an issue for this once I get closer to landing this issue

dpierce17:
FYI: I've run into a problem adapting the WFM Hibernate*Dao software (supports persistent storage to mysql database) for StreamingJobStatus (i.e. now a property constructed from two columns of a mysql table). I changed StreamingJobRequest to construct StreamingJobStatus from two columns using Embedded types, but I'm getting an java.lang.IllegalArgumentException similar to as described in this Hibernate bug report: https://hibernate.atlassian.net/browse/HHH-11537. I'll be working with this a bit longer, but wanted to report this possible issue here.

jrobble:
> FYI: I've run into a problem

If you'd like, I can help you debug if you push your code to a branch I can access.

dpierce17:
Landed with PR#433

Issue #397, opened 2018-01-08
Check validity of batch job


jrobble:
Update `JobController.createJobInternal()` to use `pipelineService.pipelineSupportsBatch()`. This is similar to how `StreamingJobController.createStreamingJobInternal()` does `!pipelineService.pipelineSupportsStreaming(streamingJobCreationRequest.getPipelineName()`. 

It's possible that someone may develop a component that only supports streaming. Using it to create a batch job will cause problems.

dpierce17:
Landed with PR#433

Issue #411, opened 2018-01-19
Add last activity frame id and activity timestamp to StreamingJobInfo


dpierce17:
StreamingJobInfo should also contain the last activity frame id and activity timestamp. This is essentially like getting the last health report on demand. Users may want to do this if they choose not to register for health report callbacks.

dpierce17:
Since we have added a status detail parameter to streaming jobs, we should include the status detail in StreamingJobInfo, along with last activity frame id and activity timestamp. Adding status detail to this task.

Issue #348, opened 2017-11-27
Add handling in WFM for messages received in JobStatus, NewActivity, SummaryReport queues


dpierce17:
The streaming component executor will send messages to WFM via queues:
•	Job status queue : for sending STALLED and IN_PROGRESS messages
•	New activity queue : for sending frame id and frame timestamp (system time)
•	Summary report queue (protobuf of video track collection)

dpierce17:
Will all timestamps be in UTC standard time?

jrobble:
Currently OpenMPF uses system time, which is not necessarily UTC.

jrobble:
Note that we’ve moved away from using a DetectionReponse for the segment summary report and are now instead using a [StreamingDetectionReponse](https://github.com/openmpf/openmpf/blob/develop/trunk/protobuf/src/main/proto/detection.proto#L241). @brosenberg42 created a [POJO representation](https://reviewable.io/reviews/openmpf/openmpf/401/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/transients/SegmentSummaryReport.java) for it in his current PR. In the past we planned on the segment summary report taking the same form as the JSON output object. Now it’s slightly different. To account for this, we need to create a JSON representation for the segment summary report in the interop package so that we can return it in the callback.

jrobble:
Note that the POJO representation does not include timestamps for tracks nor timestamps for the individual detections that make up a track. Those need to be included in the summary report JSON that's returned in the callback. 

Consider reusing `JsonDetectionOutputObject`, which has `offsetTime, and `JsonTrackOuputObject`, which has `startOffsetTime` and `stopOffsetTime`.

PR #433, opened 2018-01-30
Streaming job status

dpierce17:
WFM support issues #397 and #400

---
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/433)


jrobble:
Please do a search and replace on "Copyright 2017". Update that to "Copyright 2018" in our comment header blocks. Some of the files in this PR still have 2017.

Also, please do a search and replace on "last activity" and "lastActivity" (caps insensitive). Please remove "last" if appropriate.

---

Reviewed 3 of 38 files at r1, 43 of 46 files at r3, 8 of 9 files at r4, 3 of 3 files at r5.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/mvc/util/ModelUtils.java, line 145 at r3](https://beta.reviewable.io/reviews/openmpf/openmpf/433#-L46w1yGRXoDIOzrYnzm:-L46vx6Ni4JhYnmkNvkP:b-8fr56v) ([raw file](https://github.com/openmpf/openmpf/blob/c84d4516a287972b5d8e67e01a2fc02187fffcf7/trunk/workflow-manager/src/main/java/org/mitre/mpf/mvc/util/ModelUtils.java#L145)):*
> ```Java
> 		// some job status' may be terminal
> 		boolean isTerminal = (streamingJobStatus != null && streamingJobStatus.isTerminal());
>         // TODO add status detail to StreamingJobInfo (issue #411) using streamingJobRequest.getStatusDetail()
> ```

Putting the issue # here is fine because this should be resolved in the short term. For long-term TODOs, we should avoid putting in the issue number in case we ever change issue trackers due to Apache incubation.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/businessrules/StreamingJobRequestBo.java, line 98 at r3](https://beta.reviewable.io/reviews/openmpf/openmpf/433#-L474kds4Dls9AwtVh30:-L474kds4Dls9AwtVh31:bbeb5bb) ([raw file](https://github.com/openmpf/openmpf/blob/c84d4516a287972b5d8e67e01a2fc02187fffcf7/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/businessrules/StreamingJobRequestBo.java#L98)):*
> ```Java
> 	void jobCompleted(long jobId, StreamingJobStatus jobStatus);
> 
> 	void handleJobStatusChange(long jobId, StreamingJobStatusType status, long timestamp);
> ```

It seems like this should also take `StreamingJobStatus` (instead of `StreamingJobStatusType`). I think that we should commit to using `StreamingJobStatus` everywhere over `StreamingJobStatusType`, even if the inner detail string is empty or null.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/businessrules/impl/StreamingJobRequestBoImpl.java, line 490 at r3](https://beta.reviewable.io/reviews/openmpf/openmpf/433#-L46xP0YzrVZVKQ4k2uY:-L46xP0ZTSkGzluRi2az:b-ik448a) ([raw file](https://github.com/openmpf/openmpf/blob/c84d4516a287972b5d8e67e01a2fc02187fffcf7/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/businessrules/impl/StreamingJobRequestBoImpl.java#L490)):*
> ```Java
>             try {
>                 // make an effort to mark the streaming job as failed
>                 String warningMessage;
> ```

Let's change this to `errorMessage` and log it as an error since it results in a `JOB_CREATION_ERROR` status.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/businessrules/impl/StreamingJobRequestBoImpl.java, line 86 at r5](https://beta.reviewable.io/reviews/openmpf/openmpf/433#-L47IvVoAUy7IGIxJpuL:-L47IvVpBb0MzDm4Rixk:b-zyk1n) ([raw file](https://github.com/openmpf/openmpf/blob/8a8bf9034421370a23b60a4c292c041ed19f844b/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/businessrules/impl/StreamingJobRequestBoImpl.java#L86)):*
> ```Java
> 
>     private static final String INVALID_PIPELINE_MESSAGE = "INVALID_PIPELINE_MESSAGE";
>     private static final String CREATE_TRANSIENT_JOB_FAILED_MESSAGE = "CREATE_TRANSIENT_JOB_FAILED_MESSAGE";
> ```

I would prefer that we do away with these static strings and use more user-friendly messages. 

`CREATE_TRANSIENT_JOB_FAILED_MESSAGE` is only used in one place, so that shouldn't be a problem.

`INVALID_PIPELINE_MESSAGE` is used to create an exception in one place, and used to check for that kind of exception in another. Instead we should be throwing a specific kind of exception and catching it. See comments below.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/businessrules/impl/StreamingJobRequestBoImpl.java, line 492 at r5](https://beta.reviewable.io/reviews/openmpf/openmpf/433#-L47KScq0xwIkMl4pqxf:-L47KScr4-xyCStuFWBl:b-qvx6fy) ([raw file](https://github.com/openmpf/openmpf/blob/8a8bf9034421370a23b60a4c292c041ed19f844b/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/businessrules/impl/StreamingJobRequestBoImpl.java#L492)):*
> ```Java
>                 String warningMessage;
>                 if (INVALID_PIPELINE_MESSAGE.equals(e.getMessage())) {
>                     warningMessage = "Streaming Job #" + jobId + " did not specify a valid pipeline.";
> ```

You should instead catch `InvalidPipelineObjectWfmProcessingException` here and handle that differently.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/businessrules/impl/StreamingJobRequestBoImpl.java, line 553 at r5](https://beta.reviewable.io/reviews/openmpf/openmpf/433#-L47KN4V7xzuY04I7bl5:-L47KN4V7xzuY04I7bl6:b-xkexp0) ([raw file](https://github.com/openmpf/openmpf/blob/8a8bf9034421370a23b60a4c292c041ed19f844b/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/businessrules/impl/StreamingJobRequestBoImpl.java#L553)):*
> ```Java
>         if (transientPipeline == null) {
>             redis.setJobStatus(jobId, StreamingJobStatusType.IN_PROGRESS_ERRORS, INVALID_PIPELINE_MESSAGE);
>             throw new WfmProcessingException(INVALID_PIPELINE_MESSAGE);
> ```

This should instead throw `InvalidPipelineObjectWfmProcessingException`.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/Redis.java, line 100 at r3](https://beta.reviewable.io/reviews/openmpf/openmpf/433#-L47AWcgTh3iAhGD3_WP:-L47AWcgTh3iAhGD3_WQ:b-71onxw) ([raw file](https://github.com/openmpf/openmpf/blob/c84d4516a287972b5d8e67e01a2fc02187fffcf7/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/Redis.java#L100)):*
> ```Java
>      */
>     JobStatus getJobStatus(long jobId);
>     List<JobStatus> getJobStatuses(List<Long> jobIds);
> ```

Using the `JobStatus` superclass allows you to return a collection of both batch and streaming job ids here; however, this method is only ever used by `getJobStatusesAsString`, which is only used with streaming job ids for your health reporting behavior.

I would prefer to remove the `JobStatus` class entirely. That will also help avoid casting errors.

For now, you can rename those methods to `getStreamingJobStatuses` and `getStreamingJobStatusesAsStrings` (note the "s" on "Strings" there).

You can also split `getJobStatus` into `getBatchJobStatus` and `getStreamingJobStatus`. Our plan is to eventually split these Redis classes into a batch and streaming version in the long run. This is a step in that direction.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/RedisImpl.java, line 430 at r3](https://beta.reviewable.io/reviews/openmpf/openmpf/433#-L477SHxG_tPteAhNfTQ:-L477SHxG_tPteAhNfTR:b-nq8jgg) ([raw file](https://github.com/openmpf/openmpf/blob/c84d4516a287972b5d8e67e01a2fc02187fffcf7/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/RedisImpl.java#L430)):*
> ```Java
>             throw new WfmProcessingException("Error: this method shouldn't be called for batch jobs. Rejected this call for job " + jobId);
>         } else if ( isJobTypeStreaming(jobId) ) {
>             redisTemplate.boundHashOps(key(STREAMING_JOB, jobId)).put(STREAMING_JOB_STATUS, streamingJobStatus.name());
> ```

In the old `setJobStatus` code, the actual enum was put in Redis. Now you're putting in the String, and then converting back to an enum when you read it out. Wouldn't it be easier to store the enum?

Same goes for the batch version of this method.

If you do that, I think you may be able to get rid of the batch and streaming job status constructors that take a string and convert it into an enum. That would simplify things and is my preference.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/RedisImpl.java, line 928 at r3](https://beta.reviewable.io/reviews/openmpf/openmpf/433#-L478Ey9nuek3jdu_Eln:-L478EyAiNpYh_s4ghAx:bi5c8v4) ([raw file](https://github.com/openmpf/openmpf/blob/c84d4516a287972b5d8e67e01a2fc02187fffcf7/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/RedisImpl.java#L928)):*
> ```Java
>                 // Throw an exception if any job that may be in the long term database, but isn't in REDIS (i.e. it is not an active job), is passed to this method.
>                 throw new WfmProcessingException("Error: jobId " + " is not the id of an active Streaming job");
>             } else {
> ```

You don't need an `else` here because the method will end at the `throw` above.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/RedisImpl.java, line 1309 at r3](https://beta.reviewable.io/reviews/openmpf/openmpf/433#-L47C-x76bvCohYo2u4V:-L47C-x76bvCohYo2u4W:b-fkm6rv) ([raw file](https://github.com/openmpf/openmpf/blob/c84d4516a287972b5d8e67e01a2fc02187fffcf7/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/RedisImpl.java#L1309)):*
> ```Java
>                 if ( isActive ) {
>                     // include only those streaming jobs that do not have terminal jobStatus.
>                     return !((StreamingJobStatus)getJobStatus(jobId)).getJobStatus().isTerminal();
> ```

Once you create the`getStreamingJobStatus` method (see other comment), it would be best if this read as:

```
return !getStreamingJobStatus(jobId).isTerminal();
```

Note how I removed the second `getJobStatus` call. The `StreamingJobStatus` class should implement `isTerminal`, which is just a call-through to the underlying `StreamingJobStatusType`.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/access/hibernate/HibernateStreamingJobRequestDaoImpl.java, line 54 at r3](https://beta.reviewable.io/reviews/openmpf/openmpf/433#-L476-WFsQfIWqnEWwov:-L476-WGhEm_z05VOYlb:bi1bzl0) ([raw file](https://github.com/openmpf/openmpf/blob/c84d4516a287972b5d8e67e01a2fc02187fffcf7/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/access/hibernate/HibernateStreamingJobRequestDaoImpl.java#L54)):*
> ```Java
> 		query.setParameter("statusDetail", "shutdown: cancelling jobs in non-terminal state");
> 		query.setParameterList("nonTerminalStatuses", StreamingJobStatus.getNonTerminalStatuses());
> 		// TODO this isn't working when using StreamingJobRequests that contain embedded StreamingJobStatus
> ```

Can you please elaborate?

The streaming job status fields are parsed out into the columns in the streaming_job_request. I'm not sure what you mean by "embedded".

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/persistent/BatchJobStatus.java, line 38 at r3](https://beta.reviewable.io/reviews/openmpf/openmpf/433#-L46iFTOnpl5_TW4HHzw:-L46iFTOnpl5_TW4HHzx:b-s0tbv4) ([raw file](https://github.com/openmpf/openmpf/blob/c84d4516a287972b5d8e67e01a2fc02187fffcf7/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/persistent/BatchJobStatus.java#L38)):*
> ```Java
>  * This class includes the essential information which describes batch job status.
>  */
> public class BatchJobStatus extends JobStatus {
> ```

Since `JobStatus` is currently an empty class, it can be removed. Then this `BatchJobStatus` class won't extend it. Then I think you might as well combine this class with`BatchJobStatusType`. The end result would be one file (BatchJobStatusType.java), which would be declared as `public enum BatchJobStatusType`, similar to how the old `JobStatus` class was declared as an enum and had methods.

Actually, I think you can get the same result by basically taking the old `JobStatus` class and just renaming it to `BatchJobStatusType`.

You can't combine `StreamingJobStatusType` into `StreamingJobStatus` though. Because `StreamingJobStatus` has `statusDetail` it needs to be a class (although if `statusDetail` was a hardcoded string that would be a different story - but we shouldn't do that).

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/persistent/JobStatus.java, line 30 at r3](https://beta.reviewable.io/reviews/openmpf/openmpf/433#-L46hHhIk3iAflxoZkPn:-L46hHhIk3iAflxoZkPo:b-jhpzkz) ([raw file](https://github.com/openmpf/openmpf/blob/c84d4516a287972b5d8e67e01a2fc02187fffcf7/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/persistent/JobStatus.java#L30)):*
> ```Java
> 
> // JobStatus base class
> public class JobStatus {
> ```

If no common behaviors or data members can be put in this class, it should be removed. It currently doesn't do anything.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/persistent/StreamingJobStatus.java, line 67 at r5](https://beta.reviewable.io/reviews/openmpf/openmpf/433#-L47O4017akYL9C6QAgJ:-L47O4017akYL9C6QAgK:b-k2qsf9) ([raw file](https://github.com/openmpf/openmpf/blob/8a8bf9034421370a23b60a4c292c041ed19f844b/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/persistent/StreamingJobStatus.java#L67)):*
> ```Java
>     }
> 
>     private String status = null;
> ```

I find it a little confusing that we're storing the status as a string and an enum. We should only store it as a `StreamingJobStatusType` and return a string when necessary.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/persistent/StreamingJobStatus.java, line 72 at r5](https://beta.reviewable.io/reviews/openmpf/openmpf/433#-L47LYO11k-jig069uNn:-L47LYO11k-jig069uNo:bdyh2w7) ([raw file](https://github.com/openmpf/openmpf/blob/8a8bf9034421370a23b60a4c292c041ed19f844b/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/persistent/StreamingJobStatus.java#L72)):*
> ```Java
>         this.jobStatus = parse(status);
>     }
>     public void setStatus(StreamingJobStatusType jobStatus) {
> ```

I would prefer that we call this method `setType`.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/persistent/StreamingJobStatus.java, line 79 at r5](https://beta.reviewable.io/reviews/openmpf/openmpf/433#-L47OQII7p57mZ7hycX1:-L47OQII7p57mZ7hycX2:b532q2v) ([raw file](https://github.com/openmpf/openmpf/blob/8a8bf9034421370a23b60a4c292c041ed19f844b/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/persistent/StreamingJobStatus.java#L79)):*
> ```Java
> 
>     private String statusDetail = null;
>     public void setStatusDetail(String statusDetail) { this.statusDetail = statusDetail; }
> ```

Let's call this `setDetail` and the one below `getDetail`.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/persistent/StreamingJobStatus.java, line 83 at r5](https://beta.reviewable.io/reviews/openmpf/openmpf/433#-L47LPft8H4WuRphDyTa:-L47LPft8H4WuRphDyTb:bhi92hn) ([raw file](https://github.com/openmpf/openmpf/blob/8a8bf9034421370a23b60a4c292c041ed19f844b/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/persistent/StreamingJobStatus.java#L83)):*
> ```Java
> 
>     private StreamingJobStatusType jobStatus = DEFAULT;
>     public StreamingJobStatusType getJobStatus() { return this.jobStatus; }
> ```

I would prefer that we call this method `getType`.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/enums/StreamingJobStatusType.java, line 30 at r3](https://beta.reviewable.io/reviews/openmpf/openmpf/433#-L46jHROmB6LhSiDQHns:-L46jHROmB6LhSiDQHnt:b2hqiw6) ([raw file](https://github.com/openmpf/openmpf/blob/c84d4516a287972b5d8e67e01a2fc02187fffcf7/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/enums/StreamingJobStatusType.java#L30)):*
> ```Java
> 
> // StreamingJobStatusType enumeration describes all possible job status conditions applicable to a streaming job.
> public enum StreamingJobStatusType {
> ```

Please reduce this list to only the enums that are actually used for streaming. For example, I don't think these apply:

```
    IN_PROGRESS_ERRORS(false),
    IN_PROGRESS_WARNINGS(false),
    BUILDING_OUTPUT_OBJECT(false),
    COMPLETE(true),
    COMPLETE_WITH_ERRORS(true),
    COMPLETE_WITH_WARNINGS(true),
```

Please check the other ones too.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/openmpf/openmpf/433#-:-L46hZ8-YOQfU9FkAOLC:b-965c53)*

dpierce17:
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/RedisImpl.java, line 928 at r3](https://reviewable.io:443/reviews/openmpf/openmpf/433#-L478Ey9nuek3jdu_Eln:-L4CPxtcMyGKS7jEFeDV:b56krsc) ([raw file](https://github.com/openmpf/openmpf/blob/c84d4516a287972b5d8e67e01a2fc02187fffcf7/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/RedisImpl.java#L928)):*

You don't need an `else` here because the method will end at the `throw` above.
</blockquote></details>

I realize I don't need the else, but I like it for readability.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/433)*

dpierce17:
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/access/hibernate/HibernateStreamingJobRequestDaoImpl.java, line 54 at r3](https://reviewable.io:443/reviews/openmpf/openmpf/433#-L476-WFsQfIWqnEWwov:-L4CZynL0lNKzjOsORMN:bp26ueh) ([raw file](https://github.com/openmpf/openmpf/blob/c84d4516a287972b5d8e67e01a2fc02187fffcf7/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/access/hibernate/HibernateStreamingJobRequestDaoImpl.java#L54)):*

Can you please elaborate?

The streaming job status fields are parsed out into the columns in the streaming_job_request. I'm not sure what you mean by "embedded".
</blockquote></details>

My original implementation tried to use Embedded annotations to store job status and job detail from the database as a StreamingJobDetail type object constructed from the two columns. We're not currently doing this, so this TODO comment was obsolete. I removed it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/433)*

dpierce17:
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/enums/StreamingJobStatusType.java, line 30 at r3](https://reviewable.io:443/reviews/openmpf/openmpf/433#-L46jHROmB6LhSiDQHns:-L4CcKfAE1izNfO5TZhs:b-ndzmws) ([raw file](https://github.com/openmpf/openmpf/blob/c84d4516a287972b5d8e67e01a2fc02187fffcf7/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/enums/StreamingJobStatusType.java#L30)):*

Please reduce this list to only the enums that are actually used for streaming. For example, I don't think these apply:

```
    IN_PROGRESS_ERRORS(false),
    IN_PROGRESS_WARNINGS(false),
    BUILDING_OUTPUT_OBJECT(false),
    COMPLETE(true),
    COMPLETE_WITH_ERRORS(true),
    COMPLETE_WITH_WARNINGS(true),
```

Please check the other ones too.
</blockquote></details>

I removed these values.  I also removed UNKNOWN, because I don't think we will be using it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/433)*

dpierce17:
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/enums/StreamingJobStatusType.java, line 30 at r3](https://reviewable.io:443/reviews/openmpf/openmpf/433#-L46jHROmB6LhSiDQHns:-L4CcirV9jNrxWvdvnII:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/c84d4516a287972b5d8e67e01a2fc02187fffcf7/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/enums/StreamingJobStatusType.java#L30)):*

I removed these values.  I also removed UNKNOWN, because I don't think we will be using it.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/433)*

dpierce17:
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/enums/StreamingJobStatusType.java, line 30 at r3](https://reviewable.io:443/reviews/openmpf/openmpf/433#-L46jHROmB6LhSiDQHns:-L4CczmIjoNPwrJ860_x:b7scnha) ([raw file](https://github.com/openmpf/openmpf/blob/c84d4516a287972b5d8e67e01a2fc02187fffcf7/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/enums/StreamingJobStatusType.java#L30)):*

Done.
</blockquote></details>

Revising this, we'll be keeping UNKNOWN - I'm changing software to use UNKNOWN as the initial job status for streaming jobs.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/433)*

dpierce17:
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/persistent/BatchJobStatus.java, line 38 at r3](https://reviewable.io:443/reviews/openmpf/openmpf/433#-L46iFTOnpl5_TW4HHzw:-L4CmNwma9xEIgwCYqZE:bdfib0f) ([raw file](https://github.com/openmpf/openmpf/blob/c84d4516a287972b5d8e67e01a2fc02187fffcf7/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/persistent/BatchJobStatus.java#L38)):*

Since `JobStatus` is currently an empty class, it can be removed. Then this `BatchJobStatus` class won't extend it. Then I think you might as well combine this class with`BatchJobStatusType`. The end result would be one file (BatchJobStatusType.java), which would be declared as `public enum BatchJobStatusType`, similar to how the old `JobStatus` class was declared as an enum and had methods.

Actually, I think you can get the same result by basically taking the old `JobStatus` class and just renaming it to `BatchJobStatusType`.

You can't combine `StreamingJobStatusType` into `StreamingJobStatus` though. Because `StreamingJobStatus` has `statusDetail` it needs to be a class (although if `statusDetail` was a hardcoded string that would be a different story - but we shouldn't do that).
</blockquote></details>

I'm not sure if the instruction here is to keep BatchJobStatus class (which defines methods and contains BatchJobStatusEnum type) to be consistent with StreamingJobStatus class (which defines methods and contains StreamingJobStatusEnum type plus detail String) . I removed JobStatus class, and cleaned up BatchJobStatus and StreamingJobStatus as recommened - but I'm not sure if this comment is a request to remove BatchJobStatus class, and just keep StreamingJobStatus class plus StreamingJobStatusType enum and BatchJobStatusType enum. Can you clarify?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/433)*

jrobble:
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/persistent/BatchJobStatus.java, line 38 at r3](https://reviewable.io:443/reviews/openmpf/openmpf/433#-L46iFTOnpl5_TW4HHzw:-L4Co9DYVb7ypFn8P94k:b-c06575) ([raw file](https://github.com/openmpf/openmpf/blob/c84d4516a287972b5d8e67e01a2fc02187fffcf7/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/persistent/BatchJobStatus.java#L38)):*

I'm not sure if the instruction here is to keep BatchJobStatus class (which defines methods and contains BatchJobStatusEnum type) to be consistent with StreamingJobStatus class (which defines methods and contains StreamingJobStatusEnum type plus detail String) . I removed JobStatus class, and cleaned up BatchJobStatus and StreamingJobStatus as recommened - but I'm not sure if this comment is a request to remove BatchJobStatus class, and just keep StreamingJobStatus class plus StreamingJobStatusType enum and BatchJobStatusType enum. Can you clarify?
</blockquote></details>

Sure. I don't believe that there is a need for `BatchJobStatus` and `BatchJobStatusType`. You can define all the data members and behaviors in `BatchJobStatusType`. It will be defined as an `enum`, rather than a `class`, much like `JobStatus` is currently defined in the develop branch.

I understand that you split `BatchJobStatus` and `BatchJobStatusType` for consistency with what we're doing for streaming, but I don't think it's required. Personally, I would prefer to deal with one less class if we can.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/433)*

dpierce17:
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/businessrules/StreamingJobRequestBo.java, line 98 at r3](https://reviewable.io:443/reviews/openmpf/openmpf/433#-L474kds4Dls9AwtVh30:-L4CqK_v4S-aSOsSiUiu:b-5568r5) ([raw file](https://github.com/openmpf/openmpf/blob/c84d4516a287972b5d8e67e01a2fc02187fffcf7/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/businessrules/StreamingJobRequestBo.java#L98)):*

It seems like this should also take `StreamingJobStatus` (instead of `StreamingJobStatusType`). I think that we should commit to using `StreamingJobStatus` everywhere over `StreamingJobStatusType`, even if the inner detail string is empty or null.
</blockquote></details>

I made this change. Note that handleJobStatusChange is called from StreamingJobRoutesBuilder. Added a TODO to StreamingJobRoutesBuilder with recommendation to add JOB_STATUS_DETAIL to the streaming job status message. For future reference, what is the procedure for this? When I add TODOs, should I also follow up with adding an openmpf issue or does the reviewer evaluate the TODO first?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/433)*

dpierce17:
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/businessrules/impl/StreamingJobRequestBoImpl.java, line 492 at r5](https://reviewable.io:443/reviews/openmpf/openmpf/433#-L47KScq0xwIkMl4pqxf:-L4Cy6VlYF6l4rgE1jB-:b-9z36d3) ([raw file](https://github.com/openmpf/openmpf/blob/8a8bf9034421370a23b60a4c292c041ed19f844b/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/businessrules/impl/StreamingJobRequestBoImpl.java#L492)):*

You should instead catch `InvalidPipelineObjectWfmProcessingException` here and handle that differently.
</blockquote></details>

The only handling that is different is the error message that is produced. When the pipeline is found to be invalid during transient object creation, the streaming job is marked with a JOB_CREATION_ERROR status in REDIS then the method will throw the InvalidPipelineObjectWfmProcessingException. When the InvalidPipelineObjectWfmProcessingException is caught (or any other exception for that matter), the streaming job must still be marked with the JOB_CREATION_ERROR status in the long-term database. Revised comments to better call this out.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/433)*

jrobble:
Reviewed 1 of 46 files at r3, 19 of 20 files at r6.
Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/businessrules/StreamingJobRequestBo.java, line 98 at r3](https://beta.reviewable.io/reviews/openmpf/openmpf/433#-L474kds4Dls9AwtVh30:-L4HNrWTTh7bdBGEQWwY:b8ckxzi) ([raw file](https://github.com/openmpf/openmpf/blob/c84d4516a287972b5d8e67e01a2fc02187fffcf7/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/businessrules/StreamingJobRequestBo.java#L98)):*
I spent yesterday working on #348 and #334 and addressed many, if not all, of the TODOs you and Brian have left so far. Once you land this diff, I will merge my changes with yours and address your new TODO.

> When I add TODOs, should I also follow up with adding an openmpf issue or does the reviewer evaluate the TODO first?

Generally, if you find something completely unrelated to the task at hand, like a bug, you can go ahead and create a new issue for it if it's not already on the workboard. It may be worth checking in with me first to see if that's the case. Sometimes issues aren't named what you think they should be.

If the TODO is related to what you're working on, or another issue already on the work board, you should check in with me. We'll decide if we should drop a comment on an existing issue or create a new one. We should avoid creating too many small issues because that adds a lot of clutter.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/businessrules/impl/StreamingJobRequestBoImpl.java, line 497 at r6](https://beta.reviewable.io/reviews/openmpf/openmpf/433#-L4HhplzF9o6eWl1gzno:-L4HhplzF9o6eWl1gznp:b-uadhub) ([raw file](https://github.com/openmpf/openmpf/blob/c5078fac0778c9f37274270894faedfdfa9e37b3/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/businessrules/impl/StreamingJobRequestBoImpl.java#L497)):*
> ```Java
>             // Note that it doesn't matter what exception occurred, the streaming job has to be marked as failed in the long-term database.
>             String errorMessage;
>             if ( e instanceof InvalidPipelineObjectWfmProcessingException ) {
> ```

I would prefer that we not catch the generic `Exception` and then use `instanceof` to differentiate between the type of exceptions. From  a coding practice point of view, that doesn't feel right.

I would prefer that we move `errorMessage` above the `try` and set it to null. Catch `InvalidPipelineObjectWfmProcessingException` before `Exception` using separate `catch` statements. In each catch block set the `errorMessage` and nothing more. Then, after the catch blocks, check if `errorMessage` is null. If not, perform the rest of the logic in the current catch block, resulting in throwing `WfmProcessingException`.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/RedisImpl.java, line 928 at r3](https://beta.reviewable.io/reviews/openmpf/openmpf/433#-L478Ey9nuek3jdu_Eln:-L4Hjb7U24C6jVyZ6rWV:b-8nbjv9) ([raw file](https://github.com/openmpf/openmpf/blob/c84d4516a287972b5d8e67e01a2fc02187fffcf7/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/RedisImpl.java#L928)):*

I realize I don't need the else, but I like it for readability.
</blockquote></details>

Ok, you can leave it if that's your preference. Personally, I prefer to refactor my code so that I can move all of the stuff that throws an exception or returns quickly to the top of the method. That enables me to get rid of some of the nesting, which I think is a bit easier to work with.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/RedisImpl.java, line 491 at r6](https://beta.reviewable.io/reviews/openmpf/openmpf/433#-L4HuzYN8Dmc985TYZA_:-L4HuzYN8Dmc985TYZAa:bslr7j1) ([raw file](https://github.com/openmpf/openmpf/blob/c5078fac0778c9f37274270894faedfdfa9e37b3/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/RedisImpl.java#L491)):*
> ```Java
>                 Map jobHash = redisTemplate.boundHashOps(key(STREAMING_JOB, jobId)).entries();
>                 StreamingJobStatusType jobStatusType = (StreamingJobStatusType) jobHash.get(STREAMING_JOB_STATUS);
>                 String jobStatusDetail = (String) jobHash.get(STREAMING_JOB_STATUS_DETAIL);
> ```

Don't bother getting the detail if the type is null because you never use it.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/enums/BatchJobStatusType.java, line 8 at r6](https://beta.reviewable.io/reviews/openmpf/openmpf/433#-L4HM4XvjZ05UJx_hT8p:-L4HM4XwpUyI_h9g0HEs:b-5atzxi) ([raw file](https://github.com/openmpf/openmpf/blob/c5078fac0778c9f37274270894faedfdfa9e37b3/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/enums/BatchJobStatusType.java#L8)):*
> ```Java
>  * 52.227-14, Alt. IV (DEC 2007).                                             *
>  *                                                                            *
>  * Copyright 2017 The MITRE Corporation. All Rights Reserved.                 *
> ```

Please update this copyright and the one below to 2018.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/enums/StreamingJobStatusType.java, line 30 at r3](https://beta.reviewable.io/reviews/openmpf/openmpf/433#-L46jHROmB6LhSiDQHns:-L4HNcZfCOsiGNFCDDYl:b-srry6a) ([raw file](https://github.com/openmpf/openmpf/blob/c84d4516a287972b5d8e67e01a2fc02187fffcf7/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/enums/StreamingJobStatusType.java#L30)):*

Revising this, we'll be keeping UNKNOWN - I'm changing software to use UNKNOWN as the initial job status for streaming jobs.
</blockquote></details>

That makes sense. I feel like that should be the case for the batch code too, but let's leave that as is.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/enums/StreamingJobStatusType.java, line 8 at r6](https://beta.reviewable.io/reviews/openmpf/openmpf/433#-L4HMC0w-1A9urDgO35G:-L4HMC0w-1A9urDgO35H:b-5atzxi) ([raw file](https://github.com/openmpf/openmpf/blob/c5078fac0778c9f37274270894faedfdfa9e37b3/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/enums/StreamingJobStatusType.java#L8)):*
> ```Java
>  * 52.227-14, Alt. IV (DEC 2007).                                             *
>  *                                                                            *
>  * Copyright 2017 The MITRE Corporation. All Rights Reserved.                 *
> ```

Please update this copyright and the one below to 2018.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/enums/StreamingJobStatusType.java, line 73 at r6](https://beta.reviewable.io/reviews/openmpf/openmpf/433#-L4HNh0Ec4uTmxEInOVl:-L4HNh0Ec4uTmxEInOVm:bj5hbfv) ([raw file](https://github.com/openmpf/openmpf/blob/c5078fac0778c9f37274270894faedfdfa9e37b3/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/enums/StreamingJobStatusType.java#L73)):*
> ```Java
>     ERROR(true),
> 
>     // This section defines job statuses that may be applicable only to a streaming job.
> ```

You can remove this comment.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/enums/StreamingJobStatusType.java, line 78 at r6](https://beta.reviewable.io/reviews/openmpf/openmpf/433#-L4HN3emKc3o5-0VngM7:-L4HN3en2HPE3k-AAEqx:bitef0u) ([raw file](https://github.com/openmpf/openmpf/blob/c5078fac0778c9f37274270894faedfdfa9e37b3/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/enums/StreamingJobStatusType.java#L78)):*
> ```Java
>      * Indicates the streaming job was terminated.
>      */
>     STREAMING_JOB_TERMINATED(true),
> ```

Since this list to specific to streaming, please remove the "STREAMING_JOB_" prefix from the last three entries.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/openmpf/openmpf/433)*

dpierce17:
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/RedisImpl.java, line 928 at r3](https://reviewable.io:443/reviews/openmpf/openmpf/433#-L478Ey9nuek3jdu_Eln:-L4LseRSmVFvSEoBVKlD:b4tcyx2) ([raw file](https://github.com/openmpf/openmpf/blob/c84d4516a287972b5d8e67e01a2fc02187fffcf7/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/RedisImpl.java#L928)):*

Ok, you can leave it if that's your preference. Personally, I prefer to refactor my code so that I can move all of the stuff that throws an exception or returns quickly to the top of the method. That enables me to get rid of some of the nesting, which I think is a bit easier to work with.
</blockquote></details>

Ok, I do see that type of structure in some of the legacy portions of the RedisImpl code. I agree we should stay consistent. We could make this a wishlist issue if you like. I'd rather not make this change right now since I think this PR is close to being ready to land.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/433)*

jrobble:
Reviewed 1 of 45 files at r3, 6 of 6 files at r7.
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/RedisImpl.java, line 928 at r3](https://reviewable.io:443/reviews/openmpf/openmpf/433#-L478Ey9nuek3jdu_Eln:-L4MC1uhYLQADSwaU0A8:b-4ps3fi) ([raw file](https://github.com/openmpf/openmpf/blob/c84d4516a287972b5d8e67e01a2fc02187fffcf7/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/RedisImpl.java#L928)):*

Ok, I do see that type of structure in some of the legacy portions of the RedisImpl code. I agree we should stay consistent. We could make this a wishlist issue if you like. I'd rather not make this change right now since I think this PR is close to being ready to land.
</blockquote></details>

This is something we can do moving forward as we write new code, and see the need while working with pieces of old code. In general, don't think it's worth creating wishlist issues for updating the code style since they will never be a priority.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/enums/StreamingJobStatusType.java, line 76 at r7](https://reviewable.io:443/reviews/openmpf/openmpf/433#-L4MDLSQX6-dvybZ0lc4:-L4MDLSQX6-dvybZ0lc5:b-30mlky) ([raw file](https://github.com/openmpf/openmpf/blob/f3949e44dffa0cf13d3855b98ddf88604d0a75c9/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/enums/StreamingJobStatusType.java#L76)):*
> ```Java
>      * Indicates the streaming job was terminated.
>      */
>     JOB_TERMINATED(true),
> ```

Please remove the "JOB_" part of the prefixes as well. None of the other enums have that.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/433)*

jrobble:
Reviewed 1 of 45 files at r3, 3 of 3 files at r8.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/433)*

PR #438, opened 2018-02-02
Add stall detection to streaming component executor

brosenberg42:
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/438)


jrobble:
Reviewed 13 of 13 files at r1.
Review status: all files reviewed at latest revision, 10 unresolved discussions.

---

*[trunk/detection/executor/cpp/streaming/ExecutorUtils.h, line 57 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/438#-L4NTQ1PRwE2AkMyPXKq:-L4NTQ1PRwE2AkMyPXKr:b-3k4l0v) ([raw file](https://github.com/openmpf/openmpf/blob/256c20226d5a8a691119ca8409847f62ef3463c6/trunk/detection/executor/cpp/streaming/ExecutorUtils.h#L57)):*
> ```objc
> 
>         sleep_duration_t back_off_duration = milliseconds(1);
>         sleep_duration_t max_back_off = seconds(30);
> ```

This 30 should be declared as a constant at the top of this file.

---

*[trunk/detection/executor/cpp/streaming/ExecutorUtils.h, line 60 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/438#-L4akkgm_sJH0kKCP-g9:-L4akkgndOA7zqCQpZpA:bamm766) ([raw file](https://github.com/openmpf/openmpf/blob/256c20226d5a8a691119ca8409847f62ef3463c6/trunk/detection/executor/cpp/streaming/ExecutorUtils.h#L60)):*
> ```objc
> 
>         while (true) {
>             sleep_duration_t time_left = end_time - steady_clock::now();
> ```

Please leave a comment here explaining how `sleep_duration_t` is implementation-specific. It's nanoseconds in CentOS. Also, state that we use the most precise granularity (smallest unit of measure) to prevent oversleeping.

---

*[trunk/detection/executor/cpp/streaming/ExecutorUtils.h, line 89 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/438#-L4NUTeUYJWNs--UIn4k:-L4NUTeUYJWNs--UIn4l:b-p4pqye) ([raw file](https://github.com/openmpf/openmpf/blob/256c20226d5a8a691119ca8409847f62ef3463c6/trunk/detection/executor/cpp/streaming/ExecutorUtils.h#L89)):*
> ```objc
> 
>         sleep_duration_t back_off_duration = milliseconds(1);
>         sleep_duration_t max_back_off = seconds(30);
> ```

This 30 should also be a constant.

---

*[trunk/detection/executor/cpp/streaming/JobSettings.cpp, line 55 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/438#-L4adap91gGOylEMd9oK:-L4adap91gGOylEMd9oL:bzgln4y) ([raw file](https://github.com/openmpf/openmpf/blob/256c20226d5a8a691119ca8409847f62ef3463c6/trunk/detection/executor/cpp/streaming/JobSettings.cpp#L55)):*
Instead of using the more generic form `std::chrono::duration<TDurRep, TDurPeriod>` in the function declarations, would it be possible to use `std::chrono::milliseconds` instead?

Per our conversation, you stated:

> I think we should leave templates on the code in ExecutorUtils, but change StreamingVideoCapture to use milliseconds.

I'm on board with that.

---

*[trunk/detection/executor/cpp/streaming/main.cpp, line 180 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/438#-L4aWQnx5n2j-mV8I0Zj:-L4aWQnx5n2j-mV8I0Zk:bdgj057) ([raw file](https://github.com/openmpf/openmpf/blob/256c20226d5a8a691119ca8409847f62ef3463c6/trunk/detection/executor/cpp/streaming/main.cpp#L180)):*
> ```cpp
>                 cv::Mat frame;
>                 if (!ReadFrame(video_capture, frame)) {
>                     break;
> ```

I think this `break` will only be executed when the quit command is received on standard in. If so, for clarity, please leave a comment here stating that. Also, in the comment please state that otherwise if the frame cannot be read than an exception will be thrown.

---

*[trunk/detection/executor/cpp/streaming/main.cpp, line 240 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/438#-L4NX3de8227b3MMd1Kk:-L4NX3de8227b3MMd1Kl:bjgrcb1) ([raw file](https://github.com/openmpf/openmpf/blob/256c20226d5a8a691119ca8409847f62ef3463c6/trunk/detection/executor/cpp/streaming/main.cpp#L240)):*
> ```cpp
> 
>             bool was_read;
>             if (settings_.stall_timeout < std::chrono::milliseconds::zero()) {
> ```

Please leave a comment here explaining the special meaning when the handling a timeout value < 0.

---

*[trunk/detection/executor/cpp/streaming/main.cpp, line 254 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/438#-L4NVrF5rmX50DQenPdE:-L4NVrF6Q9Yb6_moV0fC:bzfh7yr) ([raw file](https://github.com/openmpf/openmpf/blob/256c20226d5a8a691119ca8409847f62ef3463c6/trunk/detection/executor/cpp/streaming/main.cpp#L254)):*
> ```cpp
>             throw FatalError(ExitCode::STREAM_STALLED, "It is no longer possible to read frames.");
>         }
>         catch (const InterruptedException& ex) {
> ```

At first glance, I was thinking that `InterruptedException` could be thrown because of receiving the quit command on the command line, or because of a timeout.

Looking closer at the code, I see that this exception is only thrown in the first case. I think we should change the name of the exception to be more clear. Consider `InterruptedByStandardInException`.

---

*[trunk/detection/executor/cpp/streaming/StreamingVideoCapture.cpp, line 78 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/438#-L4aXaEsyFLlB8wzE1Jr:-L4aXaEsyFLlB8wzE1Js:b-orncnv) ([raw file](https://github.com/openmpf/openmpf/blob/256c20226d5a8a691119ca8409847f62ef3463c6/trunk/detection/executor/cpp/streaming/StreamingVideoCapture.cpp#L78)):*
> ```cpp
>         LOG4CXX_WARN(logger_, "Successfully re-connected to video stream.");
> 
>         bool was_read = cvVideoCapture_.read(frame);
> ```

Consider using `Read()` here.

---

*[trunk/detection/executor/cpp/streaming/test/test_streaming_executor.cpp, line 194 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/438#-L4afdV6rxfSjzdmR2Xa:-L4afdV6rxfSjzdmR2Xb:b-sa2zof) ([raw file](https://github.com/openmpf/openmpf/blob/256c20226d5a8a691119ca8409847f62ef3463c6/trunk/detection/executor/cpp/streaming/test/test_streaming_executor.cpp#L194)):*
> ```cpp
>     ASSERT_FALSE(retry_result);
> 
>     ASSERT_EQ(count, 9);
> ```

Please leave a comment that explains where 9 came from:

(1 ms + 2 ms + 4 ms + 8 ms + 16 ms + 32 ms + 64 ms + 128 ms) + 245 ms = 500 ms

Where 245 is equal to 500 minus everything in (). That totals 9 iterations. 

---

*[trunk/detection/executor/cpp/streaming/test/test_streaming_executor.cpp, line 220 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/438#-L4ajw5Hb8rbo-ij2k9Y:-L4ajw5Hb8rbo-ij2k9Z:b-xby4i4) ([raw file](https://github.com/openmpf/openmpf/blob/256c20226d5a8a691119ca8409847f62ef3463c6/trunk/detection/executor/cpp/streaming/test/test_streaming_executor.cpp#L220)):*
> ```cpp
> 
> 
> TEST(StreamingExecutorUtilsTest, RetryWorksWhenFuncTakesTime) {
> ```

Please assert some runtime constraints on this test like you did for `RetryRetriesUntilTimeout`.

155 ms + 1 ms + 155 ms = 311 ms expected runtime

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/438)*

jrobble:
*[trunk/detection/executor/cpp/streaming/main.cpp, line 245 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/438#-L4bFo1V3OtWU6rh4F4W:-L4bFo1WxsgIsbqbNmgI:bz1rpql) ([raw file](https://github.com/openmpf/openmpf/blob/256c20226d5a8a691119ca8409847f62ef3463c6/trunk/detection/executor/cpp/streaming/main.cpp#L245)):*
> ```cpp
>             }
>             else {
>                 was_read = video_capture.ReadWithRetry(frame, settings_.stall_timeout);
> ```

I think this calculation should be `settings_.stall_timeout - settings_.stall_alert_threshold`. Sending the alert should not affect the overall timeout period. We should check to make sure we're not subtracting a negative from a negative. It may may the most sense to force those params to 0 after being read from the .ini file if they are negative.

For example, right now, if the alert threshold is 5 seconds, and the overall timeout is 30 seconds, it actually takes 35 seconds for the overall timeout. Instead, the timeout should occur 25 seconds after the alert.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/438)*

brosenberg42:
Review status: 3 of 16 files reviewed at latest revision, 12 unresolved discussions.

---

*[trunk/detection/executor/cpp/streaming/ExecutorUtils.h, line 57 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/438#-L4NTQ1PRwE2AkMyPXKq:-L4h21UFjWyOAZEcrNUV:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/256c20226d5a8a691119ca8409847f62ef3463c6/trunk/detection/executor/cpp/streaming/ExecutorUtils.h#L57)):*

This 30 should be declared as a constant at the top of this file.
</blockquote></details>

Done.

---

*[trunk/detection/executor/cpp/streaming/ExecutorUtils.h, line 60 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/438#-L4akkgm_sJH0kKCP-g9:-L4h2238zuKziuGVdIbJ:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/256c20226d5a8a691119ca8409847f62ef3463c6/trunk/detection/executor/cpp/streaming/ExecutorUtils.h#L60)):*

Please leave a comment here explaining how `sleep_duration_t` is implementation-specific. It's nanoseconds in CentOS. Also, state that we use the most precise granularity (smallest unit of measure) to prevent oversleeping.
</blockquote></details>

Done.

---

*[trunk/detection/executor/cpp/streaming/ExecutorUtils.h, line 89 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/438#-L4NUTeUYJWNs--UIn4k:-L4h24lOVMhpRX3rReaR:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/256c20226d5a8a691119ca8409847f62ef3463c6/trunk/detection/executor/cpp/streaming/ExecutorUtils.h#L89)):*

This 30 should also be a constant.
</blockquote></details>

Done.

---

*[trunk/detection/executor/cpp/streaming/JobSettings.cpp, line 55 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/438#-L4adap91gGOylEMd9oK:-L4h28oCiThppSDzVyn6:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/256c20226d5a8a691119ca8409847f62ef3463c6/trunk/detection/executor/cpp/streaming/JobSettings.cpp#L55)):*

Instead of using the more generic form `std::chrono::duration<TDurRep, TDurPeriod>` in the function declarations, would it be possible to use `std::chrono::milliseconds` instead?

Per our conversation, you stated:

> I think we should leave templates on the code in ExecutorUtils, but change StreamingVideoCapture to use milliseconds.

I'm on board with that.
</blockquote></details>

Done.

---

*[trunk/detection/executor/cpp/streaming/main.cpp, line 58 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/438#-L4h38XlhcKGiLRp8F0K:-L4h38XlhcKGiLRp8F0L:b-76o5u8) ([raw file](https://github.com/openmpf/openmpf/blob/256c20226d5a8a691119ca8409847f62ef3463c6/trunk/detection/executor/cpp/streaming/main.cpp#L58)):*
> ```cpp
> 
> 
> class StreamingComponentExecutor {
> ```

The StreamingComponentExecutor class has been moved to its own file.

---

*[trunk/detection/executor/cpp/streaming/main.cpp, line 180 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/438#-L4aWQnx5n2j-mV8I0Zj:-L4h2ou2tKCI5xPJ3ApT:b-k6opo9) ([raw file](https://github.com/openmpf/openmpf/blob/256c20226d5a8a691119ca8409847f62ef3463c6/trunk/detection/executor/cpp/streaming/main.cpp#L180)):*

I think this `break` will only be executed when the quit command is received on standard in. If so, for clarity, please leave a comment here stating that. Also, in the comment please state that otherwise if the frame cannot be read than an exception will be thrown.
</blockquote></details>

This has been changed (in the new file) a little bit to make it more clear that we exit when we get the InteruptedException.

---

*[trunk/detection/executor/cpp/streaming/main.cpp, line 240 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/438#-L4NX3de8227b3MMd1Kk:-L4h4AStRMst5WK51upO:b-fth45t) ([raw file](https://github.com/openmpf/openmpf/blob/256c20226d5a8a691119ca8409847f62ef3463c6/trunk/detection/executor/cpp/streaming/main.cpp#L240)):*

Please leave a comment here explaining the special meaning when the handling a timeout value < 0.
</blockquote></details>

This check is now done in JobSettings. The naming of the boolean variable should make it more clear.

---

*[trunk/detection/executor/cpp/streaming/main.cpp, line 245 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/438#-L4bFo1V3OtWU6rh4F4W:-L4h3Sgw_7IudKw8e5Q_:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/256c20226d5a8a691119ca8409847f62ef3463c6/trunk/detection/executor/cpp/streaming/main.cpp#L245)):*

I think this calculation should be `settings_.stall_timeout - settings_.stall_alert_threshold`. Sending the alert should not affect the overall timeout period. We should check to make sure we're not subtracting a negative from a negative. It may may the most sense to force those params to 0 after being read from the .ini file if they are negative.

For example, right now, if the alert threshold is 5 seconds, and the overall timeout is 30 seconds, it actually takes 35 seconds for the overall timeout. Instead, the timeout should occur 25 seconds after the alert.
</blockquote></details>

Done.

---

*[trunk/detection/executor/cpp/streaming/main.cpp, line 254 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/438#-L4NVrF5rmX50DQenPdE:-L4h3TbpHWtLpJK9LPCU:bqhw797) ([raw file](https://github.com/openmpf/openmpf/blob/256c20226d5a8a691119ca8409847f62ef3463c6/trunk/detection/executor/cpp/streaming/main.cpp#L254)):*

At first glance, I was thinking that `InterruptedException` could be thrown because of receiving the quit command on the command line, or because of a timeout.

Looking closer at the code, I see that this exception is only thrown in the first case. I think we should change the name of the exception to be more clear. Consider `InterruptedByStandardInException`.
</blockquote></details>

I added a comment in the new file. Maybe I don't understand your comment, but I don't think people would expect an InterruptedException to be thrown when there was no interrupt. 

---

*[trunk/detection/executor/cpp/streaming/StreamingVideoCapture.cpp, line 78 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/438#-L4aXaEsyFLlB8wzE1Jr:-L4h2YT2RdgE0EKhdpq7:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/256c20226d5a8a691119ca8409847f62ef3463c6/trunk/detection/executor/cpp/streaming/StreamingVideoCapture.cpp#L78)):*

Consider using `Read()` here.
</blockquote></details>

Done.

---

*[trunk/detection/executor/cpp/streaming/test/test_streaming_executor.cpp, line 194 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/438#-L4afdV6rxfSjzdmR2Xa:-L4h330Vcixc4yBeYOJ5:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/256c20226d5a8a691119ca8409847f62ef3463c6/trunk/detection/executor/cpp/streaming/test/test_streaming_executor.cpp#L194)):*

Please leave a comment that explains where 9 came from:

(1 ms + 2 ms + 4 ms + 8 ms + 16 ms + 32 ms + 64 ms + 128 ms) + 245 ms = 500 ms

Where 245 is equal to 500 minus everything in (). That totals 9 iterations. 
</blockquote></details>

Done.

---

*[trunk/detection/executor/cpp/streaming/test/test_streaming_executor.cpp, line 220 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/438#-L4ajw5Hb8rbo-ij2k9Y:-L4h33lSCR6jagxyFx5V:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/256c20226d5a8a691119ca8409847f62ef3463c6/trunk/detection/executor/cpp/streaming/test/test_streaming_executor.cpp#L220)):*

Please assert some runtime constraints on this test like you did for `RetryRetriesUntilTimeout`.

155 ms + 1 ms + 155 ms = 311 ms expected runtime
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/438)*

jrobble:
Reviewed 13 of 13 files at r2.
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[trunk/detection/executor/cpp/streaming/JobSettings.cpp, line 41 at r2](https://reviewable.io:443/reviews/openmpf/openmpf/438#-L4iRglpSau0ByjMRzJI:-L4iRglqnP8GHrUoE5S9:b-no2tyd) ([raw file](https://github.com/openmpf/openmpf/blob/1ddc170232ff39ebd4882d9066799ec2e0c482b0/trunk/detection/executor/cpp/streaming/JobSettings.cpp#L41)):*
> ```cpp
>     using namespace std::chrono;
> 
>     namespace {
> ```

Is this block used to ensure only internal linkage, or another reason?

---

*[trunk/detection/executor/cpp/streaming/main.cpp, line 254 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/438#-L4NVrF5rmX50DQenPdE:-L4iMZNdnmUdu5s278Wb:bno433u) ([raw file](https://github.com/openmpf/openmpf/blob/256c20226d5a8a691119ca8409847f62ef3463c6/trunk/detection/executor/cpp/streaming/main.cpp#L254)):*

I added a comment in the new file. Maybe I don't understand your comment, but I don't think people would expect an InterruptedException to be thrown when there was no interrupt. 
</blockquote></details>

I was thinking that the `InterruptedException` might also be thrown due to a timeout. I know that's not the case here, but in theory a timeout could trigger an interrupt exception. 

---

*[trunk/detection/executor/cpp/streaming/test/test_streaming_executor.cpp, line 240 at r2](https://reviewable.io:443/reviews/openmpf/openmpf/438#-L4iPcRZW--oB8_rYVXT:-L4iPcRZW--oB8_rYVXU:bydfqab) ([raw file](https://github.com/openmpf/openmpf/blob/1ddc170232ff39ebd4882d9066799ec2e0c482b0/trunk/detection/executor/cpp/streaming/test/test_streaming_executor.cpp#L240)):*
> ```cpp
> 
>     // Use expected_min_sleep_time with a little bit of wiggle room
>     milliseconds expected_max_sleep_time = expected_min_sleep_time + milliseconds(5);
> ```

You're not doing anything with `expected_max_sleep_time`.

---

*[trunk/detection/executor/cpp/streaming/test/test_streaming_executor.cpp, line 246 at r2](https://reviewable.io:443/reviews/openmpf/openmpf/438#-L4iVxOrhR_-b6_fw8fh:-L4iVxOsQ4TET7204lJr:b4c9lyr) ([raw file](https://github.com/openmpf/openmpf/blob/1ddc170232ff39ebd4882d9066799ec2e0c482b0/trunk/detection/executor/cpp/streaming/test/test_streaming_executor.cpp#L246)):*
> ```cpp
> 
> 
> void verifyRetryStrategyMapping(long stall_timeout, long alert_threshold,
> ```

:thumbsup: 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/438)*

brosenberg42:
Review status: 15 of 17 files reviewed at latest revision, 4 unresolved discussions.

---

*[trunk/detection/executor/cpp/streaming/JobSettings.cpp, line 41 at r2](https://reviewable.io:443/reviews/openmpf/openmpf/438#-L4iRglpSau0ByjMRzJI:-L4k7ttNFTuYPW_SzVyI:b-c5nou) ([raw file](https://github.com/openmpf/openmpf/blob/1ddc170232ff39ebd4882d9066799ec2e0c482b0/trunk/detection/executor/cpp/streaming/JobSettings.cpp#L41)):*

Is this block used to ensure only internal linkage, or another reason?
</blockquote></details>

My intent was to make this a private function. It does cause this function to use internal linkage, but I am using it more like the private visibility modifier. This isn't a static method because that would required me to declare this in the header file which would then mean I have to `#include` the boost libraries in the the header.

---

*[trunk/detection/executor/cpp/streaming/test/test_streaming_executor.cpp, line 240 at r2](https://reviewable.io:443/reviews/openmpf/openmpf/438#-L4iPcRZW--oB8_rYVXT:-L4kYUacCNbd5sTAn6Mr:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/1ddc170232ff39ebd4882d9066799ec2e0c482b0/trunk/detection/executor/cpp/streaming/test/test_streaming_executor.cpp#L240)):*

You're not doing anything with `expected_max_sleep_time`.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/438)*

jrobble:
Reviewed 2 of 2 files at r3.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[trunk/detection/executor/cpp/streaming/JobSettings.cpp, line 41 at r2](https://reviewable.io:443/reviews/openmpf/openmpf/438#-L4iRglpSau0ByjMRzJI:-L4l1_pF_87FPKfnUa9A:b-ad3nbm) ([raw file](https://github.com/openmpf/openmpf/blob/1ddc170232ff39ebd4882d9066799ec2e0c482b0/trunk/detection/executor/cpp/streaming/JobSettings.cpp#L41)):*

My intent was to make this a private function. It does cause this function to use internal linkage, but I am using it more like the private visibility modifier. This isn't a static method because that would required me to declare this in the header file which would then mean I have to `#include` the boost libraries in the the header.
</blockquote></details>

Okay. Please drop a comment about that.

---

*[trunk/detection/test-components/cpp/CplusplusHelloWorld/StreamingHelloWorld.cpp, line 56 at r3](https://reviewable.io:443/reviews/openmpf/openmpf/438#-L4l1OBvxKivuouBir9C:-L4l1OBw0dP4REtaSbf9:bpxtew6) ([raw file](https://github.com/openmpf/openmpf/blob/6b602cc975bbb688451aa924386303da97ce523a/trunk/detection/test-components/cpp/CplusplusHelloWorld/StreamingHelloWorld.cpp#L56)):*
> ```cpp
> bool StreamingHelloWorld::ProcessFrame(const cv::Mat &frame, int frame_number) {
>     LOG4CXX_INFO(hw_logger_, "[" << job_name_ << "] Processing frame with size: " << frame.size())
>     if (frame.empty()) {
> ```

Please drop a comment here stating that this check is here to ensure that a system test fails. In general, this function should never be passed an empty frame.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/438)*

jrobble:
Reviewed 3 of 3 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/438)*

Issue #440, opened 2018-02-06
Add activity frame id and activity timestamp to MySQL streaming_job_request


dpierce17:
To support the [GET] /rest/streaming/jobs/{id} endpoint when a job isn't running, activity frame id and activity timestamp should be added to MySQL (i.e. streaming_job_request table). 

Whenever a new activity alert message is received by WFM, the activity frame id and activity timestamp needs to be updated in both REDIS and MySQL. 

Assumption: WFM should only receive one new activity alert message per segment, so updating mySQL every time isn’t too expensive a process.

Note: when WFM responds to [GET] /rest/streaming/jobs/{id} endpoint, WFM  doesn't need to check REDIS for latest job status, activity frame id and activity timestamp, since that information is already in mysql.

dpierce17:
Landed with PR441

Issue #425, opened 2018-01-26
Remove "last" from lastActivityFrameId and lastActivityTimestamp in health report data


dpierce17:
"last" doesn't adequately describe these parameters. Drop "last" from the fieldnames. What these parameters actually are will be fully described in the REST API.

dpierce17:
Note: dropped "latest" and offered expanded description of the parameters in the most recent MPF-REST-API proposal (dated 1/26/18), but this change is not yet in the software itself.

dpierce17:
Revising this task to also include job status detail in the Health report data

dpierce17:
Landed with PR441

PR #441, opened 2018-02-06
Update streaming job info, HealthReportData and StreamingJobInfo

dpierce17:
This pull request covers openmpf issues #411, #425 and #440

---
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/441)


jrobble:
Please merge with develop. 

You will see that currently the WFM code in the develop branch sets the health report jobStatus string to `<type> : <detail>`. It sets the jobStatus in the StreamingJobInfo in the same way.

In retrospect, I like the idea of keeping the detail as a separate field in the health report and StreamingJobInfo. I was thinking that just before taking a look at this PR. I'm happy to see that we're of a like mind.

I think that keeping them as separate fields will make the health reports and StreamingJobInfo response more machine readable. Programs will care more about the status type while humans may take interest in the detail. Keeping them separate means that a third party program will not have to parse out the type from a concatenated string.

---

Review status: 0 of 10 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/441#-:-L4imCOC3vhQq9WnY7XB:b-ng4eef)*

dpierce17:
This branch has been merged with develop, and I fixed a bug in ITWebStreamingHealthReports test that I found to exist after the merge with develop

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/441#-:-L4m6u7Er4RcAi_Mc7_R:b-ysnzk)*

jrobble:
Reviewed 10 of 10 files at r1.

---

*[trunk/interop/src/main/java/org/mitre/mpf/interop/JsonHealthReportCollection.java, line 58 at r1](https://beta.reviewable.io/reviews/openmpf/openmpf/441#-L4nU5KJEOQvB16p6SIo:-L4nU5KJEOQvB16p6SIp:b-4p8fo2) ([raw file](https://github.com/openmpf/openmpf/blob/977dd0746c7d3d8ad30b7e7850b1cbb967f2e29a/trunk/interop/src/main/java/org/mitre/mpf/interop/JsonHealthReportCollection.java#L58)):*
> ```Java
> 
>         public Long getJobId() { return jobId; }
>         public String getExternalId() {
> ```

The following 4 getters return "" (the blank string) instead of `null`. Looking at some of the batch JSON output objects, and some of the other JSON we return (such as the StreamingJobInfo), I would prefer if these returned `null` as well. 

---

*[trunk/interop/src/main/java/org/mitre/mpf/interop/JsonHealthReportCollection.java, line 90 at r1](https://beta.reviewable.io/reviews/openmpf/openmpf/441#-L4mKU_a3KAloE9TYFKd:-L4mKU_b60CnmYnxHRRC:b1bep93) ([raw file](https://github.com/openmpf/openmpf/blob/977dd0746c7d3d8ad30b7e7850b1cbb967f2e29a/trunk/interop/src/main/java/org/mitre/mpf/interop/JsonHealthReportCollection.java#L90)):*
> ```Java
>         @JsonCreator
>         public JsonHealthReport(@JsonProperty("jobId") Long jobId, @JsonProperty("externalId") String externalId,
>             @JsonProperty("jobStatus") String jobStatus, @JsonProperty("jobStatusDetail") String jobStatusDetail,
> ```

Let's list each property on its own line.

---

*[trunk/interop/src/main/java/org/mitre/mpf/interop/JsonHealthReportCollection.java, line 162 at r1](https://beta.reviewable.io/reviews/openmpf/openmpf/441#-L4mMOIa9kDB-ll7pd23:-L4mMOIb54sHA6fXQMUa:b-52hvo2) ([raw file](https://github.com/openmpf/openmpf/blob/977dd0746c7d3d8ad30b7e7850b1cbb967f2e29a/trunk/interop/src/main/java/org/mitre/mpf/interop/JsonHealthReportCollection.java#L162)):*
> ```Java
>      * The activity frame ids for each streaming job in this health report.
>      * @return activity frame ids associated with each streaming job. Values within the List will be emtpy String if
>      * there has not been any activity in this streaming job.
> ```

We've shifted away from mentioning "new activity alert" alert in this file, but there are still a few places where it's mentioned. Please search for that term (case insensitive) and update the comments.

---

*[trunk/mpf-rest-api/src/main/java/org/mitre/mpf/rest/api/StreamingJobInfo.java, line 31 at r1](https://beta.reviewable.io/reviews/openmpf/openmpf/441#-L4nbgXSEQ2wL2RJJRe_:-L4nbgXT1f04eW2-7uN3:b-qbnkd4) ([raw file](https://github.com/openmpf/openmpf/blob/977dd0746c7d3d8ad30b7e7850b1cbb967f2e29a/trunk/mpf-rest-api/src/main/java/org/mitre/mpf/rest/api/StreamingJobInfo.java#L31)):*
> ```Java
> import java.util.Date;
> 
> public class StreamingJobInfo {
> ```

I think this was discussed before, but the example timestamps that show up in Swagger are misleading when you click on "Model Schema":

```
{
  "activityFrameId": "string",
  "activityTimestamp": "2018-02-08T03:42:45.405Z",
  "endDate": "2018-02-08T03:42:45.405Z",
  "jobId": 0,
  "jobStatus": "string",
  "jobStatusDetail": "string",
  "outputObjectDirectory": "string",
  "pipelineName": "string",
  "startDate": "2018-02-08T03:42:45.405Z",
  "streamUri": "string",
  "terminal": true
}
```

For example: `"startDate": "2018-02-08T03:42:45.405Z"` is actually `"startDate": 1518060150000` in the true JSON output.

First, I think that we want the true JSON output to show a formatted timestamp. Refer to the report date in the segment summary report for how I did that [here](https://github.com/openmpf/openmpf/blob/develop/trunk/interop/src/main/java/org/mitre/mpf/interop/JsonSegmentSummaryReport.java#L51).

Second, our formatted timestamps take the form "2018-02-07 22:23:04.6", which is different from that shown in the Swagger Model Schema.

To address this, I think we should try annotating this StreamingJobInfo class with `@ApiModel` (see [here](https://github.com/swagger-api/swagger-core/wiki/Annotations#apimodel)) and then annotate the date / timestamp fields with `@ApiModelProperty` (see [here](https://github.com/swagger-api/swagger-core/wiki/Annotations#apimodelproperty)).

`@ApiModelProperty` seems to support an `example` field (see [here](https://stackoverflow.com/q/40989038)). We should try `example = 2018-02-07 22:23:04.6` for those date / timestamp fields.

---

*[trunk/mpf-rest-api/src/main/java/org/mitre/mpf/rest/api/StreamingJobInfo.java, line 47 at r1](https://beta.reviewable.io/reviews/openmpf/openmpf/441#-L4mMrVOD6-LWQUhn6rn:-L4mMrVOD6-LWQUhn6ro:b-i79uon) ([raw file](https://github.com/openmpf/openmpf/blob/977dd0746c7d3d8ad30b7e7850b1cbb967f2e29a/trunk/mpf-rest-api/src/main/java/org/mitre/mpf/rest/api/StreamingJobInfo.java#L47)):*
> ```Java
> 		return jobStatus;
> 	}
> 	public void setJobStatus(String /*JobStatus*/ jobStatus) { this.jobStatus = jobStatus; }
> ```

I'm not sure why we have `/*JobStatus*/` here. If we don't need it, please do a search across the openmpf repo and remove instances of that comment used in a similar way.

---

*[trunk/mpf-rest-api/src/main/java/org/mitre/mpf/rest/api/StreamingJobInfo.java, line 47 at r1](https://beta.reviewable.io/reviews/openmpf/openmpf/441#-L4mNUY_9nffaougETny:-L4mNUY_9nffaougETnz:boxd0ek) ([raw file](https://github.com/openmpf/openmpf/blob/977dd0746c7d3d8ad30b7e7850b1cbb967f2e29a/trunk/mpf-rest-api/src/main/java/org/mitre/mpf/rest/api/StreamingJobInfo.java#L47)):*
> ```Java
> 		return jobStatus;
> 	}
> 	public void setJobStatus(String /*JobStatus*/ jobStatus) { this.jobStatus = jobStatus; }
> ```

I don't think there is a need for setters in this class since everything is set in the constructor and this class won't be modified once it's instantiated. Most of the data members here don't have setters.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/mvc/controller/StreamingJobController.java, line 275 at r1](https://beta.reviewable.io/reviews/openmpf/openmpf/441#-L4mNu8f81bYE3ht71kT:-L4mNu8f81bYE3ht71kU:b-1m30a0) ([raw file](https://github.com/openmpf/openmpf/blob/977dd0746c7d3d8ad30b7e7850b1cbb967f2e29a/trunk/workflow-manager/src/main/java/org/mitre/mpf/mvc/controller/StreamingJobController.java#L275)):*
> ```Java
>         List<StreamingJobInfo> jobInfoList = new ArrayList<StreamingJobInfo>();
>         try {
>             List<StreamingJobRequest> job_requests = new ArrayList<StreamingJobRequest>();
> ```

Please use camelcase for your variable names in Java. Using underscores is a C / C++ thing.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/mvc/util/ModelUtils.java, line 148 at r1](https://beta.reviewable.io/reviews/openmpf/openmpf/441#-L4nWbdb366L9n9hr2aG:-L4nWbdcEUC-OaE7AVHk:bdxeie5) ([raw file](https://github.com/openmpf/openmpf/blob/977dd0746c7d3d8ad30b7e7850b1cbb967f2e29a/trunk/workflow-manager/src/main/java/org/mitre/mpf/mvc/util/ModelUtils.java#L148)):*
> ```Java
>                 streamingJobRequest.getPriority(),
> 				streamingJobStatus.getType().name(),
> 				streamingJobStatus.getDetail(),
> ```

I don't think that the status detail will ever be populated because of the way `streamingJobStatus` is constructed above using only the status type:

```
new StreamingJobStatus(streamingJobRequest.getStatus());
```

I don't think there's a reason to construct StreamingJobStatus here. Just pass the type and detail into the StreamingJobInfo constructor directly from the streamingJobRequest, like the other parameters.

Also, in StreamingJobRequest.java I see this:

```
     * Note that we keep the status enumeration and statusString as separate parameters for persisting in the
     * database.
```

I think you mean `statusDetail`, not `statusString`.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/mvc/util/ModelUtils.java, line 154 at r1](https://beta.reviewable.io/reviews/openmpf/openmpf/441#-L4nYNUyCQX7lWlfmzUr:-L4nYNUyCQX7lWlfmzUs:bsuy7g3) ([raw file](https://github.com/openmpf/openmpf/blob/977dd0746c7d3d8ad30b7e7850b1cbb967f2e29a/trunk/workflow-manager/src/main/java/org/mitre/mpf/mvc/util/ModelUtils.java#L154)):*
> ```Java
>                 streamingJobRequest.getOutputObjectDirectory(),
>                 streamingJobRequest.getStreamUri(),
> 				streamingJobRequest.getActivityFrameId(),
> ```

The activity frame id and activity timestamp are showing up as `null` when I use Swagger to test the [GET] /rest/streaming/jobs/{id} endpoint. This is probably because those values are never persisted in `StreamingJobRequestBoImpl.handleNewActivityAlert()`.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/CallbackUtils.java, line 116 at r1](https://beta.reviewable.io/reviews/openmpf/openmpf/441#-L4mPa44CF580z3OGvtD:-L4mPa44CF580z3OGvtE:butgr2c) ([raw file](https://github.com/openmpf/openmpf/blob/977dd0746c7d3d8ad30b7e7850b1cbb967f2e29a/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/CallbackUtils.java#L116)):*
As discussed in https://reviewable.io/reviews/openmpf/openmpf/380:

>  I see that as of late you've been starting log comments with method names. I do the same thing when debugging code, but I don't check that in.
>
> That approach is brittle because if the name of the method changes, or the code is refactored into other places, then the log messages have to be updated. Please don't mention the method name in the log comments.
>
> If two locations in the code generate the same log statement, feel free to update the log message to disambiguate the two.

>> Ok, I'll keep that in mind. But, for general services like REDIS or MpfService I would anticipate that in the field some corner cases might crop up which could cause an error to be logged. Not knowing where those errors came from would make resolving that error more difficult. To accommodate for this, I might have to make the error messages more specific.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/CallbackUtils.java, line 119 at r1](https://beta.reviewable.io/reviews/openmpf/openmpf/441#-L4mOwQ9B5QLWs96nIAT:-L4mOwQ9B5QLWs96nIAU:b-trzsc8) ([raw file](https://github.com/openmpf/openmpf/blob/977dd0746c7d3d8ad30b7e7850b1cbb967f2e29a/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/CallbackUtils.java#L119)):*
> ```Java
>         List<String> externalIds = redis.getExternalIds(jobIds);
>         List<StreamingJobStatus> streamingJobStatuses = redis.getStreamingJobStatuses(jobIds);
>         List<String> jobStatuses = streamingJobStatuses.stream().map( jobStatus -> jobStatus.getType().name()).collect(Collectors.toList());
> ```

Let's call this `jobStatusTypes` to be clear.

Also, note that `Redis.getStreamingJobStatusesAsStrings()` is no longer used. Please remove it.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/CallbackUtils.java, line 129 at r1](https://beta.reviewable.io/reviews/openmpf/openmpf/441#-L4mP1EK-ADEhNGUWkg_:-L4mP1EK-ADEhNGUWkga:b-28dyn2) ([raw file](https://github.com/openmpf/openmpf/blob/977dd0746c7d3d8ad30b7e7850b1cbb967f2e29a/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/CallbackUtils.java#L129)):*
> ```Java
>                     activityFrameIds, activityTimestamps);
> 
>             log.info("sendHealthReportCallback: callbackUri = " + callbackUri + ", jsonBody = " + jsonBody);
> ```

This clutters the WFM log. Please remove it or make it a debug() line.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/CallbackUtils.java, line 187 at r1](https://beta.reviewable.io/reviews/openmpf/openmpf/441#-L4mX1qkBn2q7nxJXS6R:-L4mX1ql-zwwTHKwohgQ:b6vmh2t) ([raw file](https://github.com/openmpf/openmpf/blob/977dd0746c7d3d8ad30b7e7850b1cbb967f2e29a/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/CallbackUtils.java#L187)):*
> ```Java
>             });
>         } catch (WfmProcessingException | IllegalArgumentException e) {
>             log.error("Error sending post + " + post + " to callback " + callbackUri + ".", e);
> ```

`HttpPost.toString()` will return:

```
return this.getMethod() + " " + this.getURI() + " " + this.getProtocolVersion();
```

Adding the callbackUri to the error string after that is redundant. 

---

*[trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/ITWebStreamingHealthReports.java, line 127 at r1](https://beta.reviewable.io/reviews/openmpf/openmpf/441#-L4mggWB4z8lYDp9ahwf:-L4mggWC2gvdMnMYqHSK:bzgo70f) ([raw file](https://github.com/openmpf/openmpf/blob/977dd0746c7d3d8ad30b7e7850b1cbb967f2e29a/trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/ITWebStreamingHealthReports.java#L127)):*
> ```Java
> 
>     @Test(timeout = 5 * MINUTES)
>     public void testPostHealthReportCallback() throws Exception {
> ```

Change the name of this to `testPeriodicHealthReportCallback`.

---

*[trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/ITWebStreamingHealthReports.java, line 152 at r1](https://beta.reviewable.io/reviews/openmpf/openmpf/441#-L4maVaSD5cQlATmeLir:-L4maVaT7kzKABBu02eA:b-wxvke7) ([raw file](https://github.com/openmpf/openmpf/blob/977dd0746c7d3d8ad30b7e7850b1cbb967f2e29a/trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/ITWebStreamingHealthReports.java#L152)):*
> ```Java
>             healthReportPostJobId2 = Long.valueOf(obj2.getInt("jobId"));
> 
>             log.info("Streaming job #1 with jobId " + healthReportPostJobId1
> ```

Wouldn't it be better to keep this log line above the stuff for streaming job 2 in case there is an issue deserializing the response for streaming job 2?

---

*[trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/ITWebStreamingHealthReports.java, line 274 at r1](https://beta.reviewable.io/reviews/openmpf/openmpf/441#-L4mcB5gBj9OeSIBhq1s:-L4mcB5gBj9OeSIBhq1t:b-v2jhsg) ([raw file](https://github.com/openmpf/openmpf/blob/977dd0746c7d3d8ad30b7e7850b1cbb967f2e29a/trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/ITWebStreamingHealthReports.java#L274)):*
> ```Java
>                         if (gotHealthReportPostResponseForBothJobs) {
>                             log.info(
>                                 "InHealthCallback: ********  received a health report containing jobs #1 and #2 ************ ");
> ```

Please remove the *'s. Our log content will get pretty wild if we add special flair to each line. 

---

*[trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/ITWebStreamingHealthReports.java, line 302 at r1](https://beta.reviewable.io/reviews/openmpf/openmpf/441#-L4md7AeBFa0oD-Uteu6:-L4md7AeBFa0oD-Uteu7:b-pwp76w) ([raw file](https://github.com/openmpf/openmpf/blob/977dd0746c7d3d8ad30b7e7850b1cbb967f2e29a/trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/ITWebStreamingHealthReports.java#L302)):*
> ```Java
>         JSONObject stream = new JSONObject();
> //        stream.put("streamUri", "rtsp://test/test.mp4");
>         stream.put("streamUri", "rtsp://184.72.239.149/vod/mp4:BigBuckBunny_115k.mov");
> ```

Declare the stream URI as `public static final String` at the top of this file. Put this comment above it: 

```
// Use a public stream from: https://www.wowza.com/demo/rtsp
```

Remove this line:

```
//        stream.put("streamUri", "rtsp://test/test.mp4");
```

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/openmpf/openmpf/441)*

dpierce17:
*[trunk/mpf-rest-api/src/main/java/org/mitre/mpf/rest/api/StreamingJobInfo.java, line 31 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/441#-L4nbgXSEQ2wL2RJJRe_:-L4pxxOc9Bl2l15tN_Vf:b-vm6g96) ([raw file](https://github.com/openmpf/openmpf/blob/977dd0746c7d3d8ad30b7e7850b1cbb967f2e29a/trunk/mpf-rest-api/src/main/java/org/mitre/mpf/rest/api/StreamingJobInfo.java#L31)):*

I think this was discussed before, but the example timestamps that show up in Swagger are misleading when you click on "Model Schema":

```
{
  "activityFrameId": "string",
  "activityTimestamp": "2018-02-08T03:42:45.405Z",
  "endDate": "2018-02-08T03:42:45.405Z",
  "jobId": 0,
  "jobStatus": "string",
  "jobStatusDetail": "string",
  "outputObjectDirectory": "string",
  "pipelineName": "string",
  "startDate": "2018-02-08T03:42:45.405Z",
  "streamUri": "string",
  "terminal": true
}
```

For example: `"startDate": "2018-02-08T03:42:45.405Z"` is actually `"startDate": 1518060150000` in the true JSON output.

First, I think that we want the true JSON output to show a formatted timestamp. Refer to the report date in the segment summary report for how I did that [here](https://github.com/openmpf/openmpf/blob/develop/trunk/interop/src/main/java/org/mitre/mpf/interop/JsonSegmentSummaryReport.java#L51).

Second, our formatted timestamps take the form "2018-02-07 22:23:04.6", which is different from that shown in the Swagger Model Schema.

To address this, I think we should try annotating this StreamingJobInfo class with `@ApiModel` (see [here](https://github.com/swagger-api/swagger-core/wiki/Annotations#apimodel)) and then annotate the date / timestamp fields with `@ApiModelProperty` (see [here](https://github.com/swagger-api/swagger-core/wiki/Annotations#apimodelproperty)).

`@ApiModelProperty` seems to support an `example` field (see [here](https://stackoverflow.com/q/40989038)). We should try `example = 2018-02-07 22:23:04.6` for those date / timestamp fields.

</blockquote></details>

FYI, to support the @Api annotations you requested, I had to add swagger annotations to the mpf-rest-api module.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/441)*

dpierce17:
*[trunk/mpf-rest-api/src/main/java/org/mitre/mpf/rest/api/StreamingJobInfo.java, line 47 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/441#-L4mMrVOD6-LWQUhn6rn:-L4qanYM9YSyKT_19A0J:bimu6ml) ([raw file](https://github.com/openmpf/openmpf/blob/977dd0746c7d3d8ad30b7e7850b1cbb967f2e29a/trunk/mpf-rest-api/src/main/java/org/mitre/mpf/rest/api/StreamingJobInfo.java#L47)):*

I'm not sure why we have `/*JobStatus*/` here. If we don't need it, please do a search across the openmpf repo and remove instances of that comment used in a similar way.
</blockquote></details>

This is legacy commentry, I'll remove it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/441)*

dpierce17:
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/mvc/util/ModelUtils.java, line 154 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/441#-L4nYNUyCQX7lWlfmzUr:-L4r1rWj1MXGH5INby92:b-h1vnw7) ([raw file](https://github.com/openmpf/openmpf/blob/977dd0746c7d3d8ad30b7e7850b1cbb967f2e29a/trunk/workflow-manager/src/main/java/org/mitre/mpf/mvc/util/ModelUtils.java#L154)):*

The activity frame id and activity timestamp are showing up as `null` when I use Swagger to test the [GET] /rest/streaming/jobs/{id} endpoint. This is probably because those values are never persisted in `StreamingJobRequestBoImpl.handleNewActivityAlert()`.
</blockquote></details>

Should the unset value be null or empty string when using [GET] /rest/streaming/jobs/{id} endpoint?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/441)*

dpierce17:
*[trunk/mpf-rest-api/src/main/java/org/mitre/mpf/rest/api/StreamingJobInfo.java, line 31 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/441#-L4nbgXSEQ2wL2RJJRe_:-L4rF8Bb13FNDBtvvIZY:b-kta3j7) ([raw file](https://github.com/openmpf/openmpf/blob/977dd0746c7d3d8ad30b7e7850b1cbb967f2e29a/trunk/mpf-rest-api/src/main/java/org/mitre/mpf/rest/api/StreamingJobInfo.java#L31)):*

FYI, to support the @Api annotations you requested, I had to add swagger annotations to the mpf-rest-api module.
</blockquote></details>

I tried adding these annotations to StreamingJobInfo class, but I saw no change to the Swagger Model Schema:
// swagger includes
@Api(value = "streaming-jobs")
@ApiModel(description="StreamingJobInfo model")
public class StreamingJobInfo {

.
.
.

    @ApiModelProperty(dataType="String", value = "streaming job start time, local time", example = "2018-02-07 10:23:04.6")
	public String getStartDate() {
		return startDate;
	}

any suggestions?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/441)*

jrobble:
Review status: all files reviewed at latest revision, 17 unresolved discussions, some commit checks failed.

---

*[trunk/mpf-rest-api/src/main/java/org/mitre/mpf/rest/api/StreamingJobInfo.java, line 31 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/441#-L4nbgXSEQ2wL2RJJRe_:-L4rGnAvSNixyQHmpG4A:b9q2l0v) ([raw file](https://github.com/openmpf/openmpf/blob/977dd0746c7d3d8ad30b7e7850b1cbb967f2e29a/trunk/mpf-rest-api/src/main/java/org/mitre/mpf/rest/api/StreamingJobInfo.java#L31)):*
> @api(value = "streaming-jobs")
> @apimodel(description="StreamingJobInfo model")

I think the annotations are case sensitive. Those should be camelcase. 

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/mvc/util/ModelUtils.java, line 154 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/441#-L4nYNUyCQX7lWlfmzUr:-L4rGyDcnwhiX4PxuEsw:bhkqqp7) ([raw file](https://github.com/openmpf/openmpf/blob/977dd0746c7d3d8ad30b7e7850b1cbb967f2e29a/trunk/workflow-manager/src/main/java/org/mitre/mpf/mvc/util/ModelUtils.java#L154)):*

Should the unset value be null or empty string when using [GET] /rest/streaming/jobs/{id} endpoint?
</blockquote></details>

We should use null whenever we are generating a JSON value for a Java data member that is null. I can't think of a case where we should return the empty string instead.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/441)*

jrobble:
Review status: all files reviewed at latest revision, 17 unresolved discussions, some commit checks failed.

---

*[trunk/mpf-rest-api/src/main/java/org/mitre/mpf/rest/api/StreamingJobInfo.java, line 31 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/441#-L4nbgXSEQ2wL2RJJRe_:-L4rHQ3IevGJxZUjYQq3:b-4863k1) ([raw file](https://github.com/openmpf/openmpf/blob/977dd0746c7d3d8ad30b7e7850b1cbb967f2e29a/trunk/mpf-rest-api/src/main/java/org/mitre/mpf/rest/api/StreamingJobInfo.java#L31)):*

> @api(value = "streaming-jobs")
> @apimodel(description="StreamingJobInfo model")

I think the annotations are case sensitive. Those should be camelcase. 

</blockquote></details>

Also, did you clear your browser cache before you tried to look at the updated Swagger page?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/441)*

dpierce17:
*[trunk/mpf-rest-api/src/main/java/org/mitre/mpf/rest/api/StreamingJobInfo.java, line 31 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/441#-L4nbgXSEQ2wL2RJJRe_:-L4uypVsIsMb07mNjTzF:bha1l1a) ([raw file](https://github.com/openmpf/openmpf/blob/977dd0746c7d3d8ad30b7e7850b1cbb967f2e29a/trunk/mpf-rest-api/src/main/java/org/mitre/mpf/rest/api/StreamingJobInfo.java#L31)):*

Also, did you clear your browser cache before you tried to look at the updated Swagger page?
</blockquote></details>

Found the problem, this was a usage issue. I didn't realize I had to click "Model" under Repsonse Class (Status 200) to view my changes (default view was Model Schema. Note that usage of @api(value="streaming-jobs" is correct (this is same as in StreamingJobController.  One thing that isn't working though is usage of @ApiModelProperty example=.  I see the value (i.e. description) that I assigned for @ApiModelProperty - but the example I am providing isn't showing up.  According to this reference: http://docs.swagger.io/swagger-core/v1.3.12/apidocs/index.html?com/wordnik/swagger/annotations/ApiModelProperty.html, example isn't a valid element.  FYI, openmpf is using springfox.swagger.version 2.1.1. I haven't been able to find the documentation on that version yet - but maybe the example element isn't supported there either. To resolve, I'll put example time in the value  element.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/441)*

jrobble:
I still need to run this code to test for nulls.

---

Reviewed 14 of 14 files at r2.

---

*[trunk/interop/src/main/java/org/mitre/mpf/interop/JsonHealthReportCollection.java, line 77 at r2](https://beta.reviewable.io/reviews/openmpf/openmpf/441#-L4wX4XI7SPXRRAiz25d:-L4wX4XI7SPXRRAiz25e:b-n4sm81) ([raw file](https://github.com/openmpf/openmpf/blob/53ee4181c508bc1230d02e654eb27f62e309ea18/trunk/interop/src/main/java/org/mitre/mpf/interop/JsonHealthReportCollection.java#L77)):*
> ```Java
>             this.externalId = externalId;
>             this.activityFrameId = activityFrameId;
>             // Activity timestamp should remain as null if updated activity timestamp is passed in as an empty String.
> ```

This is fine for now. We may eventually refactor to have this class accept a Date object as a parameter, much like the JsonSegmentSummaryReport reportDate field.

---

*[trunk/mpf-rest-api/src/main/java/org/mitre/mpf/rest/api/StreamingJobInfo.java, line 31 at r1](https://beta.reviewable.io/reviews/openmpf/openmpf/441#-L4nbgXSEQ2wL2RJJRe_:-L4v_YLa_X58vQHK3Dxs:b-5qwoeo) ([raw file](https://github.com/openmpf/openmpf/blob/977dd0746c7d3d8ad30b7e7850b1cbb967f2e29a/trunk/mpf-rest-api/src/main/java/org/mitre/mpf/rest/api/StreamingJobInfo.java#L31)):*
> FYI, openmpf is using springfox.swagger.version 2.1.1. I haven't been able to find the documentation on that version yet - but maybe the example element isn't supported there either. 

Yeah, I was wondering if the example element was introduced in a recent version.

> To resolve, I'll put example time in the value  element.

I just took a look at Swagger. That works for me.

---

*[trunk/mpf-rest-api/src/main/java/org/mitre/mpf/rest/api/StreamingJobInfo.java, line 47 at r1](https://beta.reviewable.io/reviews/openmpf/openmpf/441#-L4mMrVOD6-LWQUhn6rn:-L4vbhFDnt-jyVR50zEZ:b-ucfkn4) ([raw file](https://github.com/openmpf/openmpf/blob/977dd0746c7d3d8ad30b7e7850b1cbb967f2e29a/trunk/mpf-rest-api/src/main/java/org/mitre/mpf/rest/api/StreamingJobInfo.java#L47)):*

This is legacy commentry, I'll remove it.
</blockquote></details>

Thanks.

---

*[trunk/mpf-rest-api/src/main/java/org/mitre/mpf/rest/api/StreamingJobInfo.java, line 44 at r2](https://beta.reviewable.io/reviews/openmpf/openmpf/441#-L4wbibF962Xuzl11FEt:-L4wbibG9bNF9lTq0WQi:bxi56le) ([raw file](https://github.com/openmpf/openmpf/blob/53ee4181c508bc1230d02e654eb27f62e309ea18/trunk/mpf-rest-api/src/main/java/org/mitre/mpf/rest/api/StreamingJobInfo.java#L44)):*
> ```Java
> 
> 	private Long jobId;
> 	@ApiModelProperty(position=1, required = true)
> ```

I don't think this position actually affects the order on which the properties are listed in Swagger does it? I'm seeing them listed alphabetically. If it's not need, and not doing anything, we should remove the position element. 

Also, it doesn't seem like the required element is reflected on the Swagger page either. Again, if it's not need, and not doing anything, we should remove it.

Maybe I'm missing something though. 

---

*[trunk/mpf-rest-api/src/main/java/org/mitre/mpf/rest/api/StreamingJobInfo.java, line 69 at r2](https://beta.reviewable.io/reviews/openmpf/openmpf/441#-L4vaTPyTWGhyuo-wuUa:-L4vaTPyTWGhyuo-wuUb:b-vy9g3j) ([raw file](https://github.com/openmpf/openmpf/blob/53ee4181c508bc1230d02e654eb27f62e309ea18/trunk/mpf-rest-api/src/main/java/org/mitre/mpf/rest/api/StreamingJobInfo.java#L69)):*
> ```Java
> // 	private float jobProgress;
> //	public float getJobProgress() { return jobProgress; }
> 	private String startDate;
> ```

Can we leave the data members as Date objects and have the getters return a Date? I third party system may use this class to deserialize the StreamingJobInfo JSON. I think it will be more convenient for them to work with the Date object.

Even if it is a Date object, I think it's fine to say `dataType="String"`. However, I would try `dataType="Date"` first though, to see if Swagger complains. If not, let's go with it.

---

*[trunk/mpf-rest-api/src/main/java/org/mitre/mpf/rest/api/StreamingJobInfo.java, line 75 at r2](https://beta.reviewable.io/reviews/openmpf/openmpf/441#-L4wcQyp4-w9pCA6OPmy:-L4wcQyp4-w9pCA6OPmz:b-htv4ik) ([raw file](https://github.com/openmpf/openmpf/blob/53ee4181c508bc1230d02e654eb27f62e309ea18/trunk/mpf-rest-api/src/main/java/org/mitre/mpf/rest/api/StreamingJobInfo.java#L75)):*
> ```Java
>      * matching the TIMESTAMP_PATTERN, which is currently defined as {@value TimeUtils#TIMESTAMP_PATTERN}
>      */
>     @ApiModelProperty(position=2, required = true, dataType="String", value = "streaming job start time, local system time. Example: 2018-01-07 10:23:04.6")
> ```

Please put a period after this example.

---

*[trunk/mpf-rest-api/src/main/java/org/mitre/mpf/rest/api/util/TimeUtils.java, line 1 at r2](https://beta.reviewable.io/reviews/openmpf/openmpf/441#-L4vgXYKkjZ-lFeMfLxq:-L4vgXYLj3EseC-1RP-h:bqur5h) ([raw file](https://github.com/openmpf/openmpf/blob/53ee4181c508bc1230d02e654eb27f62e309ea18/trunk/mpf-rest-api/src/main/java/org/mitre/mpf/rest/api/util/TimeUtils.java#L1)):*
> ```Java
> /******************************************************************************
> ```

This seems to be a copy of org.mitre.mpf.interop.util.TimeUtils. Instead of copying it, make mpf-interop a dependency of the mpf-rest-client in the latter's pom.xml.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/mvc/util/ModelUtils.java, line 148 at r2](https://beta.reviewable.io/reviews/openmpf/openmpf/441#-L4vcuhYRH8qmQVpj5bU:-L4vcuhYRH8qmQVpj5bV:b-g43c1b) ([raw file](https://github.com/openmpf/openmpf/blob/53ee4181c508bc1230d02e654eb27f62e309ea18/trunk/workflow-manager/src/main/java/org/mitre/mpf/mvc/util/ModelUtils.java#L148)):*
> ```Java
>                 streamingJobRequest.getPriority(),
>                 streamingJobRequest.getStatus().name(),
> 				null,
> ```

Pass `streamingJobRequest.getStatusDetail()` here. 

My previous comment about it being null was in reference to how `StreamingJobStatus streamingJobStatus` was not being passed the detail as a parameter to the constructor, so `streamingJobStatus.getDetail()` would always return null.

`streamingJobRequest.getStatusDetail()`may not be null, so we should pass it in.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/CallbackUtils.java, line 138 at r2](https://beta.reviewable.io/reviews/openmpf/openmpf/441#-L4wVXY9edanYZHZX8yu:-L4wVXY9edanYZHZX8yv:b-jc06id) ([raw file](https://github.com/openmpf/openmpf/blob/53ee4181c508bc1230d02e654eb27f62e309ea18/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/CallbackUtils.java#L138)):*
> ```Java
>     // Send the summary report to the URI identified by callbackUri, using the HTTP POST method.
>     public void sendSummaryReportCallback(JsonSegmentSummaryReport summaryReport, String callbackUri) {
>         log.info("Starting POST of summaryReport with jobId " + summaryReport.getJobId());
> ```

Let's update this log line to:

```
log.info("Starting POST of summaryReport to " + callbackUri + " with jobId " + summaryReport.getJobId());
```

---

*[trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/ITWebStreamingHealthReports.java, line 62 at r2](https://beta.reviewable.io/reviews/openmpf/openmpf/441#-L4wW70spxS-UwkAOwxW:-L4wW70spxS-UwkAOwxX:b-sahnn0) ([raw file](https://github.com/openmpf/openmpf/blob/53ee4181c508bc1230d02e654eb27f62e309ea18/trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/ITWebStreamingHealthReports.java#L62)):*
> ```Java
> 
> @FixMethodOrder(MethodSorters.NAME_ASCENDING)
> public class ITWebStreamingHealthReports {
> ```

Thanks for making all of the changes in this file that I asked for. Taking what you've done into consideration, I refactored this class into ITWebStreamingReports in https://github.com/openmpf/openmpf/pull/451. Depending on who lands their code first, one of us will need to merge.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/openmpf/openmpf/441#-:-L4wcxhS5F6im__Kh4KT:b-nqjuyx)*

dpierce17:
*[trunk/mpf-rest-api/src/main/java/org/mitre/mpf/rest/api/StreamingJobInfo.java, line 44 at r2](https://reviewable.io:443/reviews/openmpf/openmpf/441#-L4wbibF962Xuzl11FEt:-L59AkkZ-DqtjVQl44_u:b-20jzgs) ([raw file](https://github.com/openmpf/openmpf/blob/53ee4181c508bc1230d02e654eb27f62e309ea18/trunk/mpf-rest-api/src/main/java/org/mitre/mpf/rest/api/StreamingJobInfo.java#L44)):*

I don't think this position actually affects the order on which the properties are listed in Swagger does it? I'm seeing them listed alphabetically. If it's not need, and not doing anything, we should remove the position element. 

Also, it doesn't seem like the required element is reflected on the Swagger page either. Again, if it's not need, and not doing anything, we should remove it.

Maybe I'm missing something though. 
</blockquote></details>

According to the ApiModelProperty documentation I've seen, the position element allows defining an ordering on the properties when they are listed in the Model. But,  I found this bug report (https://github.com/springfox/springfox/issues/1280) that says that there is a bug in swingfox that makes ApiModelProperty not respect the position assignment. I left @ApiModelProperty position usage in the software in case the springfox swagger position bug is ever fixed. I updated the software by adding a TODO to update to a later version of springfox swagger if that bug has been found to be fixed. Note: at this time, opempf uses springfox.swagger 2.1.1

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/441)*

dpierce17:
*[trunk/mpf-rest-api/src/main/java/org/mitre/mpf/rest/api/StreamingJobInfo.java, line 69 at r2](https://reviewable.io:443/reviews/openmpf/openmpf/441#-L4vaTPyTWGhyuo-wuUa:-L59HENlw6JrDeNmXaiq:b8gf09d) ([raw file](https://github.com/openmpf/openmpf/blob/53ee4181c508bc1230d02e654eb27f62e309ea18/trunk/mpf-rest-api/src/main/java/org/mitre/mpf/rest/api/StreamingJobInfo.java#L69)):*

Can we leave the data members as Date objects and have the getters return a Date? I third party system may use this class to deserialize the StreamingJobInfo JSON. I think it will be more convenient for them to work with the Date object.

Even if it is a Date object, I think it's fine to say `dataType="String"`. However, I would try `dataType="Date"` first though, to see if Swagger complains. If not, let's go with it.
</blockquote></details>

We're are currently specifying in the MPF-REST-API that StreamingJobInfo is returning a formatted date (as a String) for startDate, endDate and activityTimestamp. If we change the getters in this class to return a Date, instead of a String,  aren't we no longer adhering to that API?  

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/441)*

jrobble:
Review status: all files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[trunk/mpf-rest-api/src/main/java/org/mitre/mpf/rest/api/StreamingJobInfo.java, line 44 at r2](https://reviewable.io:443/reviews/openmpf/openmpf/441#-L4wbibF962Xuzl11FEt:-L59_raUVAE3d5VxheTG:bctckfl) ([raw file](https://github.com/openmpf/openmpf/blob/53ee4181c508bc1230d02e654eb27f62e309ea18/trunk/mpf-rest-api/src/main/java/org/mitre/mpf/rest/api/StreamingJobInfo.java#L44)):*

According to the ApiModelProperty documentation I've seen, the position element allows defining an ordering on the properties when they are listed in the Model. But,  I found this bug report (https://github.com/springfox/springfox/issues/1280) that says that there is a bug in swingfox that makes ApiModelProperty not respect the position assignment. I left @ApiModelProperty position usage in the software in case the springfox swagger position bug is ever fixed. I updated the software by adding a TODO to update to a later version of springfox swagger if that bug has been found to be fixed. Note: at this time, opempf uses springfox.swagger 2.1.1
</blockquote></details>

Ok. I'm fine with a TODO.

---

*[trunk/mpf-rest-api/src/main/java/org/mitre/mpf/rest/api/StreamingJobInfo.java, line 69 at r2](https://reviewable.io:443/reviews/openmpf/openmpf/441#-L4vaTPyTWGhyuo-wuUa:-L59XiIM8eKeQ6d3bU76:babm0ht) ([raw file](https://github.com/openmpf/openmpf/blob/53ee4181c508bc1230d02e654eb27f62e309ea18/trunk/mpf-rest-api/src/main/java/org/mitre/mpf/rest/api/StreamingJobInfo.java#L69)):*
The REST API doc describes the JSON inputs and outputs. The date will be a formatted string in the StreamingJobInfo JSON. The StreamingJobInfo class doesn't necessarily need to store the field as a String, as long as when the class is converted to JSON it shows up as a formatted string.

I accomplished this in JsonSegmentSummaryReport [here](https://github.com/openmpf/openmpf/blob/develop/trunk/interop/src/main/java/org/mitre/mpf/interop/JsonSegmentSummaryReport.java#L50) by annotating the "AsString" methods as the `@JsonGetter` and `@JsonSetter`. The `getReportDate` and `setReportDate` methods work with the LocaleDateTime object.

As for changing `dataType="Date"` in Swagger, let's ignore what I said and keep it as a String. When I wrote my previous comment I was thinking that listing it as Date would be a nice clue to the end user, but I just found [this](https://swagger.io/docs/specification/data-models/data-types/): 

> string (this includes dates and files) 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/441)*

jrobble:
Reviewed 6 of 6 files at r3.
Review status: all files reviewed at latest revision, 7 unresolved discussions.

---

*[trunk/interop/src/main/java/org/mitre/mpf/interop/util/TimeUtils.java, line 95 at r3](https://reviewable.io:443/reviews/openmpf/openmpf/441#-L5AiCfxR-YyzN_k4qjy:-L5AiCfyRXGfOv6A0hBm:b-qx5ysf) ([raw file](https://github.com/openmpf/openmpf/blob/1857f9a72a61ce97507a29f7f5331d5222b61554/trunk/interop/src/main/java/org/mitre/mpf/interop/util/TimeUtils.java#L95)):*
> ```Java
>      * using the date/time pattern adhered to by OpenMPF.
>      */
>     public static Date parseStringAsDate(String timestampStr) throws DateTimeParseException, MpfInteropUsageException {
> ```

Please move this parse method above the `getDateAsString` method to be in the same order as the LocalDateTime methods.

---

*[trunk/interop/src/main/java/org/mitre/mpf/interop/util/TimeUtils.java, line 97 at r3](https://reviewable.io:443/reviews/openmpf/openmpf/441#-L5AiTaQD3AdRIMwtTF0:-L5AiTaRLc3VBQ3Sc3j9:bqyj0b) ([raw file](https://github.com/openmpf/openmpf/blob/1857f9a72a61ce97507a29f7f5331d5222b61554/trunk/interop/src/main/java/org/mitre/mpf/interop/util/TimeUtils.java#L97)):*
> ```Java
>     public static Date parseStringAsDate(String timestampStr) throws DateTimeParseException, MpfInteropUsageException {
>         if ( timestampStr == null ) {
>             throw new MpfInteropUsageException("Error, timestampStr may not be null.");
> ```

Use `timestamp String` instead of `timestampStr`.

---

*[trunk/mpf-rest-api/src/main/java/org/mitre/mpf/rest/api/StreamingJobInfo.java, line 43 at r3](https://reviewable.io:443/reviews/openmpf/openmpf/441#-L5AgF3rxSLNlSplBkY3:-L5AgF3sZhvQkMJXRJgt:b-hgs2p) ([raw file](https://github.com/openmpf/openmpf/blob/1857f9a72a61ce97507a29f7f5331d5222b61554/trunk/mpf-rest-api/src/main/java/org/mitre/mpf/rest/api/StreamingJobInfo.java#L43)):*
> ```Java
> public class StreamingJobInfo {
> 
>     // TODO update springfox.swagger version after the @ApiModelProperty postion bug is fixed
> ```

Check spelling: "postion"

---

*[trunk/mpf-rest-api/src/main/java/org/mitre/mpf/rest/api/StreamingJobInfo.java, line 44 at r3](https://reviewable.io:443/reviews/openmpf/openmpf/441#-L5AgKvLtrsA2ndR8GQA:-L5AgKvLtrsA2ndR8GQB:b-exfpv0) ([raw file](https://github.com/openmpf/openmpf/blob/1857f9a72a61ce97507a29f7f5331d5222b61554/trunk/mpf-rest-api/src/main/java/org/mitre/mpf/rest/api/StreamingJobInfo.java#L44)):*
> ```Java
> 
>     // TODO update springfox.swagger version after the @ApiModelProperty postion bug is fixed
> 	// Note: a bug in springfox swagger may be causing the @ApiModelProperty position specification to to not be honored.
> ```

Check grammar: "to to"

---

*[trunk/mpf-rest-api/src/main/java/org/mitre/mpf/rest/api/StreamingJobInfo.java, line 97 at r3](https://reviewable.io:443/reviews/openmpf/openmpf/441#-L5AkJSq3YUI0CMGzvOn:-L5AkJSq3YUI0CMGzvOo:b-9ffh9x) ([raw file](https://github.com/openmpf/openmpf/blob/1857f9a72a61ce97507a29f7f5331d5222b61554/trunk/mpf-rest-api/src/main/java/org/mitre/mpf/rest/api/StreamingJobInfo.java#L97)):*
> ```Java
>     public String getStartDateAsString() {
>         return TimeUtils.getDateAsString(startDate);
>     }
> ```

Let's change the order:

```
private Date endDate;
public Date getEndDate ...
<newline here>
public String getEndDateAsString ...
public void setEndDateFromString ...
````

I prefer getters before setters. Also, that bunches the "String" methods together.

Please do the same for endDate and activity timestamp.

---

*[trunk/mpf-rest-api/src/main/java/org/mitre/mpf/rest/api/StreamingJobInfo.java, line 126 at r3](https://reviewable.io:443/reviews/openmpf/openmpf/441#-L5Ak0yUem_0_ZzZvcki:-L5Ak0yUem_0_ZzZvckj:b9sjmgf) ([raw file](https://github.com/openmpf/openmpf/blob/1857f9a72a61ce97507a29f7f5331d5222b61554/trunk/mpf-rest-api/src/main/java/org/mitre/mpf/rest/api/StreamingJobInfo.java#L126)):*
> ```Java
>     public String getOutputObjectDirectory() {
> 		return outputObjectDirectory;
> 	}
> ```

Add a newline here.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/441)*

**Comment by jrobble** [comment]
Looking at your updates, I realize that we should be using `Date` instead of `LocalDateTime` in all of the JSON classes. I created these issues: https://github.com/openmpf/openmpf/issues/459, https://github.com/openmpf/openmpf/issues/460.

---

Review status: all files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/441#-:-L5AqCczxDVLduvuiIhy:bmja28t)*

**Comment by jrobble** [comment]
Review status: all files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[trunk/mpf-rest-api/src/main/java/org/mitre/mpf/rest/api/StreamingJobInfo.java, line 119 at r3](https://reviewable.io:443/reviews/openmpf/openmpf/441#-L5BrlnfKIskqdvT4sVt:-L5BrlngjDCDmH7om5bY:b-hgm0cc) ([raw file](https://github.com/openmpf/openmpf/blob/1857f9a72a61ce97507a29f7f5331d5222b61554/trunk/mpf-rest-api/src/main/java/org/mitre/mpf/rest/api/StreamingJobInfo.java#L119)):*
> ```Java
>      */
>     public String getEndDateAsString() {
>         return TimeUtils.getDateAsString(endDate);
> ```

I got the following in the StreamingJobInfo response:

```
{
  "jobId": 193,
  "pipelineName": "XYZ FACE DETECTION PIPELINE",
  "jobStatus": "IN_PROGRESS",
  "jobStatusDetail": null,
  "startDate": "2018-02-12 20:45:29.0",
  "endDate": "",
  "outputObjectDirectory": "/home/mpf/git/openmpf-projects/openmpf/trunk/install/share/output-objects/193",
  "streamUri": "rtsp://localhost:5554/test.mpeg4",
  "activityFrameId": null,
  "activityTimestamp": null,
  "terminal": false
}
```

Notice that `endDate` is "". It should be null. This is because of how TimeUtils.getDateAsString() is implemented. Please update that method to return a null.

I think TimeUtils.getLocateDateTimeAsString() can be updated in the same way. Once we do that, I think we can simplify the implementations of RedisImpl.getActivityTimestampAsString(), and StreamingJobInfo.getActivityTimestampAsString(), which both do a null check. They shouldn't have to do it anymore.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/441)*

jrobble:
Ran code. Works as expected.

---

Reviewed 2 of 2 files at r4.
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[trunk/interop/src/main/java/org/mitre/mpf/interop/util/TimeUtils.java, line 66 at r4](https://reviewable.io:443/reviews/openmpf/openmpf/441#-L5FUHsEe7W7TOhndgTA:-L5FUHsEe7W7TOhndgTB:byaup5y) ([raw file](https://github.com/openmpf/openmpf/blob/bde91799b1193af341809c5d2c8f9141b3262d41/trunk/interop/src/main/java/org/mitre/mpf/interop/util/TimeUtils.java#L66)):*
Now that this is updated:

> we can simplify the implementations of RedisImpl.getActivityTimestampAsString(), and StreamingJobInfo.getActivityTimestampAsString(), which both do a null check. They shouldn't have to do it anymore.

You updated StreamingJobInfo, but still need to update RedisImpl.

---

*[trunk/mpf-rest-api/src/main/java/org/mitre/mpf/rest/api/StreamingJobInfo.java, line 111 at r4](https://reviewable.io:443/reviews/openmpf/openmpf/441#-L5FSJKLkoIeLFZr1pEF:-L5FSJKLkoIeLFZr1pEG:b-19mfvq) ([raw file](https://github.com/openmpf/openmpf/blob/bde91799b1193af341809c5d2c8f9141b3262d41/trunk/mpf-rest-api/src/main/java/org/mitre/mpf/rest/api/StreamingJobInfo.java#L111)):*
> ```Java
>     }
> 
>     @ApiModelProperty(position=3, required = true, dataType="String", value = "streaming job end time, local system time. Example: 2018-01-08 00:00:00.0 or empty String if the job hasn't completed.")
> ```

The "or empty String part here" is incorrect since it can now be null.

---

*[trunk/mpf-rest-api/src/main/java/org/mitre/mpf/rest/api/StreamingJobInfo.java, line 150 at r4](https://reviewable.io:443/reviews/openmpf/openmpf/441#-L5FSztu_JsYqls7Kldm:-L5FSztvcR1vtm0BdcuR:b-dsea6o) ([raw file](https://github.com/openmpf/openmpf/blob/bde91799b1193af341809c5d2c8f9141b3262d41/trunk/mpf-rest-api/src/main/java/org/mitre/mpf/rest/api/StreamingJobInfo.java#L150)):*
> ```Java
>     public Date getActivityTimestamp() { return activityTimestamp; }
> 
>     @ApiModelProperty(position=10, required = true, dataType="String", value = "detection time associated with the activityFrameId, local system time. Example: 2018-01-07 18:30:00.5 or empty String if there has been no activity found in the job.")
> ```

The "or empty String part here" is incorrect since it can now be null.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/441#-:-L5FUrXLwX1_DaPCt_Xg:burlpuz)*

jrobble:
Reviewed 3 of 3 files at r5.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/441)*

PR #448, opened 2018-02-08
Updated version numbers to 2.0.0.

jrobble:
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/448)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -30,7 +30,7 @@
   <groupId>org.mitre.mpf</groupId>
   <artifactId>mpf-project</artifactId>
   <packaging>pom</packaging>
-  <version>1.0.0</version>
+  <version>2.0.0</version>
   <name>Media Processing Framework</name>
 
     <!-- poms in the following subdirectories -->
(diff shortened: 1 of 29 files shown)


jrobble:
This PR also addresses an issue where `mvn clean` results in trying to download jar dependencies for OpenMPF submodules that haven't been built yet, resulting in a failure to clean the workflow manager submodule. 

Specifically, the issue is with using the antrun plugin (or any non-default plugin for that matter) during the clean phase. That causes Maven to resolve all of its dependences during that phase. If that plugin is not used during that phase, Maven will not resolve its dependencies and the workflow manager submodule will be successfully cleaned. More info can be found [here](https://stackoverflow.com/a/11950440).

To address this, I removed the usage of the maven-antrun-plugin plugin during the clean phase and am now using the maven-clean-plugin.

brosenberg42:
Reviewed 29 of 29 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/448)*

Issue #502, opened 2018-03-01
WFM configures services before JGroups cluster membership is determined


jrobble:
Due to a race condition, the WFM sometimes configures services before JGroups cluster membership is determined. Specifically, the WFM does not wait until `viewAccepted()` is invoked by JGroups. This results in this kind of output:

```
-------------------------------------------------------------------
GMS: address=MasterNode:localhost.localdomain:MPF-MasterNode, cluster=MPF_Channel, physical address=10.0.2.15:7801
-------------------------------------------------------------------
2018-02-28 16:22:31,276 INFO [main] org.mitre.mpf.nms.MasterNode - Loading node manager config
2018-02-28 16:22:31,314 INFO [main] org.mitre.mpf.nms.MasterNode - Node descriptor created for expected (not yet discovered) NodeManager: localhost.localdomain
2018-02-28 16:22:34,388 WARN [main] o.m.mpf.nms.MasterNodeStateManager - The NodeManager at hostname 'localhost.localdomain' is not running and is in state 'Configured'.
2018-02-28 16:22:34,389 WARN [main] o.m.mpf.nms.MasterNodeStateManager - The service with name 'localhost.localdomain:FrameDataComponent:1' cannot be launched.
```

jrobble:
This issue was discovered by running the node-manager in one window:

```
/usr/java/latest/bin/java -Djgroups.tcp.address=localhost.localdomain -Djgroups.tcp.port=7800 -Djgroups.tcpping.initial_hosts=localhost.localdomain[7800] -jar /home/mpf/git/openmpf-projects/openmpf/trunk/install/jars/mpf-nodemanager-2.0.0.jar
``` 

And running TestSystemOnDiff in another:

```
mvn surefire:test -Dcomponents.build.components=openmpf-components:openmpf-contrib-componts:/home/mpf/git/mpf_pty_comps -Dtransport.guarantee="NONE" -Dweb.rest.protocol="http" -Dstartup.auto.registration.skip=false -DfailIfNoTests=false -Dtest=TestSystemOnDiff -DskipITs -Dit.test=none -Dmaven.tomcat.skip=true
```

The "Node descriptor created for expected (not yet discovered) NodeManager: localhost.localdomain" output would consistently be generated on my local VM unless I set break points in the WFM in attempt to debug the issue.

PR #503, opened 2018-03-02
Fix JGroups race condition.

jrobble:
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/503)


jrobble:
This addresses #502. Note that jGroupsConfig.xml was never used. jGroupsTcpConfig.xml was used instead. Added some synchronization blocks to address a concurrent modification exception.

brosenberg42:
Reviewed 6 of 7 files at r1.
Review status: 6 of 7 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[trunk/node-manager/src/main/java/org/mitre/mpf/nms/ChannelReceiver.java, line 128 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/503#-L6bOGWNvJuLye3M58dw:-L6bOGWNvJuLye3M58dx:b-2ce99m) ([raw file](https://github.com/openmpf/openmpf/blob/cb1f9f0acb7cf92a8e10b0a591474c5c65d4217b/trunk/node-manager/src/main/java/org/mitre/mpf/nms/ChannelReceiver.java#L128)):*
> ```Java
>     }
> 
>     public void handleView(View view, boolean forced) {
> ```

This should be private

---

*[trunk/node-manager/src/main/java/org/mitre/mpf/nms/ChildNodeStateManager.java, line 159 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/503#-L6bPJ7nXjiNJDSVCnCn:-L6bPJ7nXjiNJDSVCnCo:b-fh35y6) ([raw file](https://github.com/openmpf/openmpf/blob/cb1f9f0acb7cf92a8e10b0a591474c5c65d4217b/trunk/node-manager/src/main/java/org/mitre/mpf/nms/ChildNodeStateManager.java#L159)):*
> ```Java
>         // lookup node by name, remove from table, perform other actions (or let object do its own cleanup)
>         BaseServiceLauncher theApp = null;
>         synchronized (launchedAppsMap) {
> ```

Since `launchedAppsMap.get(desc.getName());` and `launchedAppsMap.remove(desc.getName());` are in separate synchronized blocks, multiple threads can try to shutdown the same service.
This can be fixed by either putting the whole method inside this synchronized block, or changing this synchronized block to be: `theApp = launchedAppsMap.remove(desc.getName());`.

---

*[trunk/node-manager/src/main/java/org/mitre/mpf/nms/ChildNodeStateManager.java, line 188 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/503#-L6bVtoTXA5nV-8MqolH:-L6bVtoTXA5nV-8MqolI:b7rblo1) ([raw file](https://github.com/openmpf/openmpf/blob/cb1f9f0acb7cf92a8e10b0a591474c5c65d4217b/trunk/node-manager/src/main/java/org/mitre/mpf/nms/ChildNodeStateManager.java#L188)):*
> ```Java
>         // create and launch a node of the specified type with the given name
>         BaseServiceLauncher launcher = null;
>         synchronized (launchedAppsMap) {
> ```

This has a similar problem to the one I pointed out in the shutdown method. I will call this synchronized block B1 and the block on line 196 B2.
There is an issue if the following series of events occurs:
1. Thread 1 executes B1
2. Thread 2 executes B1
3. Thread 1 executes B2
4. Thread 2 executes B2

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/nodeManager/NodeManagerStatus.java, line 117 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/503#-L6bUU9hYcBxlPOmDt0O:-L6bUU9hYcBxlPOmDt0P:b-1kf7yx) ([raw file](https://github.com/openmpf/openmpf/blob/cb1f9f0acb7cf92a8e10b0a591474c5c65d4217b/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/nodeManager/NodeManagerStatus.java#L117)):*
> ```Java
>             }
>         } catch (InterruptedException e) {
>             log.warn("Interruped while waiting for cluster view update.");
> ```

Since you are not re-throwing the InterruptedException, you should call `Thread.currentThread().interrupt();` in order to preserve the fact that the thread was interrupted. Also, you misspelled interrupted in the log message.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/503)*

jrobble:
Review status: 2 of 8 files reviewed at latest revision, 4 unresolved discussions.

---

*[trunk/node-manager/src/main/java/org/mitre/mpf/nms/ChannelReceiver.java, line 128 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/503#-L6bOGWNvJuLye3M58dw:-L6crG8Uk8P-iKlA2HHF:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/cb1f9f0acb7cf92a8e10b0a591474c5c65d4217b/trunk/node-manager/src/main/java/org/mitre/mpf/nms/ChannelReceiver.java#L128)):*

This should be private
</blockquote></details>

Done.

---

*[trunk/node-manager/src/main/java/org/mitre/mpf/nms/ChildNodeStateManager.java, line 159 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/503#-L6bPJ7nXjiNJDSVCnCn:-L6cxHB34Q9DQH8HvcpQ:b3ia7wp) ([raw file](https://github.com/openmpf/openmpf/blob/cb1f9f0acb7cf92a8e10b0a591474c5c65d4217b/trunk/node-manager/src/main/java/org/mitre/mpf/nms/ChildNodeStateManager.java#L159)):*

Since `launchedAppsMap.get(desc.getName());` and `launchedAppsMap.remove(desc.getName());` are in separate synchronized blocks, multiple threads can try to shutdown the same service.
This can be fixed by either putting the whole method inside this synchronized block, or changing this synchronized block to be: `theApp = launchedAppsMap.remove(desc.getName());`.
</blockquote></details>

I was trying to avoid the situation where this method is blocking the while loop in `run`. `theApp.shutdown();` can take some time, so I didn't want the sync block to contain that and block other services from shutting down. I'll go with your second suggestion.

---

*[trunk/node-manager/src/main/java/org/mitre/mpf/nms/ChildNodeStateManager.java, line 188 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/503#-L6bVtoTXA5nV-8MqolH:-L6eI_3jODGk3oCP9Qz7:b-e6k4jj) ([raw file](https://github.com/openmpf/openmpf/blob/cb1f9f0acb7cf92a8e10b0a591474c5c65d4217b/trunk/node-manager/src/main/java/org/mitre/mpf/nms/ChildNodeStateManager.java#L188)):*

This has a similar problem to the one I pointed out in the shutdown method. I will call this synchronized block B1 and the block on line 196 B2.
There is an issue if the following series of events occurs:
1. Thread 1 executes B1
2. Thread 2 executes B1
3. Thread 1 executes B2
4. Thread 2 executes B2
</blockquote></details>

Yeah, there's a similar situation here where `launcher.startup` can take some time and block other services from launching ... or so I thought. Looking more closely, that code spawns a separate thread for the launch, so that's not really a concern.

Still though, I don't like the idea of capturing the sleeps used in this method in the sync block. The "things to propagate" might have something to do with the while loop in `run`. I don't want to block that by holding onto launchedAppsMap during one of the sleeps.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/nodeManager/NodeManagerStatus.java, line 117 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/503#-L6bUU9hYcBxlPOmDt0O:-L6crQVJza5a7dzCSfNV:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/cb1f9f0acb7cf92a8e10b0a591474c5c65d4217b/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/nodeManager/NodeManagerStatus.java#L117)):*

Since you are not re-throwing the InterruptedException, you should call `Thread.currentThread().interrupt();` in order to preserve the fact that the thread was interrupted. Also, you misspelled interrupted in the log message.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/503)*

brosenberg42:
Reviewed 5 of 7 files at r2.
Review status: 7 of 8 files reviewed at latest revision, 1 unresolved discussion.

---

*[trunk/node-manager/src/main/java/org/mitre/mpf/nms/ChildNodeStateManager.java, line 188 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/503#-L6bVtoTXA5nV-8MqolH:-L6qSbEjZ-RiwMKnJz4Q:b-l7cmyv) ([raw file](https://github.com/openmpf/openmpf/blob/cb1f9f0acb7cf92a8e10b0a591474c5c65d4217b/trunk/node-manager/src/main/java/org/mitre/mpf/nms/ChildNodeStateManager.java#L188)):*

Yeah, there's a similar situation here where `launcher.startup` can take some time and block other services from launching ... or so I thought. Looking more closely, that code spawns a separate thread for the launch, so that's not really a concern.

Still though, I don't like the idea of capturing the sleeps used in this method in the sync block. The "things to propagate" might have something to do with the while loop in `run`. I don't want to block that by holding onto launchedAppsMap during one of the sleeps.
</blockquote></details>

This changes the logic from the original. In the original, nothing happens when the service is already in `launchedAppsMap`. In your code, when the service is already in `launchedAppsMap`, you set the state of the service to `NodeManagerConstants.States.Inactive`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/503)*

jrobble:
Review status: 7 of 8 files reviewed at latest revision, 1 unresolved discussion.

---

*[trunk/node-manager/src/main/java/org/mitre/mpf/nms/ChildNodeStateManager.java, line 188 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/503#-L6bVtoTXA5nV-8MqolH:-L6qm_xH0WxNfScv3Rdq:b6zrrkh) ([raw file](https://github.com/openmpf/openmpf/blob/cb1f9f0acb7cf92a8e10b0a591474c5c65d4217b/trunk/node-manager/src/main/java/org/mitre/mpf/nms/ChildNodeStateManager.java#L188)):*

This changes the logic from the original. In the original, nothing happens when the service is already in `launchedAppsMap`. In your code, when the service is already in `launchedAppsMap`, you set the state of the service to `NodeManagerConstants.States.Inactive`.
</blockquote></details>

Good catch.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/503)*

brosenberg42:
:lgtm:

---

Reviewed 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/503#-:-L6vr7F-D27109_8c9ID:bnfp4nl)*

PR #506, opened 2018-03-12
Add frame rate cap with 25 unit tests for FRAME_INTERVAL, FRAME_CAP_RATE combos

dpierce17:
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/506)


jrobble:
Reviewed 10 of 11 files at r1.
Review status: all files reviewed at latest revision, 14 unresolved discussions.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/DetectionSplitter.java, line 179 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/506#-L7QJuQktqJFs5i3DHuP:-L7QJuQktqJFs5i3DHuQ:b-w5uhso) ([raw file](https://github.com/openmpf/openmpf/blob/d325af8959eb23d3b674ff59efee9cfa09bb7152/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/DetectionSplitter.java#L179)):*
> ```Java
>             // FRAME_RATE_CAP override is allowed at 5 levels: system properties (lowest priority), action properties, job properties, algorithm properties or media properties (highest priority).
> 
>             Double mediaFPS = null;
> ```

Let's use primitives (e.g. `double`) instead of objects where possible. 

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/DetectionSplitter.java, line 309 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/506#-L7QSvlf5ZNQkSPeDgn1:-L7QSvlf5ZNQkSPeDgn2:b-jnhwig) ([raw file](https://github.com/openmpf/openmpf/blob/d325af8959eb23d3b674ff59efee9cfa09bb7152/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/DetectionSplitter.java#L309)):*
> ```Java
>                 // Also persist the update to REDIS so that the change will be in the json output object.
>                 if ( computedFrameInterval != null ){
>                     // For videos, phasing out FRAME_INTERVAL property in favor of COMPUTED_FRAME_INTERVAL_PROPERTY property sent to each sub-job.
> ```

CCing @brosenberg42.

I noticed that some of our example components look for FRAME_INTERVAL in the job properties, and I'm thinking a bit more about things from the perspective of a third party component developer.

As a component developer, I specified support for "FRAME_INTERVAL" in my descriptor.json file, so I may be looking for that property in my code. I may not be using MPFVideoCapture and instead rolling my own frame skipping solution. I will not be looking for "COMPUTED_FRAME_INTERVAL" in my code.

That said, let's not remove "FRAME_INTERVAL" from the set of properties passed down to the component. It will be the component's responsibility to ignore them and let MPFVideoCapture do the right thing, or handle both the passed FRAME_INTERVAL and FRAME_RATE_CAP.

After running a test job, and looking the the JSON output object, I don't see COMPUTED_FRAME_INTERVAL anywhere. I thought you mentioned that you did something to make it show up? Am I missing something? I see that COMPUTED_FRAME_INTERVAL is sent in the ActiveMQ message(s).

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/DetectionSplitter.java, line 308 at r2](https://reviewable.io:443/reviews/openmpf/openmpf/506#-L7_N5Wpd_2MMlTUj2Uq:-L7_N5Wpd_2MMlTUj2Ur:b-hxvixy) ([raw file](https://github.com/openmpf/openmpf/blob/046ff1ccf529ba5586899d265789a5d092ffef3d/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/DetectionSplitter.java#L308)):*
> ```Java
>                 // and store the computed frame interval in the TransientJob algorithm properties.
>                 // Also persist the update to REDIS so that the change will be in the json output object.
>                 if ( computedFrameInterval != null ){
> ```

Could we just determine the `computedFrameInterval` once here, using the final values of frame rate cap and frame interval, rather than 4 times? I assume that the final values (after all of the necessary overrides) will be in `modifiedMap`.

I see a lot of duplicated code above. I don't think `getComputedFrameIntervalForVideo` needs to be called 4 times.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/DetectionSplitter.java, line 494 at r2](https://reviewable.io:443/reviews/openmpf/openmpf/506#-L7_LdxVkK2k8bpmNbep:-L7_LdxVkK2k8bpmNbeq:b-8zhuui) ([raw file](https://github.com/openmpf/openmpf/blob/046ff1ccf529ba5586899d265789a5d092ffef3d/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/DetectionSplitter.java#L494)):*
> ```Java
>      * @return Computed frame interval, may be null.
>      */
>     private Double getComputedFrameIntervalForVideo( Double mediaFPS, Map<String,String> propertyOverrideMap, Double lastComputedFrameInterval) {
> ```

Use `double` over `Double` in this method signature and wherever possible. 

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/mediainspection/MediaInspectionProcessor.java, line 94 at r2](https://reviewable.io:443/reviews/openmpf/openmpf/506#-L7_NYayV5Ef5gilXSDH:-L7_NYayV5Ef5gilXSDI:b-tlkqje) ([raw file](https://github.com/openmpf/openmpf/blob/046ff1ccf529ba5586899d265789a5d092ffef3d/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/mediainspection/MediaInspectionProcessor.java#L94)):*
> ```Java
> 				} catch(IOException ioe) {
> 					transientMedia.setFailed(true);
> 					transientMedia.setMessage("Could not calculate the SHA-256 hash for the file due to an exception.");
> ```

As mentioned in previous PRs, please declare a string so that you only need to hardcode the error string once. Then use that message to log the error and set the message. Please do this for the three instances below that you recently added.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/transients/TransientJob.java, line 63 at r2](https://reviewable.io:443/reviews/openmpf/openmpf/506#-L7_OFxpCQRTGpxNy0Yb:-L7_OFxqDYCGa_zo2f9t:bbx8op9) ([raw file](https://github.com/openmpf/openmpf/blob/046ff1ccf529ba5586899d265789a5d092ffef3d/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/transients/TransientJob.java#L63)):*
> ```Java
> 	public Map<String, Map> getOverriddenAlgorithmProperties() { return overriddenAlgorithmProperties; }
> 	public void setOverriddenAlgorithmProperties(Map<String, Map> overriddenAlgorithmProperties) { this.overriddenAlgorithmProperties = overriddenAlgorithmProperties; }
>     public void addOverriddenAlgorithmProperty(String algorithmName, String algorithmPropertyKey, String algorithmPropertyValue) {
> ```

This method is not used. Please remove it.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/transients/TransientJob.java, line 69 at r2](https://reviewable.io:443/reviews/openmpf/openmpf/506#-L7_OHmadPMsRaN5Bk_R:-L7_OHmadPMsRaN5Bk_S:bbx8op9) ([raw file](https://github.com/openmpf/openmpf/blob/046ff1ccf529ba5586899d265789a5d092ffef3d/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/transients/TransientJob.java#L69)):*
> ```Java
>         }
> 	}
> 	public void renameOverriddenAlgorithmProperty(String algorithmName, String oldKey, String newKey) {
> ```

This method is not used. Please remove it.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/transients/TransientJob.java, line 86 at r2](https://reviewable.io:443/reviews/openmpf/openmpf/506#-L7_PNpBdjbMQ2Y8qax_:-L7_PNpBdjbMQ2Y8qaxa:bxyr03j) ([raw file](https://github.com/openmpf/openmpf/blob/046ff1ccf529ba5586899d265789a5d092ffef3d/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/transients/TransientJob.java#L86)):*
> ```Java
> 	private Map<String, String> overriddenJobProperties;
> 	public Map<String, String> getOverriddenJobProperties() { return overriddenJobProperties; }
> 	public void setOverriddenJobProperties(Map<String, String> overriddenJobProperties) { this.overriddenJobProperties = overriddenJobProperties; }
> ```

The following methods are not used, so please remove them:

```
public void setOverriddenJobProperties(Map<String, String> overriddenJobProperties) { this.overriddenJobProperties = overriddenJobProperties; }
public void addOverriddenJobProperty(String key, String value) { overriddenJobProperties.put(key,value); }
public void renameOverriddenJobProperty(String oldKey, String newKey) {
	    String value = overriddenJobProperties.remove(oldKey);
        overriddenJobProperties.put(newKey, value);
}
public void replaceOverriddenJobProperty(String oldKey, String newKey, String newValue) {
        overriddenJobProperties.remove(oldKey);
        overriddenJobProperties.put(newKey, newValue);
}
```

---

*[trunk/workflow-manager/src/main/resources/properties/mpf.properties, line 92 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/506#-L7QUakbKes4YMcHhkiV:-L7QUakbKes4YMcHhkiW:b-azdmtd) ([raw file](https://github.com/openmpf/openmpf/blob/d325af8959eb23d3b674ff59efee9cfa09bb7152/trunk/workflow-manager/src/main/resources/properties/mpf.properties#L92)):*
Add:

> May be disabled by setting it <= 0.

---

*[trunk/workflow-manager/src/main/resources/properties/mpf.properties, line 99 at r1](https://reviewable.io:443/reviews/openmpf/openmpf/506#-L7QUmuXIxt0JKpzwloj:-L7QUmuXIxt0JKpzwlok:b-vxv2b3) ([raw file](https://github.com/openmpf/openmpf/blob/d325af8959eb23d3b674ff59efee9cfa09bb7152/trunk/workflow-manager/src/main/resources/properties/mpf.properties#L99)):*
Update this to:

> The DEFAULT frame rate cap system property. FRAME_RATE_CAP property can be used to set a threshold on the maximum number of frames to process within one second of the native video time. This may be overridden in a detection action by providing the FRAME_RATE_CAP property. May be disabled by setting it <= 0.

---

*[trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/camelOps/TestDetectionSplitter.java, line 395 at r2](https://reviewable.io:443/reviews/openmpf/openmpf/506#-L7_QIjS6ypUe0mu8Igs:-L7_QIjS6ypUe0mu8Igt:b-dk8cmn) ([raw file](https://github.com/openmpf/openmpf/blob/046ff1ccf529ba5586899d265789a5d092ffef3d/trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/camelOps/TestDetectionSplitter.java#L395)):*
> ```Java
> 
>     // The following set of DetectionSplitter tests specifically test FRAME_RATE_CAP vs. FRAME_INTERVAL property
>     // overrides at various category levels (system, action, job, algorithm, and media (with media properties being the highest ranking).
> ```

Missing end ) in this comment.

---

*[trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/camelOps/TestDetectionSplitter.java, line 445 at r2](https://reviewable.io:443/reviews/openmpf/openmpf/506#-L7_n8-ahuTe4-VPdr8C:-L7_n8-ahuTe4-VPdr8D:byiw5ks) ([raw file](https://github.com/openmpf/openmpf/blob/046ff1ccf529ba5586899d265789a5d092ffef3d/trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/camelOps/TestDetectionSplitter.java#L445)):*
> ```Java
>     }
> 
>     private DetectionProtobuf.DetectionRequest createFrameRateTestTransientJobAndPerformDetectionSplit(long jobId, String externalId,
> ```

Consider beefing up this method to make the rest of your code less verbose:

- Instead of passing in collections, for each collection instead pass in two Integers: one that's the value of the frame interval for that property collection, and one that's the value of the frame rate cap. Using the Integer object allows you to specify `null` to represent that a value isn't specified. You can then create the collections within this method.

- Building on the previous bullet, this method can also take in parameters for the mocked values of the system properties. (See other comment on how to use Mockito.)

- Pass in the expected computed frame interval and perform the check in this method.

- Have this method determine the job id and clear redis.

- Basically, make this method do as much as possible. It would be great if we can simplify the tests to the point where each test case simply calls this method. Please take a look at what Brian did with the gtests [here](https://github.com/openmpf/openmpf-cpp-component-sdk/blob/edae6ab322792a23617264f186385fd8f066e503/detection/api/test/test_detection_api.cpp#L68).

---

*[trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/camelOps/TestDetectionSplitter.java, line 502 at r2](https://reviewable.io:443/reviews/openmpf/openmpf/506#-L7_S32T_gt04JDz0bCj:-L7_S32T_gt04JDz0bCk:bcb3mp7) ([raw file](https://github.com/openmpf/openmpf/blob/046ff1ccf529ba5586899d265789a5d092ffef3d/trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/camelOps/TestDetectionSplitter.java#L502)):*
> ```Java
>                  propertiesUtil.getFrameRateCap(), propertiesUtil.getSamplingInterval());
> 
>         if ( propertiesUtil.getFrameRateCap() < 0 ) {
> ```

I see what you're doing here and why; however, I think that a unit test should always test for the same thing and not depend on external factors. Thus, I recommend we always test for each case.

We should use Mockito to mock the return values of `PropertiesUtil.getSamplingInterval()` and `PropertiesUtil.getFrameRateCap()` so that `DetectionSplitter.performSplit()` uses the values we want it to for each test. Refer to [this code](https://github.com/openmpf/openmpf/blob/0992ddaca607ccddad499f8db8c3420279ba0b89/trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/service/component/TestStartupComponentRegistrationService.java#L88) for an example of how to do that:

```
_mockPropertiesUtil = mock(PropertiesUtil.class);
		when(_mockPropertiesUtil.getUploadedComponentsDirectory())
				.thenReturn(_componentUploadDir.getRoot());
```

Note that you will probably need to do `@InjectMocks` on `private StageSplitter detectionStageSplitter` in order to inject the mocked `PropertiesUtil`.

Note that you're only testing for 2 out of 4 possible combinations. Each system property can have one of two possible states: { > 0 (specified), < 0 (specified disabled) }. Since we have two properties, that's 2 x 2 = 4 combinations.

---

*[trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/camelOps/TestDetectionSplitter.java, line 619 at r2](https://reviewable.io:443/reviews/openmpf/openmpf/506#-L7_W89CQJASAdVD9OLo:-L7_W89CQJASAdVD9OLp:b-qqt978) ([raw file](https://github.com/openmpf/openmpf/blob/046ff1ccf529ba5586899d265789a5d092ffef3d/trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/camelOps/TestDetectionSplitter.java#L619)):*
> ```Java
> 
>         // Run actionProperty test case #5: both FRAME_RATE_CAP and FRAME_INTERVAL are disabled.
>         // In this case setting should fall back to system property values.
> ```

We should only fall back to a system property value if the property is not specified in any other way (job, algorithm, media, pipeline). In this case, the "-1" settings override the system property settings, so the behavior should be the same as if the system property settings were both "-1".

You have multiple test cases where you assume that setting "-1" at a higher level should result in using the value at one lower level. That is incorrect. If the value is specified at a higher level, then that's the value we use. We only fallback when the value is not specified at all.

For example, in the media properties, if frame interval is set to "-1" , and frame rate cap is not set, then we start falling back to determine the frame rate cap property value, starting with the algorithm property. Frame interval is already set to "-1" so we don't need to look back any further to determine its value. If we eventually get to the system property for frame rate cap, and it's "-1", then we use a computed frame interval of 1.

Currently, you have 5 test methods, one for each of the property levels (system, action (pipeline), job, algorithm, and media). Each test method has 5 test cases:

- test case # 1: just FRAME_INTERVAL property is specified and not disabled
- test case # 2: FRAME_RATE_CAP and FRAME_INTERVAL are both specified and not disabled
- test case # 3: FRAME_RATE_CAP is disabled and FRAME_INTERVAL is specified and not disabled
- test case # 4: FRAME_RATE_CAP is specified and not disabled and FRAME_INTERVAL disabled
- test case # 5: both FRAME_RATE_CAP and FRAME_INTERVAL are disabled

If we think about all of the possible combinations, each property can have one of three possible states: { > 0 (specified), < 0 (specified disabled), not specified }, except for the system properties, which we can assume will always be specified, so those can have only two possible states. I admit, I did not think of all of these states when I wrote the task.

Since we have 2 properties, that is a total of 2 x 2 = 4 combinations at the system property level, and 3 x 3 =  9 combinations at each of the other 4 levels. Thus, we have a total of 4 * 9 * 9 * 9 * 9 = 26,244 possible combinations. Of course, we're not going to test for all of those.

Minimally, I think we should test all of the combinations at each level, so that's 4 + 9 + 9 + 9 + 9 = 40 test cases. When we need to fallback to get a value in the case where one or both properties aren't specified at the current level, then we should write the test so that it only needs to fallback to the next lowest level, where the value is "> 0 (specified)". We already have other tests that test for falling back to even lower levels.

In light of this, you may be able to merge the "check" method code into the respective "test" method code. If you can, that will help simplify things.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/506)*

dpierce17:
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/mediainspection/MediaInspectionProcessor.java, line 94 at r2](https://reviewable.io:443/reviews/openmpf/openmpf/506#-L7_NYayV5Ef5gilXSDH:-L7jI6iqLLrONiysoHiF:b-jfa4bs) ([raw file](https://github.com/openmpf/openmpf/blob/046ff1ccf529ba5586899d265789a5d092ffef3d/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/mediainspection/MediaInspectionProcessor.java#L94)):*

As mentioned in previous PRs, please declare a string so that you only need to hardcode the error string once. Then use that message to log the error and set the message. Please do this for the three instances below that you recently added.
</blockquote></details>

I only see 2 instances where there is a duplicated error string, so I'm fixing the two instances. FYI, this is legacy code (I didn't add it).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/506)*

dpierce17:
*[trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/camelOps/TestDetectionSplitter.java, line 619 at r2](https://reviewable.io:443/reviews/openmpf/openmpf/506#-L7_W89CQJASAdVD9OLo:-L7jnoiCCexPV5fCcw6u:bg1dxjl) ([raw file](https://github.com/openmpf/openmpf/blob/046ff1ccf529ba5586899d265789a5d092ffef3d/trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/camelOps/TestDetectionSplitter.java#L619)):*

We should only fall back to a system property value if the property is not specified in any other way (job, algorithm, media, pipeline). In this case, the "-1" settings override the system property settings, so the behavior should be the same as if the system property settings were both "-1".

You have multiple test cases where you assume that setting "-1" at a higher level should result in using the value at one lower level. That is incorrect. If the value is specified at a higher level, then that's the value we use. We only fallback when the value is not specified at all.

For example, in the media properties, if frame interval is set to "-1" , and frame rate cap is not set, then we start falling back to determine the frame rate cap property value, starting with the algorithm property. Frame interval is already set to "-1" so we don't need to look back any further to determine its value. If we eventually get to the system property for frame rate cap, and it's "-1", then we use a computed frame interval of 1.

Currently, you have 5 test methods, one for each of the property levels (system, action (pipeline), job, algorithm, and media). Each test method has 5 test cases:

- test case # 1: just FRAME_INTERVAL property is specified and not disabled
- test case # 2: FRAME_RATE_CAP and FRAME_INTERVAL are both specified and not disabled
- test case # 3: FRAME_RATE_CAP is disabled and FRAME_INTERVAL is specified and not disabled
- test case # 4: FRAME_RATE_CAP is specified and not disabled and FRAME_INTERVAL disabled
- test case # 5: both FRAME_RATE_CAP and FRAME_INTERVAL are disabled

If we think about all of the possible combinations, each property can have one of three possible states: { > 0 (specified), < 0 (specified disabled), not specified }, except for the system properties, which we can assume will always be specified, so those can have only two possible states. I admit, I did not think of all of these states when I wrote the task.

Since we have 2 properties, that is a total of 2 x 2 = 4 combinations at the system property level, and 3 x 3 =  9 combinations at each of the other 4 levels. Thus, we have a total of 4 * 9 * 9 * 9 * 9 = 26,244 possible combinations. Of course, we're not going to test for all of those.

Minimally, I think we should test all of the combinations at each level, so that's 4 + 9 + 9 + 9 + 9 = 40 test cases. When we need to fallback to get a value in the case where one or both properties aren't specified at the current level, then we should write the test so that it only needs to fallback to the next lowest level, where the value is "> 0 (specified)". We already have other tests that test for falling back to even lower levels.

In light of this, you may be able to merge the "check" method code into the respective "test" method code. If you can, that will help simplify things.


</blockquote></details>

Isn't there an exception to this strategy? Previously, we said that if FRAME_RATE_CAP is not set, then fallback to FRAME_INTERVAL. In the above statement, we just redefined the fallback for not having FRAME_RATE_CAP set to be using the FRAME_RATE_CAP system property (which is set to -1).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/506)*

jrobble:
Review status: all files reviewed at latest revision, 14 unresolved discussions.

---

*[trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/camelOps/TestDetectionSplitter.java, line 619 at r2](https://reviewable.io:443/reviews/openmpf/openmpf/506#-L7_W89CQJASAdVD9OLo:-L7jvKMmXAFaOPw4F9XC:b-lwzs8l) ([raw file](https://github.com/openmpf/openmpf/blob/046ff1ccf529ba5586899d265789a5d092ffef3d/trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/camelOps/TestDetectionSplitter.java#L619)):*
> Previously, we said that if FRAME_RATE_CAP is not set, then fallback to FRAME_INTERVAL

Yes, we do this on a level-by-level basis.

At the media property level, if FRAME_RATE_CAP is not set, check if FRAME_INTERVAL is set at the same level. If not, drop down to the next level and look for both of those values there, starting with FRAME_RATE_CAP.

>  In the above statement, we just redefined the fallback for not having FRAME_RATE_CAP set to be using the FRAME_RATE_CAP system property (which is set to -1).

Please quote the part of the above statement you're referring to.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/506)*

dpierce17:
*[trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/camelOps/TestDetectionSplitter.java, line 619 at r2](https://reviewable.io:443/reviews/openmpf/openmpf/506#-L7_W89CQJASAdVD9OLo:-L7k3RKvHZV4clASiOwF:brcpsl3) ([raw file](https://github.com/openmpf/openmpf/blob/046ff1ccf529ba5586899d265789a5d092ffef3d/trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/camelOps/TestDetectionSplitter.java#L619)):*

> Previously, we said that if FRAME_RATE_CAP is not set, then fallback to FRAME_INTERVAL

Yes, we do this on a level-by-level basis.

At the media property level, if FRAME_RATE_CAP is not set, check if FRAME_INTERVAL is set at the same level. If not, drop down to the next level and look for both of those values there, starting with FRAME_RATE_CAP.

>  In the above statement, we just redefined the fallback for not having FRAME_RATE_CAP set to be using the FRAME_RATE_CAP system property (which is set to -1).

Please quote the part of the above statement you're referring to.

</blockquote></details>

Reading this again, I'm just not getting the override strategy when the property is disabled (i.e. set to -1). 
Considering that the FRAME_RATE_CAP / FRAME_INTERVAL overrides are checked on a level-by-level basis, does the following describe the override strategy we're looking for?  If not, what have I gotten wrong?

- If FRAME_RATE_CAP is present and not disabled, ignore FRAME_INTERVAL at the current and at any lower property level and use that FRAME_RATE_CAP to derive computed frame interval.
- If FRAME_RATE_CAP is present and disabled, and FRAME_INTERVAL is present and not disabled, then use the FRAME_INTERVAL value at this level.
- If FRAME_RATE_CAP is present and disabled, and FRAME_INTERVAL is not present then consider FRAME_INTERVAL at lower property levels until the default FRAME_INTERVAL system property is reached. FRAME_RATE_CAP at lower levels will be ignored because FRAME_RATE_CAP at a higher level was disabled.
- If FRAME_RATE_CAP is not present and FRAME_INTERVAL is not present check for FRAME_RATE_CAP or FRAME_INTERVAL at lower property levels until the default FRAME_RATE_CAP or FRAME_INTERVAL system properties are reached (preference would be for FRAME_RATE_CAP at the lower levels).
- If FRAME_RATE_CAP is not present and FRAME_INTERVAL is present and not disabled, then use the FRAME_INTERVAL value at this level.
- If FRAME_RATE_CAP is not present and FRAME_INTERVAL is present and disabled, then only consider FRAME_RATE_CAP at lower property levels until the default FRAME_RATE_CAP system property is reached. But, if FRAME_RATE_CAP system property is -1, then use FRAME_INTERVAL system property default. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/506)*

jrobble:
Review status: all files reviewed at latest revision, 14 unresolved discussions.

---

*[trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/camelOps/TestDetectionSplitter.java, line 619 at r2](https://reviewable.io:443/reviews/openmpf/openmpf/506#-L7_W89CQJASAdVD9OLo:-L7kdNHQk3gZCzQEQo-I:b-epq810) ([raw file](https://github.com/openmpf/openmpf/blob/046ff1ccf529ba5586899d265789a5d092ffef3d/trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/camelOps/TestDetectionSplitter.java#L619)):*
These are right with the following exception:

> If FRAME_RATE_CAP is not present and FRAME_INTERVAL is present and disabled, then only consider FRAME_RATE_CAP at lower property levels until the default FRAME_RATE_CAP system property is reached. But, if FRAME_RATE_CAP system property is -1, then use FRAME_INTERVAL system property default.

Specifically:

> But, if FRAME_RATE_CAP system property is -1, then use FRAME_INTERVAL system property default.

FRAME_INTERVAL was disabled at a higher level, so it retains the -1 setting. Then we determine that the FRAME_RATE_CAP setting is also -1 (at the system property level). That combination of both disabled properties should result in setting the frame interval to 1 (no matter what the frame interval system property is).

If you're referring to the the value of 1 as the "system property default", then you have this right too. Although, I think you will have to hardcode "1" in your code. There's no where to pull it from.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/506)*

dpierce17:
*[trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/camelOps/TestDetectionSplitter.java, line 619 at r2](https://reviewable.io:443/reviews/openmpf/openmpf/506#-L7_W89CQJASAdVD9OLo:-L7yHnofNDoNzdUY_WQt:b-49bqor) ([raw file](https://github.com/openmpf/openmpf/blob/046ff1ccf529ba5586899d265789a5d092ffef3d/trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/camelOps/TestDetectionSplitter.java#L619)):*

These are right with the following exception:

> If FRAME_RATE_CAP is not present and FRAME_INTERVAL is present and disabled, then only consider FRAME_RATE_CAP at lower property levels until the default FRAME_RATE_CAP system property is reached. But, if FRAME_RATE_CAP system property is -1, then use FRAME_INTERVAL system property default.

Specifically:

> But, if FRAME_RATE_CAP system property is -1, then use FRAME_INTERVAL system property default.

FRAME_INTERVAL was disabled at a higher level, so it retains the -1 setting. Then we determine that the FRAME_RATE_CAP setting is also -1 (at the system property level). That combination of both disabled properties should result in setting the frame interval to 1 (no matter what the frame interval system property is).

If you're referring to the the value of 1 as the "system property default", then you have this right too. Although, I think you will have to hardcode "1" in your code. There's no where to pull it from.

</blockquote></details>

Thanks Jeff. One last question, Should the software throw an error if either FRAME_RATE_CAP or FRAME_INTERVAL property is set to 0?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/506)*

jrobble:
*[trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/camelOps/TestDetectionSplitter.java, line 619 at r2](https://reviewable.io:443/reviews/openmpf/openmpf/506#-L7_W89CQJASAdVD9OLo:-L7yU15blKJUgXe1IRAB:bhptqvt) ([raw file](https://github.com/openmpf/openmpf/blob/046ff1ccf529ba5586899d265789a5d092ffef3d/trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/camelOps/TestDetectionSplitter.java#L619)):*

Thanks Jeff. One last question, Should the software throw an error if either FRAME_RATE_CAP or FRAME_INTERVAL property is set to 0?
</blockquote></details>

No. Treat anything <= 0 in the same way (disabled).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/506)*

dpierce17:
*[trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/camelOps/TestDetectionSplitter.java, line 445 at r2](https://reviewable.io:443/reviews/openmpf/openmpf/506#-L7_n8-ahuTe4-VPdr8C:-L8EQY-qgWGAhWX9IWQy:bct7tbq) ([raw file](https://github.com/openmpf/openmpf/blob/046ff1ccf529ba5586899d265789a5d092ffef3d/trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/camelOps/TestDetectionSplitter.java#L445)):*

Consider beefing up this method to make the rest of your code less verbose:

- Instead of passing in collections, for each collection instead pass in two Integers: one that's the value of the frame interval for that property collection, and one that's the value of the frame rate cap. Using the Integer object allows you to specify `null` to represent that a value isn't specified. You can then create the collections within this method.

- Building on the previous bullet, this method can also take in parameters for the mocked values of the system properties. (See other comment on how to use Mockito.)

- Pass in the expected computed frame interval and perform the check in this method.

- Have this method determine the job id and clear redis.

- Basically, make this method do as much as possible. It would be great if we can simplify the tests to the point where each test case simply calls this method. Please take a look at what Brian did with the gtests [here](https://github.com/openmpf/openmpf-cpp-component-sdk/blob/edae6ab322792a23617264f186385fd8f066e503/detection/api/test/test_detection_api.cpp#L68).
</blockquote></details>

I made this change as requested, except for adding mock support. Since I'm not that familiar with mock I'm going to submit that change in a later drop. What couldn't be supported was passing in the expected computed frame interval. Since the result depends upon the conditions of the test, there was a lot of variances involved with defining the assertions.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/506)*

dpierce17:
*[trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/camelOps/TestDetectionSplitter.java, line 502 at r2](https://reviewable.io:443/reviews/openmpf/openmpf/506#-L7_S32T_gt04JDz0bCj:-L8EROWuP0Xfk7-KDVQQ:b-ko382z) ([raw file](https://github.com/openmpf/openmpf/blob/046ff1ccf529ba5586899d265789a5d092ffef3d/trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/camelOps/TestDetectionSplitter.java#L502)):*
>  Since I'm not that familiar with mock I'm going to submit that change at a later time.  Since I have all of the 9 level-by-level media, algorithm, job and action property tests working (with media, algorithm and job property tests each with an additional 18 one-property-level down FRAME_INTERVAL or FRAME_RATE_CAP non-disabled tests), I wanted to get the media, algorithm, job and action property unit tests checked in.  The system property unit tests (which require usage of Mock) I'll address next.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/506)*

dpierce17:
*[trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/camelOps/TestDetectionSplitter.java, line 619 at r2](https://reviewable.io:443/reviews/openmpf/openmpf/506#-L7_W89CQJASAdVD9OLo:-L8ESxJVImqOHEcgLTLv:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/046ff1ccf529ba5586899d265789a5d092ffef3d/trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/camelOps/TestDetectionSplitter.java#L619)):*

No. Treat anything <= 0 in the same way (disabled).
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/506)*

dpierce17:
*[trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/camelOps/TestDetectionSplitter.java, line 619 at r2](https://reviewable.io:443/reviews/openmpf/openmpf/506#-L7_W89CQJASAdVD9OLo:-L8ETf_XSsre5XaEqc_i:b-m7wsbn) ([raw file](https://github.com/openmpf/openmpf/blob/046ff1ccf529ba5586899d265789a5d092ffef3d/trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/camelOps/TestDetectionSplitter.java#L619)):*

Done.
</blockquote></details>

I have all of the 9 level-by-level media, algorithm, job and action property tests working (9 + 9 + 9 + 9 = 36 test cases working without the mocked system properties which I'll address next). Since you also requested including  one-property-level down checks, this required an additional 18 tests for the media, algorithm and job property unit tests ( where 9 tests are required to support providing FRAME_INTERVAL one property level down and 9 tests are required to support providing FRAME_RATE_CAP one property level down).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/506)*

jrobble:
I branched off of add-frame-rate-cap and created [add-frame-rate-cap-jrobble](https://github.com/openmpf/openmpf/tree/add-frame-rate-cap-jrobble). Please take a look at [this commit](https://github.com/openmpf/openmpf/commit/a3aa33ec34fe44bf904cf7eb8dc382b0d76f6f18).

Using what you've done as inspiration, I refactored your approach by adding a `calculateFrameInterval` method to `AggregateJobPropertiesUtil`. That allowed me to simplify the test code quite a bit, and does away with the need to use Mockito for the system properties.

I commented out a lot of your code in my branch. Everything I commented out can probably be removed, except for some annotations in the test class. Also, I added some TODOs. Please merge my branch into yours, address the TODOs (which are mostly about removing a class and finishing the rest of the tests), and update this PR.

---

Reviewed 7 of 7 files at r3.
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/DetectionSplitter.java, line 345 at r3](https://reviewable.io:443/reviews/openmpf/openmpf/506#-L8LNIQ7KZ1F1KWYqEc8:-L8LNIQ7KZ1F1KWYqEc9:bhfzelm) ([raw file](https://github.com/openmpf/openmpf/blob/6e93741b04c3f9ef121755e65fbe23334ef3eb84/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/DetectionSplitter.java#L345)):*
> ```Java
> 
>                 // TODO, I don't see any protection in here for a user specifying FRAME_INTERVAL of -1 for non-video media. The
>                 // createSegmentingPlan method will send a warning about using FRAME_INTERVAL set to 1 if this occurs. Is this ok?
> ```

Good point. I think the segmenting plan is only used by the `VideoMediaSegmenter`. Move the code which creates the segmenting plan to the `if ( transientMedia.containsMetadata("FPS") ) {` block. Runs some video and non-video jobs to make sure that works as expected and doesn't cause any problems.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/transients/TransientJob.java, line 63 at r2](https://reviewable.io:443/reviews/openmpf/openmpf/506#-L7_OFxpCQRTGpxNy0Yb:-L8LUdrPbRWsXMO7O-ZJ:b-r2sf9v) ([raw file](https://github.com/openmpf/openmpf/blob/046ff1ccf529ba5586899d265789a5d092ffef3d/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/transients/TransientJob.java#L63)):*

This method is not used. Please remove it.
</blockquote></details>

I removed this in my branch.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/openmpf/openmpf/506#-:-L8LSFE33mMmyJe1HBTb:b-ha6wcc)*

dpierce17:
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/DetectionSplitter.java, line 345 at r3](https://beta.reviewable.io/reviews/openmpf/openmpf/506#-L8LNIQ7KZ1F1KWYqEc8:-L8co5Z399VP6lAH9rNa:bllx8ka) ([raw file](https://github.com/openmpf/openmpf/blob/6e93741b04c3f9ef121755e65fbe23334ef3eb84/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/DetectionSplitter.java#L345)):*

Good point. I think the segmenting plan is only used by the `VideoMediaSegmenter`. Move the code which creates the segmenting plan to the `if ( transientMedia.containsMetadata("FPS") ) {` block. Runs some video and non-video jobs to make sure that works as expected and doesn't cause any problems.
</blockquote></details>

I made this change, and ran two jobs each with a face detection pipeline, with a .jpg image that had two faces in it.  One job produced no tracks, and the other produced 2 tracks but had a really low confidence level. I'm going the check the software in but this needs to be checked by somebody who is more familiar with what might be wrong with this (why no tracks and confidence so low?)

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/openmpf/openmpf/506)*

jrobble:
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/DetectionSplitter.java, line 345 at r3](https://beta.reviewable.io/reviews/openmpf/openmpf/506#-L8LNIQ7KZ1F1KWYqEc8:-L8ctQn69ubedJUMMrGw:b-sc51vi) ([raw file](https://github.com/openmpf/openmpf/blob/6e93741b04c3f9ef121755e65fbe23334ef3eb84/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/DetectionSplitter.java#L345)):*

I made this change, and ran two jobs each with a face detection pipeline, with a .jpg image that had two faces in it.  One job produced no tracks, and the other produced 2 tracks but had a really low confidence level. I'm going the check the software in but this needs to be checked by somebody who is more familiar with what might be wrong with this (why no tracks and confidence so low?)
</blockquote></details>

What output do the jobs produce in the develop branch? Is the output different?

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/openmpf/openmpf/506)*

dpierce17:
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/DetectionSplitter.java, line 345 at r3](https://beta.reviewable.io/reviews/openmpf/openmpf/506#-L8LNIQ7KZ1F1KWYqEc8:-L8cudET0_r38bxxvOt0:b-em51gm) ([raw file](https://github.com/openmpf/openmpf/blob/6e93741b04c3f9ef121755e65fbe23334ef3eb84/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/DetectionSplitter.java#L345)):*

What output do the jobs produce in the develop branch? Is the output different?
</blockquote></details>

I'm not sure, I'll run this test tomorrow as I have to leave in a few minutes.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/openmpf/openmpf/506)*

jrobble:
Reviewed 1 of 11 files at r1, 5 of 8 files at r4, 1 of 2 files at r6, 3 of 3 files at r7.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/AdaptiveFrameIntervalPropertyState.java, line 37 at r7](https://beta.reviewable.io/reviews/openmpf/openmpf/506#-L8ctdlWEA-Bv88UwiXV:-L8ctdlWEA-Bv88UwiXW:bvcnd1d) ([raw file](https://github.com/openmpf/openmpf/blob/4fddd1dd0dffd4196c79b996d25165c95b9e0c9a/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/AdaptiveFrameIntervalPropertyState.java#L37)):*
> ```Java
> import java.util.Map;
> 
> // TODO: Remove this class.
> ```

Remove this class.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/DetectionContext.java, line 67 at r7](https://beta.reviewable.io/reviews/openmpf/openmpf/506#-L8dAVGd0jLcMjl_RZN-:-L8cu9Ok2IAGaeKcQNGH:bgzkc7g) ([raw file](https://github.com/openmpf/openmpf/blob/4fddd1dd0dffd4196c79b996d25165c95b9e0c9a/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/DetectionContext.java#L67)):*
> ```Java
> 
> 	// Constructor for jobs that don't require a segmenting plan, such as non-video jobs.
>     public DetectionContext(
> ```

I don't think this constructor is necessary. Just pass null as the last argument to the original constructor wherever you would invoke this constructor.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/DetectionSplitter.java, line 345 at r3](https://beta.reviewable.io/reviews/openmpf/openmpf/506#-L8LNIQ7KZ1F1KWYqEc8:-L8dCNaIAO8YcCjllDqe:bms7hbk) ([raw file](https://github.com/openmpf/openmpf/blob/6e93741b04c3f9ef121755e65fbe23334ef3eb84/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/DetectionSplitter.java#L345)):*

I'm not sure, I'll run this test tomorrow as I have to leave in a few minutes.
</blockquote></details>

I didn't observe any issues running OCV face detection jobs on images on my local machine. Also, those Jenkins face detection in images tests passed. However, you should still check your test results against develop.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/DetectionSplitter.java, line 163 at r7](https://beta.reviewable.io/reviews/openmpf/openmpf/506#-L8d2_054SFWTdnD1Ter:-L8d2_054SFWTdnD1Tes:bup233l) ([raw file](https://github.com/openmpf/openmpf/blob/4fddd1dd0dffd4196c79b996d25165c95b9e0c9a/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/DetectionSplitter.java#L163)):*
> ```Java
> 			// If this is not the first detection stage, we should build segments based off of the previous stage's
> 			// tracks. Note that the TimePairs created for these Tracks use the non-feed-forward version of timeUtils.createTimePairsForTracks
> 			// TODO look here for any modifications required to be made to support feed-forward
> ```

Please remove this TODO.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/DetectionSplitter.java, line 280 at r7](https://beta.reviewable.io/reviews/openmpf/openmpf/506#-L8dAbYz9Bm1xKZjpSNt:-L8dAbYz9Bm1xKZjpSNu:b-5rzxyo) ([raw file](https://github.com/openmpf/openmpf/blob/4fddd1dd0dffd4196c79b996d25165c95b9e0c9a/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/DetectionSplitter.java#L280)):*
> ```Java
>                         segmentingPlan);
> 
> 				} else {
> ```

Instead of this if/else, let's do:

```
SegmentingPlan segmentingPlan = null;

if (transientMedia.containsMetadata("FPS")) {
    String calcframeInterval = AggregateJobPropertiesUtil.calculateFrameInterval(
							transientAction, transientJob, transientMedia,
							propertiesUtil.getSamplingInterval(), propertiesUtil.getFrameRateCap(),
							Double.valueOf(transientMedia.getMetadata("FPS")));
    modifiedMap.put(MpfConstants.MEDIA_SAMPLING_INTERVAL_PROPERTY, calcframeInterval);

    // Segmenting plan is only used by the VideoMediaSegmenter
    segmentingPlan = createSegmentingPlan(modifiedMap);
}

List<AlgorithmPropertyProtocolBuffer.AlgorithmProperty> algorithmProperties
                        = convertPropertiesMapToAlgorithmPropertiesList(modifiedMap);

DetectionContext detectionContext = new DetectionContext(
                        transientJob.getId(),
                        transientJob.getCurrentStage(),
                        transientStage.getName(),
                        actionIndex,
                        transientAction.getName(),
                        isFirstDetectionStage,
                        algorithmProperties,
                        previousTracks,
                        segmentingPlan);
```

I think that's less verbose.

---

*[trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/camelOps/TestDetectionSplitter.java, line 98 at r7](https://beta.reviewable.io/reviews/openmpf/openmpf/506#-L8d9_Tq3JfMaHUTfOcL:-L8d9VrD2yLiZ8QoT2OJ:b-foru13) ([raw file](https://github.com/openmpf/openmpf/blob/4fddd1dd0dffd4196c79b996d25165c95b9e0c9a/trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/camelOps/TestDetectionSplitter.java#L98)):*
> ```Java
>         final boolean testOutputEnabled = true;
>         TransientJob testJob = new TransientJob(testId, testExternalId, testPipe, testStage, testPriority, testOutputEnabled, false);
>         TransientMedia testMedia = new TransientMedia(nextId(), ioUtils.findFile("/samples/new_face_video.avi").toString());
> ```

        
Add this center line to fix Jenkins failure:

```
TransientMedia testMedia = new TransientMedia(nextId(), ioUtils.findFile("/samples/new_face_video.avi").toString());
testMedia.addMetadata("FPS", "30");
testMedia.setType("VIDEO");
```

---

*[trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/camelOps/TestDetectionSplitter.java, line 353 at r7](https://beta.reviewable.io/reviews/openmpf/openmpf/506#-L8d9sv16o05hATrBRxs:-L8d9sv16o05hATrBRxt:b-eut8mw) ([raw file](https://github.com/openmpf/openmpf/blob/4fddd1dd0dffd4196c79b996d25165c95b9e0c9a/trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/camelOps/TestDetectionSplitter.java#L353)):*
> ```Java
>         final boolean testOutputEnabled = true;
>         TransientJob testJob = new TransientJob(testJobId, testExternalId, testPipe, testStage, testPriority, testOutputEnabled, false);
>         TransientMedia testMedia = new TransientMedia(nextId(), ioUtils.findFile(mediaUri).toString());
> ```

Add this center line to fix Jenkins failure:

```
TransientMedia testMedia = new TransientMedia(nextId(), ioUtils.findFile(mediaUri).toString());
testMedia.addMetadata("FPS", "30");
testMedia.setLength(300);
```

---

*[trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/camelOps/TestDetectionSplitter.java, line 372 at r7](https://beta.reviewable.io/reviews/openmpf/openmpf/506#-L8cvq2_6rwpggIKiRoa:-L8cvq2a3GRCqwy_ryE_:bgyv0e2) ([raw file](https://github.com/openmpf/openmpf/blob/4fddd1dd0dffd4196c79b996d25165c95b9e0c9a/trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/camelOps/TestDetectionSplitter.java#L372)):*
> ```Java
>     // overrides at various category levels (system, action, job, algorithm, and media, with media properties being the highest ranking).
> 
>     private TransientJob createFrameRateCapTestTransientJob(
> ```

Rename this `createSimpleJobForFrameRateCapTest`.

---

*[trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/camelOps/TestDetectionSplitter.java, line 466 at r7](https://beta.reviewable.io/reviews/openmpf/openmpf/506#-L8d0HkB1ZqqNiII5Wfk:-L8d0HkB1ZqqNiII5Wfl:b-r5abei) ([raw file](https://github.com/openmpf/openmpf/blob/4fddd1dd0dffd4196c79b996d25165c95b9e0c9a/trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/camelOps/TestDetectionSplitter.java#L466)):*
> ```Java
>         // Tests 1-9: test 9 combinations of action property FRAME_INTERVAL and FRAME_RATE_CAP with the lower property levels not specified.
> 
>         // Argument order for checkCalcFrameInterval is: frameInterval, frameRateCap pair for property levels in this order: system, action, job, algorithm, media -
> ```

I don't think we need this comment in every test method. If you want to keep it, please put in the first test method and remove it from the rest: `testFrameRateCapOverrideSystemLevel`.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/openmpf/openmpf/506)*

dpierce17:
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/DetectionSplitter.java, line 280 at r7](https://beta.reviewable.io/reviews/openmpf/openmpf/506#-L8dAbYz9Bm1xKZjpSNt:-L8gNx5J4aEoj_RgVieq:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/4fddd1dd0dffd4196c79b996d25165c95b9e0c9a/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/DetectionSplitter.java#L280)):*

Instead of this if/else, let's do:

```
SegmentingPlan segmentingPlan = null;

if (transientMedia.containsMetadata("FPS")) {
    String calcframeInterval = AggregateJobPropertiesUtil.calculateFrameInterval(
							transientAction, transientJob, transientMedia,
							propertiesUtil.getSamplingInterval(), propertiesUtil.getFrameRateCap(),
							Double.valueOf(transientMedia.getMetadata("FPS")));
    modifiedMap.put(MpfConstants.MEDIA_SAMPLING_INTERVAL_PROPERTY, calcframeInterval);

    // Segmenting plan is only used by the VideoMediaSegmenter
    segmentingPlan = createSegmentingPlan(modifiedMap);
}

List<AlgorithmPropertyProtocolBuffer.AlgorithmProperty> algorithmProperties
                        = convertPropertiesMapToAlgorithmPropertiesList(modifiedMap);

DetectionContext detectionContext = new DetectionContext(
                        transientJob.getId(),
                        transientJob.getCurrentStage(),
                        transientStage.getName(),
                        actionIndex,
                        transientAction.getName(),
                        isFirstDetectionStage,
                        algorithmProperties,
                        previousTracks,
                        segmentingPlan);
```

I think that's less verbose.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/openmpf/openmpf/506)*

dpierce17:
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/DetectionSplitter.java, line 280 at r7](https://beta.reviewable.io/reviews/openmpf/openmpf/506#-L8dAbYz9Bm1xKZjpSNt:-L8gO0q-70tLzujpl6TB:b-jon00a) ([raw file](https://github.com/openmpf/openmpf/blob/4fddd1dd0dffd4196c79b996d25165c95b9e0c9a/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/DetectionSplitter.java#L280)):*

Done.
</blockquote></details>

3 of the legacy tests in TestDetectionSplitter that use sample file new_face_video.avi are failing with a NullPointerException in VideoMediaSegmenter.  The reason this would happen is either: (1) MediaInspectionProcessor isn't getting called to set FPS in metadata or (2) the .avi file doesn't have videoFrameRate defined in the metadata.  (1) appears to be the cause, so I had to add code to the legacy tests to include FPS in the video media metadata.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/openmpf/openmpf/506)*

dpierce17:
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/DetectionSplitter.java, line 345 at r3](https://beta.reviewable.io/reviews/openmpf/openmpf/506#-L8LNIQ7KZ1F1KWYqEc8:-L8hrQ8q9txFLCF3WvAo:bb0em29) ([raw file](https://github.com/openmpf/openmpf/blob/6e93741b04c3f9ef121755e65fbe23334ef3eb84/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/DetectionSplitter.java#L345)):*

I didn't observe any issues running OCV face detection jobs on images on my local machine. Also, those Jenkins face detection in images tests passed. However, you should still check your test results against develop.
</blockquote></details>

I confirmed that I get the same results when I run with develop.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/openmpf/openmpf/506)*

jrobble:
Reviewed 6 of 6 files at r8.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/openmpf/openmpf/506)*

PR #547, opened 2018-04-21
Fixed video segmenting on single-frame gifs.

jrobble:
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/547)


jrobble:
I recommend looking at this in Reviewable with the option to view whitespace changes turned off. I updated the file to use spaces instead of tabs.

PR #548, opened 2018-04-26
det system properties changes applied without mpf restart (new actions/pipelines only)

dpierce17:
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/548)


dpierce17:
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/mvc/controller/AdminPropertySettingsController.java, line 203 at r1](https://reviewable.io/reviews/openmpf/openmpf/548#-LB2pyzS7m76X8rWal6f:-LBLMdcH5JiQh9Y7geIr:bftaa54) ([raw file](https://github.com/openmpf/openmpf/blob/3e7bf83146ccc51c223ee514073369a54f09f023/trunk/workflow-manager/src/main/java/org/mitre/mpf/mvc/controller/AdminPropertySettingsController.java#L203)):*

Let's update `PropertiesUtil.setAndSaveCustomProperties()` to return the filtered list of property models instead of the whole list. Right now the return value of that method is never used for anything, so you shouldn't have to worry about breaking anything when you make the change.
</blockquote></details>

PropertiesUtil.setAndSaveCustomProperties() returns a void, so I don't understand what you're looking for. This method does set a class property which is an unfiltered ImmutableConfiguration, and I don't think we want to change this behavior.

---


*Comments from [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/548)*

dpierce17:
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/DetectionSplitter.java, line 161 at r1](https://reviewable.io/reviews/openmpf/openmpf/548#-LB6Wx8K2kI10bHKWrDu:-LBLP0hT4gIveL-SnKxY:b-985tgt) ([raw file](https://github.com/openmpf/openmpf/blob/3e7bf83146ccc51c223ee514073369a54f09f023/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/DetectionSplitter.java#L161)):*

Because we refresh the algorithm definitions when detection system properties are set through the web UI, this line will initialize the modifiedMap with the most recent detection system properties. This is a problem when the detection system properties are updated between two stages of a pipeline. I noticed this when running through the test procedure I sent you with the detection.track.min.gap property rather than the detection.sampling.interval property. (Note that I was able to use the existing “﻿OCV FACE DETECTION (WITH MOG MOTION PREPROCESSOR) PIPELINE” for that, since none of the actions in the pipeline set MIN_GAP_BETWEEN_TRACKS.)

To address this, update the getAlgorithmProperties() method in this class to iterate over every algorithm PropertyDefinition. For those which have a proprietiesKey, pull the value from the TransientDetectionSystemProperties snapshot.

To support this, you can add a lookup() method to TransientDetectionSystemProperties class that will take a key (String) and return a value (String). Under the hood it can be implemented by calling get() on the internal HashMap. If the key doesn't exist in the map, then it will return null. When a null is returned, the getAlgorithmProperties() method should use the default value of that PropertyDefinition.
</blockquote></details>

Do you mean "For the detection properties, pull the value from the TransientDetectionSystemProperties" instead of "For those which have a proprietiesKey, pull the value from the TransientDetectionSystemProperties"? I'm assuming this is the case when making this change

---


*Comments from [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/548)*

dpierce17:
*[trunk/workflow-manager/src/main/webapp/WEB-INF/html/admin/property_settings/layout.html, line 29 at r1](https://reviewable.io/reviews/openmpf/openmpf/548#-LB2TMBA-lMxJoy3O_Gs:-LBLnd8o9h6WizWg17Ed:b-193j0r) ([raw file](https://github.com/openmpf/openmpf/blob/3e7bf83146ccc51c223ee514073369a54f09f023/trunk/workflow-manager/src/main/webapp/WEB-INF/html/admin/property_settings/layout.html#L29)):*
> This view allows an admin user to modify OpenMPF properties. Updating the General System Properties requires an administrator to restart the server for the changes to take effect. Updating the Default Job Properties does not. The new Default Job Properties values will be used for all new jobs. Currently running jobs will not be affected. Use the buttons at the bottom of the page to save and reset all properties.
Done, but added this "however new Actions / Pipelines may need to be created if property override is of concern." because while this processing is inherent the original statement hide the facts that most likely new Actions and new Pipelines most likely need to be created to use the new detection system property values. Also, changed "Default Job Properties" to "Default Detection Properties"

---


*Comments from [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/548)*

jrobble:
Review status: all files reviewed at latest revision, 35 unresolved discussions.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/mvc/controller/AdminPropertySettingsController.java, line 203 at r1](https://reviewable.io/reviews/openmpf/openmpf/548#-LB2pyzS7m76X8rWal6f:-LBLus6t--mXtDddgR-u:b-c2on4z) ([raw file](https://github.com/openmpf/openmpf/blob/3e7bf83146ccc51c223ee514073369a54f09f023/trunk/workflow-manager/src/main/java/org/mitre/mpf/mvc/controller/AdminPropertySettingsController.java#L203)):*

PropertiesUtil.setAndSaveCustomProperties() returns a void, so I don't understand what you're looking for. This method does set a class property which is an unfiltered ImmutableConfiguration, and I don't think we want to change this behavior.
</blockquote></details>

Sorry, I was getting this method confused with `MpfPropertiesConfigurationBuilder.setAndSaveCustomProperties()`. That method's return value is used. Please ignore what I said before.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/DetectionSplitter.java, line 161 at r1](https://reviewable.io/reviews/openmpf/openmpf/548#-LB6Wx8K2kI10bHKWrDu:-LBLx8bU00vTX58GgWWe:b-20imh3) ([raw file](https://github.com/openmpf/openmpf/blob/3e7bf83146ccc51c223ee514073369a54f09f023/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/DetectionSplitter.java#L161)):*
> "For the detection properties, pull the value from the TransientDetectionSystemProperties"

It's not that simple. You only want to pull the value from your POJO if the PropertyDefinition has a `proprietiesKey` set. If it's not set, then the `defaultValue` should be used (which is what `getAlgorithmProperties()` currently does).

Remember that the `propertiesKey` denotes that the value should be pulled from a system property - in this case, the cached system property in your POJO. 

You're right in that we only care about updating the detection system properties using your POJO. You can perform the `TransientDetectionSystemProperties.lookup()` method I described above. If it returns null, which it will for all non-detection properties, then just use the `defaultValue` (again, which is what `getAlgorithmProperties()` currently does).

---

*[trunk/workflow-manager/src/main/webapp/resources/mytheme/js/controllers/admin/AdminPropertySettingsCtrl.js, line 74 at r1](https://reviewable.io/reviews/openmpf/openmpf/548#-LBM6Oz27llu7VN8BEhi:-LBM6Oz27llu7VN8BEhj:bddx8nv) ([raw file](https://github.com/openmpf/openmpf/blob/3e7bf83146ccc51c223ee514073369a54f09f023/trunk/workflow-manager/src/main/webapp/resources/mytheme/js/controllers/admin/AdminPropertySettingsCtrl.js#L74)):*
> ```JavaScript
> 
>     // Get the list of all system properties.
>     queryAll: function () {
> ```

This function does not seemed to be used. If that's the case, please remove it.

---

*[trunk/workflow-manager/src/main/webapp/resources/mytheme/js/controllers/admin/AdminPropertySettingsCtrl.js, line 107 at r1](https://reviewable.io/reviews/openmpf/openmpf/548#-LBM6qklARToHP1fPMDr:-LBM6qklARToHP1fPMDs:b-3a8d9m) ([raw file](https://github.com/openmpf/openmpf/blob/3e7bf83146ccc51c223ee514073369a54f09f023/trunk/workflow-manager/src/main/webapp/resources/mytheme/js/controllers/admin/AdminPropertySettingsCtrl.js#L107)):*
> ```JavaScript
>     // Get the list of all immutable system properties.
>     queryImmutable: function () {
>       serverProperties = { };
> ```

Both this function and `queryMutable()` clear-out the serverProperties map. That seems odd. For example, if you call `queryImmutable()` after `queryMutable()`, the first thing it will do is clear-out all of the mutable properties from the map. Is that intentional?

---

*[trunk/workflow-manager/src/main/webapp/WEB-INF/html/admin/property_settings/layout.html, line 29 at r1](https://reviewable.io/reviews/openmpf/openmpf/548#-LB2TMBA-lMxJoy3O_Gs:-LBM0ISA4Xpd7iWg661Q:bfllwae) ([raw file](https://github.com/openmpf/openmpf/blob/3e7bf83146ccc51c223ee514073369a54f09f023/trunk/workflow-manager/src/main/webapp/WEB-INF/html/admin/property_settings/layout.html#L29)):*

> This view allows an admin user to modify OpenMPF properties. Updating the General System Properties requires an administrator to restart the server for the changes to take effect. Updating the Default Job Properties does not. The new Default Job Properties values will be used for all new jobs. Currently running jobs will not be affected. Use the buttons at the bottom of the page to save and reset all properties.
Done, but added this "however new Actions / Pipelines may need to be created if property override is of concern." because while this processing is inherent the original statement hide the facts that most likely new Actions and new Pipelines most likely need to be created to use the new detection system property values. Also, changed "Default Job Properties" to "Default Detection Properties"
</blockquote></details>

Okay. Works for me.

---


*Comments from [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/548)*

jrobble:
*[trunk/workflow-manager/src/main/webapp/resources/mytheme/js/controllers/admin/AdminPropertySettingsCtrl.js, line 107 at r1](https://reviewable.io/reviews/openmpf/openmpf/548#-LBM6qklARToHP1fPMDr:-LBMJktn-39xPm_U5gjk:bxsey6c) ([raw file](https://github.com/openmpf/openmpf/blob/3e7bf83146ccc51c223ee514073369a54f09f023/trunk/workflow-manager/src/main/webapp/resources/mytheme/js/controllers/admin/AdminPropertySettingsCtrl.js#L107)):*

Both this function and `queryMutable()` clear-out the serverProperties map. That seems odd. For example, if you call `queryImmutable()` after `queryMutable()`, the first thing it will do is clear-out all of the mutable properties from the map. Is that intentional?
</blockquote></details>

The reason your code currently works may be because of a race condition where an empty serverProperties is cleared out before the server handles the requests to get the immutable/mutable properties.

---


*Comments from [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/548)*

dpierce17:
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/DetectionSplitter.java, line 161 at r1](https://reviewable.io/reviews/openmpf/openmpf/548#-LB6Wx8K2kI10bHKWrDu:-LBLa8Cy3xBcmu8dBvwj:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/3e7bf83146ccc51c223ee514073369a54f09f023/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/DetectionSplitter.java#L161)):*

Do you mean "For the detection properties, pull the value from the TransientDetectionSystemProperties" instead of "For those which have a proprietiesKey, pull the value from the TransientDetectionSystemProperties"? I'm assuming this is the case when making this change
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/548)*

dpierce17:
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/DetectionSplitter.java, line 161 at r1](https://reviewable.io/reviews/openmpf/openmpf/548#-LB6Wx8K2kI10bHKWrDu:-LBMXeCP67RoZlkKJebS:bjtct9z) ([raw file](https://github.com/openmpf/openmpf/blob/3e7bf83146ccc51c223ee514073369a54f09f023/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/DetectionSplitter.java#L161)):*

Done.
</blockquote></details>

Thanks for the clarification. Using your latest comment, I revised my implementation and the test that changes MIN_GAP_BETWEEN_TRACKS using the suggested pipeline appears to work

---


*Comments from [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/548)*

dpierce17:
*[trunk/workflow-manager/src/main/webapp/resources/mytheme/js/controllers/admin/AdminPropertySettingsCtrl.js, line 74 at r1](https://reviewable.io/reviews/openmpf/openmpf/548#-LBM6Oz27llu7VN8BEhi:-LBMw31bAswILqWaCFLe:bthhe01) ([raw file](https://github.com/openmpf/openmpf/blob/3e7bf83146ccc51c223ee514073369a54f09f023/trunk/workflow-manager/src/main/webapp/resources/mytheme/js/controllers/admin/AdminPropertySettingsCtrl.js#L74)):*

This function does not seemed to be used. If that's the case, please remove it.
</blockquote></details>

Marked this as TODO

---


*Comments from [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/548)*

dpierce17:
*[trunk/workflow-manager/src/main/webapp/resources/mytheme/js/controllers/admin/AdminPropertySettingsCtrl.js, line 107 at r1](https://reviewable.io/reviews/openmpf/openmpf/548#-LBM6qklARToHP1fPMDr:-LBMw9RC6BUW5yDKZkJD:b6epoou) ([raw file](https://github.com/openmpf/openmpf/blob/3e7bf83146ccc51c223ee514073369a54f09f023/trunk/workflow-manager/src/main/webapp/resources/mytheme/js/controllers/admin/AdminPropertySettingsCtrl.js#L107)):*

The reason your code currently works may be because of a race condition where an empty serverProperties is cleared out before the server handles the requests to get the immutable/mutable properties.
</blockquote></details>

In functions queryMutable and queryImmutable, added two TODOs to clear out existing properties prior to getting an updated set of properties of the specified type from the server

---


*Comments from [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/548)*

dpierce17:
*[trunk/workflow-manager/src/main/webapp/resources/mytheme/js/controllers/admin/AdminPropertySettingsCtrl.js, line 107 at r1](https://reviewable.io/reviews/openmpf/openmpf/548#-LBM6qklARToHP1fPMDr:-LBMwxrx1eFpfneJ-nu2:b-55ia30) ([raw file](https://github.com/openmpf/openmpf/blob/3e7bf83146ccc51c223ee514073369a54f09f023/trunk/workflow-manager/src/main/webapp/resources/mytheme/js/controllers/admin/AdminPropertySettingsCtrl.js#L107)):*

In functions queryMutable and queryImmutable, added two TODOs to clear out existing properties prior to getting an updated set of properties of the specified type from the server
</blockquote></details>

Added TODOs to check for possible race conditions

---


*Comments from [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/548)*

dpierce17:
*[trunk/workflow-manager/src/main/webapp/resources/mytheme/js/controllers/admin/AdminPropertySettingsCtrl.js, line 193 at r1](https://reviewable.io/reviews/openmpf/openmpf/548#-LB2hjmG0LrwO9yUR5-I:-LBMxM1OBAK1TNtuTzRW:b-g6div1) ([raw file](https://github.com/openmpf/openmpf/blob/3e7bf83146ccc51c223ee514073369a54f09f023/trunk/workflow-manager/src/main/webapp/resources/mytheme/js/controllers/admin/AdminPropertySettingsCtrl.js#L193)):*

I talked to @brosenberg42 about this briefly. It would be more efficient to make one REST call rather then two. You can combine the mutable and immutable properties into one collection and make the call with that.
</blockquote></details>

Added a TODO marking where two calls should be replaced by the single REST call

---


*Comments from [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/548)*

jrobble:
Reviewed 15 of 22 files at r2.
Review status: 19 of 26 files reviewed at latest revision, 8 unresolved discussions.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/DetectionSplitter.java, line 444 at r2](https://reviewable.io/reviews/openmpf/openmpf/548#-LBN1XMzDNGjiCvrE9WI:-LBN1XMzDNGjiCvrE9WJ:b-mbvba9) ([raw file](https://github.com/openmpf/openmpf/blob/92f430e1304fa80337a39125cf0bca0051d18888/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/DetectionSplitter.java#L444)):*
Implementation looks right at a glance, but the comment needs to be updated:

> Info: the propertiesKey set to **a non-null value** denotes that the value should be pulled from a

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/transients/TransientDetectionSystemProperties.java, line 120 at r2](https://reviewable.io/reviews/openmpf/openmpf/548#-LBN2iFS5MD0n51iTUep:-LBN2iFS5MD0n51iTUeq:b-qzmfag) ([raw file](https://github.com/openmpf/openmpf/blob/92f430e1304fa80337a39125cf0bca0051d18888/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/transients/TransientDetectionSystemProperties.java#L120)):*
> ```Java
> 
>     public String lookup(String propertyName) {
>         if ( detectionSystemPropertiesSnapshot.containsKey(propertyName ) ) {
> ```

I don't think we need the `containsKey` check here. `get` will return null if it doesn't contain the key. You can cast a null to a String. It will still be null.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/PropertiesUtil.java, line 213 at r2](https://reviewable.io/reviews/openmpf/openmpf/548#-LBN3SCz2OX7zIjCkAdn:-LBN3SD01dfxAGqiE1Mg:b-mtk0xt) ([raw file](https://github.com/openmpf/openmpf/blob/92f430e1304fa80337a39125cf0bca0051d18888/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/PropertiesUtil.java#L213)):*
> ```Java
>     public TransientDetectionSystemProperties createDetectionSystemPropertiesSnapshot() {
>         TransientDetectionSystemProperties transientDetectionSystemProperties = new TransientDetectionSystemProperties(getDetectionConfiguration());
>         return transientDetectionSystemProperties;
> ```

This can be shortened to one line that does the `new` and `return`.

---

*[trunk/workflow-manager/src/main/webapp/resources/mytheme/js/controllers/admin/AdminPropertySettingsCtrl.js, line 52 at r2](https://reviewable.io/reviews/openmpf/openmpf/548#-LBN3fNb093iP5-jpVCG:-LBN3fNb093iP5-jpVCH:b-ege6jj) ([raw file](https://github.com/openmpf/openmpf/blob/92f430e1304fa80337a39125cf0bca0051d18888/trunk/workflow-manager/src/main/webapp/resources/mytheme/js/controllers/admin/AdminPropertySettingsCtrl.js#L52)):*
> ```JavaScript
> 
>       propertiesResource.prototype.valueChanged = function () {
>         // TODO: investigate possible race condition, serverProperties may be used before it is set by queryMutable or queryImmutable functions
> ```

Check these race conditions.

---

*[trunk/workflow-manager/src/main/webapp/resources/mytheme/js/controllers/admin/AdminPropertySettingsCtrl.js, line 79 at r2](https://reviewable.io/reviews/openmpf/openmpf/548#-LBN4XKQ0GEzNm8iqA9n:-LBN4XKS4OyljytEjlGe:bg6ig1c) ([raw file](https://github.com/openmpf/openmpf/blob/92f430e1304fa80337a39125cf0bca0051d18888/trunk/workflow-manager/src/main/webapp/resources/mytheme/js/controllers/admin/AdminPropertySettingsCtrl.js#L79)):*
> ```JavaScript
>           if ( serverProperties ) {
>             for (var key in serverProperties){
>               if (serverProperties.hasOwnProperty(key)) {
> ```

Is `hasOwnProperty` necessary here?

This code should be refactored to return true as soon as the first property that needs to be restarted is detected.

---

*[trunk/workflow-manager/src/main/webapp/resources/mytheme/js/controllers/admin/AdminPropertySettingsCtrl.js, line 93 at r2](https://reviewable.io/reviews/openmpf/openmpf/548#-LBN5-4OEzokHEDasIWI:-LBN5-4OEzokHEDasIWJ:b-hnfj4w) ([raw file](https://github.com/openmpf/openmpf/blob/92f430e1304fa80337a39125cf0bca0051d18888/trunk/workflow-manager/src/main/webapp/resources/mytheme/js/controllers/admin/AdminPropertySettingsCtrl.js#L93)):*
> ```JavaScript
>         // TODO: this function might not be required. If not, remove it.
>         // Get the list of all system properties.
>         queryAll: function () {
> ```

See if this function needs to be removed.

---

*[trunk/workflow-manager/src/main/webapp/resources/mytheme/js/controllers/admin/AdminPropertySettingsCtrl.js, line 113 at r2](https://reviewable.io/reviews/openmpf/openmpf/548#-LBN55lx0Dc1gOkw2-Wk:-LBN55lx0Dc1gOkw2-Wl:bomxudk) ([raw file](https://github.com/openmpf/openmpf/blob/92f430e1304fa80337a39125cf0bca0051d18888/trunk/workflow-manager/src/main/webapp/resources/mytheme/js/controllers/admin/AdminPropertySettingsCtrl.js#L113)):*
> ```JavaScript
>         // Get the list of all mutable system properties.
>         queryMutable: function () {
>           // TODO: only the mutable system properties should be cleared before getting a fresh set of mutable properties from the server
> ```

Fix how serverProperties is cleared here, and in the function below.

---


*Comments from [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/548)*

jrobble:
*[trunk/workflow-manager/src/main/webapp/resources/mytheme/js/controllers/admin/AdminPropertySettingsCtrl.js, line 79 at r2](https://reviewable.io/reviews/openmpf/openmpf/548#-LBN4XKQ0GEzNm8iqA9n:-LBRwhjaEwbJT2AYHAWJ:b8dp64q) ([raw file](https://github.com/openmpf/openmpf/blob/92f430e1304fa80337a39125cf0bca0051d18888/trunk/workflow-manager/src/main/webapp/resources/mytheme/js/controllers/admin/AdminPropertySettingsCtrl.js#L79)):*

Is `hasOwnProperty` necessary here?

This code should be refactored to return true as soon as the first property that needs to be restarted is detected.
</blockquote></details>

Yes, it is necessary to avoid picking up on native JS properties.

---


*Comments from [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/548)*

jrobble:
Reviewed 3 of 22 files at r2, 5 of 5 files at r3.

---



*Comments from [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/548)*

brosenberg42:
Reviewed 5 of 25 files at r1, 17 of 22 files at r2, 5 of 5 files at r3.
Review status: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/mvc/controller/AdminPropertySettingsController.java, line 120 at r1](https://reviewable.io/reviews/openmpf/openmpf/548#-LBM9Bwv2Qp0-bC3Xx9j:-LBM9Bwv2Qp0-bC3Xx9k:br8xt54) ([raw file](https://github.com/openmpf/openmpf/blob/3e7bf83146ccc51c223ee514073369a54f09f023/trunk/workflow-manager/src/main/java/org/mitre/mpf/mvc/controller/AdminPropertySettingsController.java#L120)):*
> ```Java
>         // Get an updated list of property models. Each element contains current value. Check the flags to see if WFM restart is required to apply any change.
>         Optional<PropertyModel> propertyModel = propertiesUtil.getCustomProperties().stream().filter(pm -> pm.getNeedsRestart()).findFirst();
>         return propertyModel.isPresent();
> ```

This should be written as:
```
return propertiesUtil.getCustomProperties().stream()
        .anyMatch(PropertyModel::getNeedsRestart);
```

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/mvc/controller/AdminPropertySettingsController.java, line 145 at r3](https://reviewable.io/reviews/openmpf/openmpf/548#-LBWmWfx29uEBucMLAwj:-LBWmWfx29uEBucMLAwk:b-6n8b57) ([raw file](https://github.com/openmpf/openmpf/blob/75241400e37b40483b8ce0bd3521eee13aeeacc6/trunk/workflow-manager/src/main/java/org/mitre/mpf/mvc/controller/AdminPropertySettingsController.java#L145)):*
> ```Java
>         // Get an updated list of property models. Adjust the returned list of PropertyModels so they will indicate
>         // whether or not a WFM restart is required to apply a change.
>         List<PropertyModel> updatedPropertyModels = propertiesUtil.getCustomProperties().stream().filter(updatedPM -> {
> ```

This should be written as:
```
Set<String> savedProperties = propertyModels.stream()
        .map(PropertyModel::getKey)
        .collect(toSet());
        
List<PropertyModel> updatedPropertyModels = propertiesUtil.getCustomProperties().stream()
        .filter(pm -> savedProperties.contains(pm.getKey()))
        .collect(toList());
```

The current implementation is O(n^2), because you loop through propertyModels for every element in the stream.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/PropertiesUtil.java, line 172 at r3](https://reviewable.io/reviews/openmpf/openmpf/548#-LBWsD6i5aIJxMYPhcqQ:-LBWsD6i5aIJxMYPhcqR:b-kdjfud) ([raw file](https://github.com/openmpf/openmpf/blob/75241400e37b40483b8ce0bd3521eee13aeeacc6/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/PropertiesUtil.java#L172)):*
> ```Java
>         // Get an updated list of property models. Each element contains current value. Return only the immutable system properties by
>         // filtering out the mutable detection properties from the list.
>         return getCustomProperties().stream().filter(pm -> !isDetectionProperty(pm.getKey())).collect(toList());
> ```

Each method call that accepts a lambda parameter should be written on its own line. That is the standard style for Java. It is also the style used in all the documentation, such as the examples from the comments here: https://docs.oracle.com/javase/8/docs/api/java/util/stream/package-summary.html

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/PropertiesUtil.java, line 191 at r3](https://reviewable.io/reviews/openmpf/openmpf/548#-LBXFNs98bp1eSy74Vbz:-LBXFNs98bp1eSy74Vc-:bjmjip6) ([raw file](https://github.com/openmpf/openmpf/blob/75241400e37b40483b8ce0bd3521eee13aeeacc6/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/PropertiesUtil.java#L191)):*
> ```Java
> 
>     public TransientDetectionSystemProperties createDetectionSystemPropertiesSnapshot() {
>         Map detMap = new HashMap<String, String>();
> ```

This should be: `Map<String, String> detMap = new HashMap<>()`

---

*[trunk/workflow-manager/src/main/webapp/resources/mytheme/css/styles.css, line 295 at r1](https://reviewable.io/reviews/openmpf/openmpf/548#-LBM8i3a0diGA5ljCjFx:-LBM8i3a0diGA5ljCjFy:blq9pph) ([raw file](https://github.com/openmpf/openmpf/blob/3e7bf83146ccc51c223ee514073369a54f09f023/trunk/workflow-manager/src/main/webapp/resources/mytheme/css/styles.css#L295)):*
> ```CSS
> }
> 
> .labelNeedsServerRestart {: width:;
> ```

Please remove the `: width:;`

---

*[trunk/workflow-manager/src/main/webapp/resources/mytheme/js/controllers/admin/AdminPropertySettingsCtrl.js, line 67 at r3](https://reviewable.io/reviews/openmpf/openmpf/548#-LBX8pF188fbpXTjYOUk:-LBX8pF188fbpXTjYOUl:b-uekfa6) ([raw file](https://github.com/openmpf/openmpf/blob/75241400e37b40483b8ce0bd3521eee13aeeacc6/trunk/workflow-manager/src/main/webapp/resources/mytheme/js/controllers/admin/AdminPropertySettingsCtrl.js#L67)):*
> ```JavaScript
>             return {
> 
>                 serverNeedsRestart: function () {
> ```

I think this is easier to understand: `return _.some(serverProperties, 'needsRestart');`. It takes care of the hasOwnProperty thing.


If you disagree, at least change it to get rid of the count

---

*[trunk/workflow-manager/src/main/webapp/resources/mytheme/js/controllers/admin/AdminPropertySettingsCtrl.js, line 69 at r3](https://reviewable.io/reviews/openmpf/openmpf/548#-LBXG3A85O52ZSE8s0Rp:-LBXG3A85O52ZSE8s0Rq:ba3jo44) ([raw file](https://github.com/openmpf/openmpf/blob/75241400e37b40483b8ce0bd3521eee13aeeacc6/trunk/workflow-manager/src/main/webapp/resources/mytheme/js/controllers/admin/AdminPropertySettingsCtrl.js#L69)):*
> ```JavaScript
>                 serverNeedsRestart: function () {
>                     var immutablePropertiesChangedCount = 0;
>                     if (serverProperties) {
> ```

What are checking for here? {} evaluates to true.

---


*Comments from [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/548)*

jrobble:
Review status: 22 of 26 files reviewed at latest revision, 7 unresolved discussions.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/mvc/controller/AdminPropertySettingsController.java, line 120 at r1](https://reviewable.io/reviews/openmpf/openmpf/548#-LBM9Bwv2Qp0-bC3Xx9j:-LBbCnrO-_CuSngJNrF1:bb6un1j) ([raw file](https://github.com/openmpf/openmpf/blob/3e7bf83146ccc51c223ee514073369a54f09f023/trunk/workflow-manager/src/main/java/org/mitre/mpf/mvc/controller/AdminPropertySettingsController.java#L120)):*

This should be written as:
```
return propertiesUtil.getCustomProperties().stream()
        .anyMatch(PropertyModel::getNeedsRestart);
```
</blockquote></details>

Done

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/mvc/controller/AdminPropertySettingsController.java, line 145 at r3](https://reviewable.io/reviews/openmpf/openmpf/548#-LBWmWfx29uEBucMLAwj:-LBbE1WK0xIbObh6hvnS:bb6un1j) ([raw file](https://github.com/openmpf/openmpf/blob/75241400e37b40483b8ce0bd3521eee13aeeacc6/trunk/workflow-manager/src/main/java/org/mitre/mpf/mvc/controller/AdminPropertySettingsController.java#L145)):*

This should be written as:
```
Set<String> savedProperties = propertyModels.stream()
        .map(PropertyModel::getKey)
        .collect(toSet());
        
List<PropertyModel> updatedPropertyModels = propertiesUtil.getCustomProperties().stream()
        .filter(pm -> savedProperties.contains(pm.getKey()))
        .collect(toList());
```

The current implementation is O(n^2), because you loop through propertyModels for every element in the stream.
</blockquote></details>

Done

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/PropertiesUtil.java, line 172 at r3](https://reviewable.io/reviews/openmpf/openmpf/548#-LBWsD6i5aIJxMYPhcqQ:-LBbFMltA7VsnTzM5TFv:b-8290po) ([raw file](https://github.com/openmpf/openmpf/blob/75241400e37b40483b8ce0bd3521eee13aeeacc6/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/PropertiesUtil.java#L172)):*

Each method call that accepts a lambda parameter should be written on its own line. That is the standard style for Java. It is also the style used in all the documentation, such as the examples from the comments here: https://docs.oracle.com/javase/8/docs/api/java/util/stream/package-summary.html
</blockquote></details>

Good advice. Done.

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/PropertiesUtil.java, line 191 at r3](https://reviewable.io/reviews/openmpf/openmpf/548#-LBXFNs98bp1eSy74Vbz:-LBbF_cK143-qIirUZvr:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/75241400e37b40483b8ce0bd3521eee13aeeacc6/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/PropertiesUtil.java#L191)):*

This should be: `Map<String, String> detMap = new HashMap<>()`
</blockquote></details>

Done.

---

*[trunk/workflow-manager/src/main/webapp/resources/mytheme/css/styles.css, line 295 at r1](https://reviewable.io/reviews/openmpf/openmpf/548#-LBM8i3a0diGA5ljCjFx:-LBbGKuE2OczOTi_rSdl:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/3e7bf83146ccc51c223ee514073369a54f09f023/trunk/workflow-manager/src/main/webapp/resources/mytheme/css/styles.css#L295)):*

Please remove the `: width:;`
</blockquote></details>

Done.

---

*[trunk/workflow-manager/src/main/webapp/resources/mytheme/js/controllers/admin/AdminPropertySettingsCtrl.js, line 67 at r3](https://reviewable.io/reviews/openmpf/openmpf/548#-LBX8pF188fbpXTjYOUk:-LBbJtCb6yvUOYAbh1Nl:bp7ripc) ([raw file](https://github.com/openmpf/openmpf/blob/75241400e37b40483b8ce0bd3521eee13aeeacc6/trunk/workflow-manager/src/main/webapp/resources/mytheme/js/controllers/admin/AdminPropertySettingsCtrl.js#L67)):*

I think this is easier to understand: `return _.some(serverProperties, 'needsRestart');`. It takes care of the hasOwnProperty thing.


If you disagree, at least change it to get rid of the count
</blockquote></details>

Much simpler. Yeah, _.some() uses Object.keys(), which does not pull in properties in the prototype chain. Done.

---

*[trunk/workflow-manager/src/main/webapp/resources/mytheme/js/controllers/admin/AdminPropertySettingsCtrl.js, line 69 at r3](https://reviewable.io/reviews/openmpf/openmpf/548#-LBXG3A85O52ZSE8s0Rp:-LBbJpFr-CPl10W3JPqj:bwv3zhe) ([raw file](https://github.com/openmpf/openmpf/blob/75241400e37b40483b8ce0bd3521eee13aeeacc6/trunk/workflow-manager/src/main/webapp/resources/mytheme/js/controllers/admin/AdminPropertySettingsCtrl.js#L69)):*

What are checking for here? {} evaluates to true.
</blockquote></details>

This is a remnant. Addressed.

---


*Comments from [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/548)*

PR #555, opened 2018-05-08
Fixed streaming timestamp so hour is in 0-23 range, not 1-24.

jrobble:


jrobble:
Changed from `kk` to `HH`:

```
   k       clock-hour-of-am-pm (1-24)  number            0
   H       hour-of-day (0-23)          number            0
```

Tested via:

```
    public static void main(String args[]) {
        Instant instant = Instant.ofEpochMilli(1525738354000L); // Tue, 08 May 2018 00:12:34 GMT
        LocalDateTime date = instant.atZone(ZoneId.of("GMT")).toLocalDateTime();
        System.out.println(getLocalDateTimeAsString(date));
    }
```

PR #600, opened 2018-06-20
Modify JobController so that it doesn't say a job does not exist just because it is not in user's session.

brosenberg42:
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/600)

### DIFF
--- a/trunk/workflow-manager/src/main/java/org/mitre/mpf/mvc/controller/JobController.java
+++ b/trunk/workflow-manager/src/main/java/org/mitre/mpf/mvc/controller/JobController.java
@@ -26,37 +26,13 @@
 
 package org.mitre.mpf.mvc.controller;
 
-import static java.util.stream.Collectors.joining;
-
-import io.swagger.annotations.Api;
-import io.swagger.annotations.ApiOperation;
-import io.swagger.annotations.ApiParam;
-import io.swagger.annotations.ApiResponse;
-import io.swagger.annotations.ApiResponses;
-import java.io.File;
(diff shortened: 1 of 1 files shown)


jrobble:
Reviewed 1 of 1 files at r1.
Review status: all files reviewed, 1 unresolved discussion (waiting on @brosenberg42)

---

*[trunk/workflow-manager/src/main/java/org/mitre/mpf/mvc/controller/JobController.java, line 226 at r1](https://reviewable.io/reviews/openmpf/openmpf/600#-LFVYzQmDPcK4Nyrrm22:-LFVYzQmDPcK4Nyrrm23:bkgejq7) ([raw file](https://github.com/openmpf/openmpf/blob/b537ac924adce9c87385fd34e281a20d9be35f62/trunk/workflow-manager/src/main/java/org/mitre/mpf/mvc/controller/JobController.java#L226)):*
> ```Java
>         if (jobRequest == null) {
>             log.error("getJobStatusRest: Error retrieving the SingleJobInfo model for the job with id '{}'", jobId);
>             return new ResponseEntity<>(HttpStatus.BAD_REQUEST);
> ```

While I'm not going to ask you to update all the ResponseEntity creation lines to use the static methods, I will ask that we at least be consistent within each method. Thus, please use`ResponseEntity.badRequest()` here.

Accepting PR as is. Don't need to review again for this.

---


*Comments from [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/600)*

Issue #642, opened 2018-07-25
Allow REDIS_HOST to be set to something other than localhost


jrobble:
The /etc/redis.conf file on Jenkins (and development VMs) have the following line:

`bind 127.0.0.1`

That means Redis will only accept connections on the loopback interface. In order to accept incoming connections from other hosts on the network, that line will need to be updated with the hostname of the master node.

This will be more important when we start using Redis to store and retrieve frame data across an OpenMPF cluster.

Issue #641, opened 2018-07-25
Allow MYSQL_HOST to be set to something other than localhost


jrobble:
In order to establish a JDBC connection to a mySQL instance using a hostname other than localhost, entries must be added to the mysql.user table. On  development VM, the table looks as follows:

```
mysql> select User,Host from user;
+------+-----------------------+
| User | Host                  |
+------+-----------------------+
| mpf  | %                     |
| root | 127.0.0.1             |
| root | ::1                   |
|      | localhost             |
| mpf  | localhost             |
| root | localhost             |
|      | localhost.localdomain |
| mpf  | localhost.localdomain |
| root | localhost.localdomain |
+------+-----------------------+
```

New entries must be added for other hostnames, one for each user.

Note that Docker deployments of OpenMPF have a way around this.

clnowacki:
Things are different in Postgres. Hasn't been a problem.

PR #643, opened 2018-07-25
Setting MYSQL_HOST and REDIS_HOST to localhost.

jrobble:
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/643)


jrobble:
I added these wishlist issues to address later:
- https://github.com/openmpf/openmpf/issues/641
- https://github.com/openmpf/openmpf/issues/642

PR #710, opened 2018-10-23
Locate Test Resources for Integration Tests

bquinn17:
Currently system tests fail with a distributed docker deployment of Openmpf, due to the fact that the test resource files (image and video samples) are only located on the workflow manager container. This PR will allow the tests that run on the node manager container to look for the test resources in the shared folder if they are not found. (#711)

---
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/710)


jrobble:
Closing in favor of https://github.com/openmpf/openmpf/pull/879.

PR #818, opened 2019-01-24
Fix excessive Redis use

brosenberg42:
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/818)


jrobble:
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/transients/TransientMedia.java, line 37 at r2](https://reviewable.io/reviews/openmpf/openmpf/818#-LX_p8R27b8UU40zqSoo:-LX_p8R27b8UU40zqSop:b-cxdp2) ([raw file](https://github.com/openmpf/openmpf/blob/536d2403f4dde32dfa262952cc6f23690488b889/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/transients/TransientMedia.java#L37)):*
> ```Java
> 
> /** An in-flight media instance. */
> public interface TransientMedia {
> ```

Why use an interface? Do you anticipate using a different implementation for a different purpose? Same question goes for TransientJob and TransientStreamingJob. The other transient classes don't have interfaces.

jrobble:
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/transients/TransientMedia.java, line 37 at r2](https://reviewable.io/reviews/openmpf/openmpf/818#-LX_p8R27b8UU40zqSoo:-LX_r1EmEV6UuGznrT3-:b-le2nqk) ([raw file](https://github.com/openmpf/openmpf/blob/536d2403f4dde32dfa262952cc6f23690488b889/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/transients/TransientMedia.java#L37)):*

Why use an interface? Do you anticipate using a different implementation for a different purpose? Same question goes for TransientJob and TransientStreamingJob. The other transient classes don't have interfaces.
</blockquote></details>

Does it have to do with making those immutable unless casted?

brosenberg42:
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/RedisImpl.java, line 146 at r2](https://reviewable.io/reviews/openmpf/openmpf/818#-LXeSBVoEw64nPOI4Kvd:-LXek3cf0JmCFUPZVLpg:b-7rp2i7) ([raw file](https://github.com/openmpf/openmpf/blob/536d2403f4dde32dfa262952cc6f23690488b889/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/RedisImpl.java#L146)):*

I think you mentioned this before, but I'll ask again. There's no way to just specify that you want to delete everything that starts with `key(BATCH_JOB, job.getId(), MEDIA, media.getId(), TRACK)`?
</blockquote></details>

The closest thing you can do is use a command like `KEYS BATCHJOB:123:*` to get a list of keys and then delete them. This seems wasteful since we already can figure out that list of keys without asking Redis. Also, the documentation for the KEYS command ( https://redis.io/commands/keys ) says you shouldn't use it in production.

jrobble:
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/jobcreation/JobCreationProcessor.java, line 142 at r5](https://reviewable.io/reviews/openmpf/openmpf/818#-LYEHt2R7c-k011Y6zF-:-LYS-GlSCu3PK9VVaTqP:b-704g4r) ([raw file](https://github.com/openmpf/openmpf/blob/37a65c20efe09bbd57a2aff7545ecbe80f676b99/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/jobcreation/JobCreationProcessor.java#L142)):*

Currently, if a job has some valid and some invalid media, we still run detection on the valid pieces of media. With that behavior, the job status is supposed to be `COMPLETE_WITH_ERRORS` since we ran detection. If we want to make the end status be `ERROR`, then we need to change the code so that we don't run detection at all when at least one piece of media has an error. Do you want me to change the code so that we don't run detection on valid media when there is at least one piece of media with an error?
</blockquote></details>

Let's ask the users what their preference is.

PR #845, opened 2019-02-14
Add S3 support

brosenberg42:
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/845)


brosenberg42:
> Also, please fix up the "Source Path" when the URL is "file:/" so that it's "file:///".

That was because `File.toURI()` uses `file:/`. `Path.toUri()` uses `file:///`

PR #869, opened 2019-03-08
Added additional artifact extraction properties

clnowacki:
- Made the definition of non-visual types a set defined in mpf.properties instead of hardcoded.
- changed the Artifact Extraction Policy choices to NONE, VISUAL_TYPES_ONLY, ALL_TYPES, and ALL_FRAMES.
- added a set of properties that control which frames get extracted and how many, when the artifact extraction policy is either VISUAL_TYPES_ONLY or ALL_TYPES.

---
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/869)


brosenberg42:
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/artifactextraction/ArtifactExtractionSplitterImpl.java, line 225 at r1](https://reviewable.io/reviews/openmpf/openmpf/869#-L_x0JMm2G9wDRCBAZNw:-L_x0JMm2G9wDRCBAZNx:b-fz47dh) ([raw file](https://github.com/openmpf/openmpf/blob/c9ad4c66280d17b6257e8a31747ea0e7acbc7ea4/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/artifactextraction/ArtifactExtractionSplitterImpl.java#L225)):*
> ```Java
>                     }
>                     if (_propertiesUtil.getArtifactExtractionPolicyTopConfidenceCount() > 0) {
>                         List<Detection> detections_copy = new ArrayList(track.getDetections());
> ```

The right hand side of the assignment is missing `<>` so it should be `new ArrayList<>(track.getDetections());`. When you don't have the `<>` you are essentially creating a non-typed ArrayList. The existing code works, but generates compiler warning.

jrobble:
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/artifactextraction/ArtifactExtractionSplitterImpl.java, line 214 at r1](https://reviewable.io/reviews/openmpf/openmpf/869#-L_nyZ3yCFp4x1X9sRrY:-L_nyZ3yCFp4x1X9sRrZ:b-77tbgv) ([raw file](https://github.com/openmpf/openmpf/blob/c9ad4c66280d17b6257e8a31747ea0e7acbc7ea4/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/artifactextraction/ArtifactExtractionSplitterImpl.java#L214)):*
> ```Java
>                         int smallest_distance = Integer.MAX_VALUE;
>                         int middle_frame = 0;
>                         for (Detection detection : track.getDetections()) {
> ```

This approach is O(n), but since the detections are pre-sorted (SortedSet) we should be able to get this down to O(log n) with a binary search. You should be able to use `Collections.binarySearch()`[(see here)](https://www.geeksforgeeks.org/collections-binarysearch-java-examples/) by passing a custom Comparator as the second param. If for some reason that doesn't work, there is [this](https://stackoverflow.com/questions/30245166/find-the-nearest-closest-value-in-a-sorted-list?answertab=votes#tab-top).

jrobble:
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/artifactextraction/ArtifactExtractionSplitterImpl.java, line 214 at r1](https://reviewable.io/reviews/openmpf/openmpf/869#-L_nyZ3yCFp4x1X9sRrY:-LaWnUTr9V9-r6VpjBkb:bq43zlq) ([raw file](https://github.com/openmpf/openmpf/blob/c9ad4c66280d17b6257e8a31747ea0e7acbc7ea4/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/artifactextraction/ArtifactExtractionSplitterImpl.java#L214)):*
> I'd need to extract the set of frame indices from the track,

I think you can implement the Comparator to use `Detection.getMediaOffsetFrame()`and have it operate on the set of Detections.

> I still need to check the detections before and after the would-be insertion point to decide which is closer. 

Yeah, that does seem to be the case. Maybe using [this approach](https://stackoverflow.com/questions/30245166/find-the-nearest-closest-value-in-a-sorted-list?answertab=votes#tab-top) is more straight-forward.

jrobble:
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/artifactextraction/ArtifactExtractionSplitterImpl.java, line 214 at r1](https://reviewable.io/reviews/openmpf/openmpf/869#-L_nyZ3yCFp4x1X9sRrY:-LaWvWXY3rAO4GvjhNUN:b28uvrg) ([raw file](https://github.com/openmpf/openmpf/blob/c9ad4c66280d17b6257e8a31747ea0e7acbc7ea4/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/artifactextraction/ArtifactExtractionSplitterImpl.java#L214)):*
<details><summary><i>Previously, clnowacki wrote…</i></summary><blockquote>

Sure, I could create a Comparator, but the binarySearch wants a set of objects, plus an instance of an object to find. So I can give it the set of Detections to search, but I don't have a Detection instance for it to look for, even if I give it a comparator that looks at the mediaOffsetFrame. I guess I could give it a dummy Detection with the frame index set and everything else set to 0. As long as I'm only comparing the frame index maybe that would work.
</blockquote></details>

Yeah, maybe something like:

```
int index = Collections.binarySearch(track.getDetections(), new Detection(0, 0, 0, 0, 0, middle_frame, 0, Collections.emptyMap()), Comparator.comparing(Detections::getMediaOffsetFrame));
```

jrobble:
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/artifactextraction/ArtifactExtractionSplitterImpl.java, line 214 at r1](https://reviewable.io/reviews/openmpf/openmpf/869#-L_nyZ3yCFp4x1X9sRrY:-LaWxqSN61pfK11DfyHr:bhnwm7k) ([raw file](https://github.com/openmpf/openmpf/blob/c9ad4c66280d17b6257e8a31747ea0e7acbc7ea4/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/artifactextraction/ArtifactExtractionSplitterImpl.java#L214)):*

Yeah, maybe something like:

```
int index = Collections.binarySearch(track.getDetections(), new Detection(0, 0, 0, 0, 0, middle_frame, 0, Collections.emptyMap()), Comparator.comparing(Detections::getMediaOffsetFrame));
```

</blockquote></details>

[Alternatively](https://stackoverflow.com/a/36739594), this seems cleaner:

```
Detection middleDetection = track.getDetections().stream().min((d1,d2) -> Math.abs(d1.getMediaOffsetFrame() - middleFrame) - Math.abs(d2.getMediaOffsetFrame() - middleFrame)));
```

Also, please use camelCase for variable names here.

jrobble:
___
*[trunk/workflow-manager/src/main/resources/workflow-properties.json, line 24 at r11](https://reviewable.io/reviews/openmpf/openmpf/869#-LrFj2YYEiDrreuF3n9s:-LrFj2YZ1vZEdjUMxGZd:b-lqjp87) ([raw file](https://github.com/openmpf/openmpf/blob/aa1735be04d4f71e495646bff9333da8d41f9b60/trunk/workflow-manager/src/main/resources/workflow-properties.json#L24)):*
> ```JSON
>     },
>     {
>         "name": "ARTIFACT_EXTRACTION_POLICY",
> ```

Move this property directly above ARTIFACT_EXTRACTION_POLICY_EXEMPLAR_FRAME_PLUS.

Reword "will be extracted according to the ARTIFACTION_EXTRACTION_POLICY* properties below." to:

"will be extracted according to the ARTIFACT_EXTRACTION_POLICY* properties."

Notice that I corrected the spelling of "ARTIFACTION".

jrobble:
___
*[trunk/workflow-manager/src/main/resources/workflow-properties.json, line 46 at r11](https://reviewable.io/reviews/openmpf/openmpf/869#-LrVOXf56AhRO_LhR0L6:-LrVOXf56AhRO_LhR0L7:b-d668t0) ([raw file](https://github.com/openmpf/openmpf/blob/aa1735be04d4f71e495646bff9333da8d41f9b60/trunk/workflow-manager/src/main/resources/workflow-properties.json#L46)):*
> ```JSON
>     {
>         "name": "ARTIFACT_EXTRACTION_POLICY_EXEMPLAR_FRAME_PLUS",
>         "description": "Extract the exemplar frame from the track, plus N frames before and after the exemplar. If N = 0, then only the exemplar frame will be extracted. If N > 0, then the exemplar frame plus N frames before and after it will be extracted, if they exist. If N < 0, then this property is disabled.",
> ```

To clarify, please state:

"plus N frames before and after it will be extracted, if they exist in the track".

I want to make it clear that all frames extracted are part of the track.

jrobble:
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/PropertiesUtil.java, line 399 at r12](https://reviewable.io/reviews/openmpf/openmpf/869#-Lru-ReW52ML0ElPFtQ_:-Lru-ReW52ML0ElPFtQa:bjehlbl) ([raw file](https://github.com/openmpf/openmpf/blob/b6156bdcc0d5ba1c11ce11bfac6ffa7488319934/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/PropertiesUtil.java#L399)):*
> ```Java
>     }
> 
>     public Set<String> getArtifactExtractionNonvisualTypesList() {
> ```

Capitalize "V" in "Nonvisual".

jrobble:
___
*[trunk/workflow-manager/src/main/resources/properties/mpf.properties, line 33 at r15](https://reviewable.io/reviews/openmpf/openmpf/869#-Lv7nWRPCaviEI_QI8ip:-Lv7nWRPCaviEI_QI8iq:b-r9x95m) ([raw file](https://github.com/openmpf/openmpf/blob/db5dbb6acebb9a72868d55bfd7854871e3361777/trunk/workflow-manager/src/main/resources/properties/mpf.properties#L33)):*
> ```INI
> output.site.name=mpf1
> amq.broker.uri=${env:ACTIVE_MQ_BROKER_URI}
> mpf.output.objects.enabled=true
> ```

I've been looking at how `mpf.output.objects.enabled` is used because I was trying to test it in conjunction with the other `mpf.output.objects.*` properties. In general, it's used in two places:

**1. EndOfStageProcessor checks `mpf.output.objects.enabled` here:**

```
			//notify of completion - use
			if(!job.isOutputEnabled()) {
				jobStatusBroadcaster.broadcast(
						jobId,
						100,
						job.isCancelled() ? BatchJobStatusType.CANCELLED : BatchJobStatusType.COMPLETE,
						new Date());
				jobProgressStore.setJobProgress(jobId, 100.0f);
			} else {
				jobStatusBroadcaster.broadcast(jobId, 99, BatchJobStatusType.BUILDING_OUTPUT_OBJECT, new Date());
				jobProgressStore.setJobProgress(jobId, 99.0f);
			}	
```

This doesn't really make sense because the output object is always created in JobCompleteProcessorImpl:

```
			try {
				markJobStatus(jobId, BatchJobStatusType.BUILDING_OUTPUT_OBJECT);

				// NOTE: jobStatus is mutable - it __may__ be modified in createOutputObject!
				createOutputObject(jobId, jobStatus);
```

Looking back, it seems that the last time `mpf.output.objects.enabled` was properly used for batch jobs was in R2.1.0 in OutputObjectSenderProcessor:

```
			if (jsonJobRequest.isOutputObjectEnabled()) {
				JsonOutputObjectSummary jsonOutputObjectSummary = new JsonOutputObjectSummary(jobRequest.getOutputObjectPath());
				exchange.getOut().setHeader(MpfHeaders.SEND_OUTPUT_OBJECT, Boolean.TRUE);
				exchange.getOut().setHeader(MpfHeaders.JOB_ID, jobId);
				exchange.getOut().setBody(jsonUtils.serializeAsText(jsonOutputObjectSummary));
			}
```

Which resulted in placing a message in the old MPF_OUTPUT queue. From JobCompletedRouteBuilder:

```
			.choice()
				.when(header(MpfHeaders.SEND_OUTPUT_OBJECT).isEqualTo(Boolean.TRUE))
					.to(String.format("output:queue:%s", propertiesUtil.getOutputQueueName()), exitPoint)
```

As of R3.0.0, we stopped using that queue. I'm wondering if we should just remove support for `mpf.output.objects.enabled` for batch jobs and only keep it for streaming jobs. 

**2. `mpf.output.objects.enabled` is used to determine if streaming job summary reports should be written to disk.**

In StreamingJobController:

```
                boolean enableOutputToDisk = propertiesUtil.isOutputObjectsEnabled();
                if ( streamingJobCreationRequest.getEnableOutputToDisk() != null ) {
                  enableOutputToDisk = streamingJobCreationRequest.getEnableOutputToDisk();
                }
```

**Thoughts**

In our next meeting let's talk about renaming `mpf.output.objects.enabled` to `streaming.output.objects.to.disk.enabled` and remove all batch support. Currently, the only way to get batch output is by reading the output object JSON. On the other hand, we can return streaming output via callbacks, so writing the JSON to disk is not strictly necessary.

Currently, I think `mpf.output.objects.enabled` will confuse users.

jrobble:
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/artifactextraction/ArtifactExtractionSplitterImpl.java, line 283 at r15](https://reviewable.io/reviews/openmpf/openmpf/869#-Lv7Mlsn0FFTpmEikZiJ:-LvBKNrq1EWVB09JVAV8:b-iup8qb) ([raw file](https://github.com/openmpf/openmpf/blob/db5dbb6acebb9a72868d55bfd7854871e3361777/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/artifactextraction/ArtifactExtractionSplitterImpl.java#L283)):*
<details><summary><i>Previously, clnowacki wrote…</i></summary><blockquote>

All we care about here is whether the sign has changed so this is unnecessary.
</blockquote></details>

We also care about whether we are looking at the exact middle index (the case where distance == 0). Consider that for some reason distance is set to -0.0000001 due to floating point precision inaccuracy. In that case we'd go through another iteration of the loop unnecessarily.

jrobble:
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/artifactextraction/ArtifactExtractionSplitterImpl.java, line 283 at r15](https://reviewable.io/reviews/openmpf/openmpf/869#-Lv7Mlsn0FFTpmEikZiJ:-LvBMhoSAS1fS8cW9ufd:b-hpicxq) ([raw file](https://github.com/openmpf/openmpf/blob/db5dbb6acebb9a72868d55bfd7854871e3361777/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/artifactextraction/ArtifactExtractionSplitterImpl.java#L283)):*

We also care about whether we are looking at the exact middle index (the case where distance == 0). Consider that for some reason distance is set to -0.0000001 due to floating point precision inaccuracy. In that case we'd go through another iteration of the loop unnecessarily.
</blockquote></details>

Thinking about this more, if we can actually find the exact middle frame then we must be dealing with whole number distances, so, in theory, there should be no floating point precision inaccuracy. I would hope. I'm not 100% sure of this.

jrobble:
___
*[trunk/workflow-manager/src/main/resources/properties/mpf.properties, line 33 at r15](https://reviewable.io/reviews/openmpf/openmpf/869#-Lv7nWRPCaviEI_QI8ip:-LvC0fYt8CdOyuBckMWA:b-nwhq4g) ([raw file](https://github.com/openmpf/openmpf/blob/db5dbb6acebb9a72868d55bfd7854871e3361777/trunk/workflow-manager/src/main/resources/properties/mpf.properties#L33)):*

I've been looking at how `mpf.output.objects.enabled` is used because I was trying to test it in conjunction with the other `mpf.output.objects.*` properties. In general, it's used in two places:

**1. EndOfStageProcessor checks `mpf.output.objects.enabled` here:**

```
			//notify of completion - use
			if(!job.isOutputEnabled()) {
				jobStatusBroadcaster.broadcast(
						jobId,
						100,
						job.isCancelled() ? BatchJobStatusType.CANCELLED : BatchJobStatusType.COMPLETE,
						new Date());
				jobProgressStore.setJobProgress(jobId, 100.0f);
			} else {
				jobStatusBroadcaster.broadcast(jobId, 99, BatchJobStatusType.BUILDING_OUTPUT_OBJECT, new Date());
				jobProgressStore.setJobProgress(jobId, 99.0f);
			}	
```

This doesn't really make sense because the output object is always created in JobCompleteProcessorImpl:

```
			try {
				markJobStatus(jobId, BatchJobStatusType.BUILDING_OUTPUT_OBJECT);

				// NOTE: jobStatus is mutable - it __may__ be modified in createOutputObject!
				createOutputObject(jobId, jobStatus);
```

Looking back, it seems that the last time `mpf.output.objects.enabled` was properly used for batch jobs was in R2.1.0 in OutputObjectSenderProcessor:

```
			if (jsonJobRequest.isOutputObjectEnabled()) {
				JsonOutputObjectSummary jsonOutputObjectSummary = new JsonOutputObjectSummary(jobRequest.getOutputObjectPath());
				exchange.getOut().setHeader(MpfHeaders.SEND_OUTPUT_OBJECT, Boolean.TRUE);
				exchange.getOut().setHeader(MpfHeaders.JOB_ID, jobId);
				exchange.getOut().setBody(jsonUtils.serializeAsText(jsonOutputObjectSummary));
			}
```

Which resulted in placing a message in the old MPF_OUTPUT queue. From JobCompletedRouteBuilder:

```
			.choice()
				.when(header(MpfHeaders.SEND_OUTPUT_OBJECT).isEqualTo(Boolean.TRUE))
					.to(String.format("output:queue:%s", propertiesUtil.getOutputQueueName()), exitPoint)
```

As of R3.0.0, we stopped using that queue. I'm wondering if we should just remove support for `mpf.output.objects.enabled` for batch jobs and only keep it for streaming jobs. 

**2. `mpf.output.objects.enabled` is used to determine if streaming job summary reports should be written to disk.**

In StreamingJobController:

```
                boolean enableOutputToDisk = propertiesUtil.isOutputObjectsEnabled();
                if ( streamingJobCreationRequest.getEnableOutputToDisk() != null ) {
                  enableOutputToDisk = streamingJobCreationRequest.getEnableOutputToDisk();
                }
```

**Thoughts**

In our next meeting let's talk about renaming `mpf.output.objects.enabled` to `streaming.output.objects.to.disk.enabled` and remove all batch support. Currently, the only way to get batch output is by reading the output object JSON. On the other hand, we can return streaming output via callbacks, so writing the JSON to disk is not strictly necessary.

Currently, I think `mpf.output.objects.enabled` will confuse users.
</blockquote></details>

Created this issue for a future PR: https://github.com/openmpf/openmpf-contrib-components/issues/54

PR #878, opened 2019-03-26
Unmanaged component registration

jrobble:
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/878)

### DIFF
--- a/trunk/detection/executor/cpp/batch/main.cpp
+++ b/trunk/detection/executor/cpp/batch/main.cpp
@@ -91,7 +91,6 @@ int main(int argc, char* argv[]) {
 
     log4cxx::xml::DOMConfigurator::configure(app_dir + "/../config/Log4cxxConfig.xml");
     log4cxx::LoggerPtr logger = log4cxx::Logger::getLogger("org.mitre.mpf.detection");
-    LOG4CXX_INFO(logger, "Created Logger");
 
     if (argc < 4) {
         LOG4CXX_ERROR(logger,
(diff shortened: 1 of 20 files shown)


brosenberg42:
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/service/component/AddComponentServiceImpl.java, line 180 at r1](https://reviewable.io/reviews/openmpf/openmpf/878#-LavcuoaE7C2bAaesrsh:-Lbcy0gX-Pze18Q5nBk8:b1dwz59) ([raw file](https://github.com/openmpf/openmpf/blob/a67782d7c1dfadb215b7592dff9143070c03f5c6/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/service/component/AddComponentServiceImpl.java#L180)):*

What do you think about adding this comment?
</blockquote></details>

I don't understand it. The descriptor path gets set for all components.

jrobble:
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/service/component/AddComponentServiceImpl.java, line 180 at r1](https://reviewable.io/reviews/openmpf/openmpf/878#-LavcuoaE7C2bAaesrsh:-Lbd2eba8hSDJnEWgiCb:b-vjfzrl) ([raw file](https://github.com/openmpf/openmpf/blob/a67782d7c1dfadb215b7592dff9143070c03f5c6/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/service/component/AddComponentServiceImpl.java#L180)):*

I don't understand it. The descriptor path gets set for all components.
</blockquote></details>

When I wrote this comment I didn't realize that. After looking at your tests it's pretty clear that the path is there even for unmanaged components.

More generally, when I first looked at this method it didn't seem clear why only a managed component would get here. Thinking about it more, only managed components are deployed.

Feel free to disregard this.

---

I ended up looking in `$MPF_HOME/plugins` on the WFM to verify the descriptor path for an unmanaged component. While I was there, I saw a `string` dir that got generated due to a failed attempt to use the REST endpoint with the placeholder descriptor JSON content that you get when you click the "Model Schema" text area. I don't think that dir should be there.

Issue #884, opened 2019-04-15
Transition from Oracle JDK to OpenJDK


jrobble:
Transition from Oracle JDK to OpenJDK. Update the Docker build to use OpenJDK instead of a user-downloaded RPM for the Oracle JDK.

PR #898, opened 2019-05-23
Java 11 upgrade

bquinn17:
Changes needed to upgrade to Java 11, Tomcat 8.5, and Mockito 2.26, Surefire 2.22.1, and new tomcat plugin

openmpf/openmpf#884
openmpf/openmpf#947

---
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/898)


bquinn17:
Maybe we should also try upgrading spark-core and spring-security-core in order to remove those security warnings.

bquinn17:
Maybe we should also try upgrading spark-core and spring-security-core in order to remove those security warnings.

jrobble:
> Maybe we should also try upgrading spark-core and spring-security-core in order to remove those security warnings.

I agree. I created https://github.com/openmpf/openmpf/issues/901 for later. It's not a high priority at the moment. If for some reason you do need to update Spring as part of the OpenJDK task, then we should also address that task.

jrobble:
I see you have a Docker Jenkins build running. Please let me know how it goes.

brosenberg42:
___
*[trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/ITComponentRegistration.java, line 218 at r7](https://reviewable.io/reviews/openmpf/openmpf/898#-LhCXcZoD95MclKtlGX8:-LhCZeCf2eyBk0vRoL-5:b-onb721) ([raw file](https://github.com/openmpf/openmpf/blob/ab8e2c773348cc4ec6ba1af66ccfdc1a96da32df/trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/ITComponentRegistration.java#L218)):*

Loggers should be declared once at the top of the class like in other .java files that use loggers.
</blockquote></details>

I recommended that he add this in order to debug an issue on Jenkins. I think we just want to remove it.

Issue #897, opened 2019-05-18
File indexing results in memory leak when files removed


jrobble:
In general, OpenMPF will attempt to index files located in `$MPF_HOME/share` as soon as the webapp is started by Tomcat. This is so that those files can be listed in a directory tree in the Create Job UI. The main problem is that once a file gets indexed it’s never removed from the cache, even if the file is manually deleted. That’s the memory leak. We never designed the file indexer to handle the case where files are removed.

Users sometimes manually clear the out the generated artifacts and output object JSON files. If they do that when the WFM is still running that has no impact on the cache. Note that the cache only exists in the Java heap; it’s not persisted when the WFM is restarted.
 
Note that when the user creates an OpenMPF job through the REST endpoint using an HTTP or HTTPS URL, the file is temporarily downloaded to the `$MPF_HOME/share/remote-media` directory and then deleted when the job is complete. If the file indexer is still running, it may detect the temporary files and add them to the cache before they are deleted.
 
Also, note that the file index cache is not impacted by the size of the files, only the number of files. If the users have a ton of artifacts and output objects due to the volume of jobs this may be a major issue.
 
Note that file indexing is not on any kind of timer. It’s kicked off when the webapp starts, but once the initial round of indexing is complete the only other time indexing takes place is if a user presses the refresh button on the Create Job UI. When a user uploads a file to a remote-media directory through the UI, an entry gets manually added to the index.

The long-term solution may be a combination of:
 
- Lazily indexing files on an as-needed basis
- Allowing the user to delete files through the web UI
- Setting the default file indexer path to something that doesn’t include the generated artifacts, markup, and output files
- Periodically re-indexing to check for deleted directories and files
- [Watch for changes](https://docs.oracle.com/javase/tutorial/essential/io/notification.html)

brosenberg42:
I think another possible solution is to only send one or two directory levels at a time to web ui. When a user expands a directory, we would at that time use AJAX to get the directory contents. We would have to see if that slows down the web ui enough to cause issues. If we do it this way, we don't need to do any indexing and we don't need to send the entire directory tree to the web ui.

jrobble:
> I think another possible solution is to only send one or two directory levels at a time to web ui. 

This is similar to what I was referring to with "Lazily indexing files on an as-needed basis". The need is when the user expands the directory or checks the box to include everything in the directory; however, I was thinking of only updating the index when a directory is used.

[...]

PR #929, opened 2019-07-03
File watching for media directory

bquinn17:
https://github.com/openmpf/openmpf/issues/897


jrobble:
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/service/FileWatcherServiceImpl.java, line 199 at r5](https://reviewable.io/reviews/openmpf/openmpf/929#-LmB7GT0Cs6sR5ykGN7P:-LmByr614pIvdge-7NUr:b-9qxe6p) ([raw file](https://github.com/openmpf/openmpf/blob/3447d0315eb3c234cfd117a78fc446427a5f231e/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/service/FileWatcherServiceImpl.java#L199)):*

Note to self: What if root is a symlink?
</blockquote></details>

The content of symlinked dirs is displayed in the Create Job page on startup (root dir or a subdir), but when I upload a file to a symlinked dir it's never displayed in the web UI. The file is actually uploaded though, as I can see it in the file browser. Refreshing the Create Job page does nothing. It's not until I restart the WFM that the file is displayed.

Adding a new dir to the symlinked dir result in similar behavior: The web UI is never updated, but I can see the dir in a file browser.

Also, when I delete a file in the symlinked dir it still appears in the UI even after refreshing the UI.

PR #960, opened 2019-08-26
Incomplete pipelines

brosenberg42:
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/960)


brosenberg42:
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/JsonUtils.java, line 57 at r4](https://reviewable.io/reviews/openmpf/openmpf/960#-LnsD40Z7x20R-ODGuF-:-LnssL2QAvYXZOAPXy73:b-j11itr) ([raw file](https://github.com/openmpf/openmpf/blob/d354aa77f1478d617d92a72f14f419fdefc2ce0b/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/JsonUtils.java#L57)):*

Use `_log` since you're using underscores in this class. In some classes we use `LOG`.
</blockquote></details>

I normally use all caps for static fields.

jrobble:
"Update delete buttons on Pipelines 2 page" - This comment is incorrect. That should be the Pipelines 1 page, but oh well, it will fall out in the squash.

brosenberg42:
___
*[trunk/workflow-manager/src/main/webapp/resources/mytheme/js/controllers/PipelinesCtrl.js, line 112 at r4](https://reviewable.io/reviews/openmpf/openmpf/960#-LnsAoMV1kjPRiXlmewM:-Lo6nId2ApBcBWNLWiYe:bto3ep2) ([raw file](https://github.com/openmpf/openmpf/blob/d354aa77f1478d617d92a72f14f419fdefc2ce0b/trunk/workflow-manager/src/main/webapp/resources/mytheme/js/controllers/PipelinesCtrl.js#L112)):*

Trying to save an action without specifying a name on the Pipelines 2 page generates an error in the web console log. I don't think that's the behavior we want.

I can generate a pipeline without a name because "CUSTOM PIPELINE" is automatically selected for me. This is inconsistent with trying to save an action without a name.

In both cases the user should not be allowed to generate a pipeline or action without specifying a name. The Pipelines 1 page will show an alert that says "missing fields" whenever a name or description is not provided.
</blockquote></details>

That total lack of input validation is a pre-existing issue.

jrobble:
___
*[trunk/workflow-manager/src/main/webapp/resources/mytheme/js/controllers/PipelinesCtrl.js, line 112 at r4](https://reviewable.io/reviews/openmpf/openmpf/960#-LnsAoMV1kjPRiXlmewM:-LoLMIH-9_TrwuAOIzmq:b-rr3g17) ([raw file](https://github.com/openmpf/openmpf/blob/d354aa77f1478d617d92a72f14f419fdefc2ce0b/trunk/workflow-manager/src/main/webapp/resources/mytheme/js/controllers/PipelinesCtrl.js#L112)):*

That total lack of input validation is a pre-existing issue.
</blockquote></details>

Got it. Is that an easy fix or something I should create a separate issue for? Are there other input validation issues you're aware of?

jrobble:
To be more precise, I'm more concerned about a single user seeing inconsistency in the UI while only using the UI while nothing is being changed via REST.

PR #983, opened 2019-10-04
Update nightly test expected output JSON for OpenCV 3.4.7 upgrade.

jrobble:
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/983)


jrobble:
___
*[trunk/interop/src/main/java/org/mitre/mpf/interop/util/TimeUtils.java, line 47 at r2](https://reviewable.io/reviews/openmpf/openmpf/983#-LqhN2QRCa1J7RvmZO_c:-LqhN2QRCa1J7RvmZO_d:bmerahl) ([raw file](https://github.com/openmpf/openmpf/blob/7d58646c5e9d9dac7eefbaaef61b37fa9c3a4d47/trunk/interop/src/main/java/org/mitre/mpf/interop/util/TimeUtils.java#L47)):*
> ```Java
>         return instant == null
>                 ? null
>                 : timestampFormatter.format(instant.truncatedTo(ChronoUnit.SECONDS));
> ```

Without this change, the completed time in the JSON output has sub-second precision, while the received time has second precision. This is due to how the latter is written to the DB and then read from the DB before being serialized to JSON. The completed time is no longer written to / read from the DB before being serialized to JSON.

jrobble:
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/service/StorageService.java, line 90 at r2](https://reviewable.io/reviews/openmpf/openmpf/983#-LqhNfOJ67dYXpSNqW1r:-LqhNfOK-pFZHZlMGqA6:b-tjrbsy) ([raw file](https://github.com/openmpf/openmpf/blob/7d58646c5e9d9dac7eefbaaef61b37fa9c3a4d47/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/service/StorageService.java#L90)):*
> ```Java
>             outputObject.getJobWarnings().add(
>                     "This output object was stored locally because storing it remotely failed due to: " + ex);
>             JobStatusCalculator.checkErrorMessages(outputObject, jobStatus);
> ```

I like this approach because it seems clean; however, a downside is that this warning message is never written to the DB as part of the job request. Similarly, if this method throws an error, we don't generate an error message in the catch block that gets written to the DB. In both cases, the final job status is written to the DB, which is what's most important.

Do you think it's important that the messages make it to the DB? I think we may have talked about this before and decided that's not important.

brosenberg42:
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/service/StorageService.java, line 90 at r2](https://reviewable.io/reviews/openmpf/openmpf/983#-LqhNfOJ67dYXpSNqW1r:-LqlSVDi8n9ChW-Lkxqt:b-iqr56k) ([raw file](https://github.com/openmpf/openmpf/blob/7d58646c5e9d9dac7eefbaaef61b37fa9c3a4d47/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/service/StorageService.java#L90)):*

Wouldn't this be the same as changing the end of `JobCompleteProcessorImpl` to
```
checkErrorMessages(jsonOutputObject, jobStatus);
var outputObjectUri = storageService.store(jsonOutputObject, jobStatus); // this may update the job status
// Check for storage issues.
checkErrorMessages(jsonOutputObject, jobStatus);
return outputObjectUri;
```
?
That way checkErrorMessages can stay private in `JobCompleteProcessorImpl` and the fact that we are checking the error messages twice is more obvious. Also, I think it is better if StorageService doesn't need to know what part of the job life cycle it is in. It just stores stuff and adds warnings if necessary, the whole process of what needs to happen at the end of job is better handled in JobCompleteProcessor. None of the other `StorageService.store*` call `JobStatusCalculator.checkErrorMessages` so it is inconsistent.

For warning message issue, I think this should have always done _inProgressJobs.addJobWarning(outputObject.getJobId(), "This output object was stored locally because storing it remotely failed due to: " + ex));
</blockquote></details>

Looking at this more closely don't worry about moving checkErrorMessages, but I think we will want to add the warning to _inProgressJobs.

Issue #1005, opened 2019-11-15
Attempting to get output JSON via REST before job is complete results in a NullPointerException


jrobble:
Attempting to get output JSON via REST before job is complete results in a NullPointerException.

For example, if job 32 is still in progress, this call:

http://localhost:8080/workflow-manager/jobs/output-object?id=32

Results in this NPE:

```
2019-11-15 14:23:10,703 ERROR [http-nio-8080-exec-4] o.m.m.m.ControllerUncaughtExceptionHandler - Request for http://localhost:8080/workflow-manager/jobs/32/output/detection raised an uncaught exception
java.lang.NullPointerException: null
	at java.net.URI$Parser.parse(URI.java:3106) ~[?:?]
	at java.net.URI.<init>(URI.java:600) ~[?:?]
	at java.net.URI.create(URI.java:881) ~[?:?]
	at org.mitre.mpf.mvc.controller.JobController.getSerializedDetectionOutputRest(JobController.java:284) ~[classes/:?]
	at jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method) ~[?:?]
```

The REST client gets this:

```
<html lang="en">
<head>
	<script>
	</script>
</head>
<!-- must not have an html body - will cause issues with the home view -->
<div id="errorResponseDiv">		
    <strong>Warning!</strong> There was a problem processing your request.  More information is available in your web server's log
    
    	<p class="error-message">An unknown error has occurred. Check the Workflow Manager log for details.</p>
    
    
</div>
</html>"; line: 9, column: 2]
```

PR #1006, opened 2019-11-15
Fix bug that causes a NullPointerException when attempting to get out…

jrobble:
…put JSON via REST before job is complete.

---
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1006)


jrobble:
Address https://github.com/openmpf/openmpf/issues/1005

PR #1036, opened 2020-02-05
Add final padding to track detections.

jrobble:
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1036)

### DIFF
--- a/trunk/ansible/install/ansible/roles/child/tasks/compdeploy.yml
+++ b/trunk/ansible/install/ansible/roles/child/tasks/compdeploy.yml
@@ -75,7 +75,7 @@
     - name: Get the component top level directory
       set_fact:
         component_tld: "{{ component_tld|default([])|union([{'tld': item}]) }}"
-      when: "{{ item | match(component_tld_regex) }}"
+      when: item is match(component_tld_regex)
       with_items: "{{ component_contents.stdout_lines }}"
       register: tld_r
 
--- a/trunk/mpf-system-tests/src/test/resources/output/face/runDetectMarkupMultipleMediaTypes.json
+++ b/trunk/mpf-system-tests/src/test/resources/output/face/runDetectMarkupMultipleMediaTypes.json
@@ -50,7 +50,10 @@
       "sha256": "c067e7eed23a0fe022140c30dbfa993ae720309d6567a803d111ecec739a6713",
       "message": null,
       "status": "COMPLETE",
-      "mediaMetadata": {},
+      "mediaMetadata": {
+        "FRAME_HEIGHT": "600",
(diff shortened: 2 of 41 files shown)


jrobble:
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/DetectionPaddingProcessor.java, line 249 at r2](https://reviewable.io/reviews/openmpf/openmpf/1036#-M-QLG5t12LPkitPL4OM:-M-WhXxWACcfrrin9dnr:b-4v8km6) ([raw file](https://github.com/openmpf/openmpf/blob/a85809f07b0bdd9cff1e231df75e639157ed62f5/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/DetectionPaddingProcessor.java#L249)):*

I really don't think silently dropping detections or tracks is the way we want to handle this. The padding properties are supposed to modify the size of the detection, not filter detections out. Doing that when a user just asked to shrink detections is confusing. We have a convention of returning the whole frame when the component can't provide a bounding box. This is just a convention and I could see a component developer deciding to return empty detections in that case.

 I think some of the options to handle empty detections are:
1. Don't modify the detection size. In the case that the user provided a number of pixels, I think leaving the detection alone makes sense since the detection is already smaller than what they asked for. In the case of a %, it is less intuitive to leave the detection alone. I think just rejecting jobs with more than 50% padding would reasonable.
2. Keep the current behavior and just add a job warning.
3. Shrink the detection to 1px X 1px.
4. Don't allow negative padding.
</blockquote></details>

As we discussed, I will:

- Do a sanity check during job creation to make sure percentages are > -50%.
- Not drop detections, but rather ensure a 1-pixel size minimum and generate a WFM warning when that happens.

jrobble:
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/DetectionPaddingProcessor.java, line 249 at r2](https://reviewable.io/reviews/openmpf/openmpf/1036#-M-QLG5t12LPkitPL4OM:-M-pQtfpC5nbGXwI6nk_:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/a85809f07b0bdd9cff1e231df75e639157ed62f5/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/DetectionPaddingProcessor.java#L249)):*

As we discussed, I will:

- Do a sanity check during job creation to make sure percentages are > -50%.
- Not drop detections, but rather ensure a 1-pixel size minimum and generate a WFM warning when that happens.
</blockquote></details>

Done.

PR #1041, opened 2020-02-13
Update detection padding for R4.1.

jrobble:
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1041)

### DIFF
--- a/trunk/mpf-system-tests/src/main/java/org/mitre/mpf/mst/TestSystem.java
+++ b/trunk/mpf-system-tests/src/main/java/org/mitre/mpf/mst/TestSystem.java
@@ -307,7 +307,6 @@ public abstract class TestSystem {
     }
 
 
-
     public List<Point2D.Double> getCorners(JsonDetectionOutputObject detection) {
         double rotationDegrees = Double.parseDouble(detection.getDetectionProperties()
                                                             .getOrDefault("ROTATION", "0"));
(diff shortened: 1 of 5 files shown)


jrobble:
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/padding/DetectionPaddingProcessor.java, line 329 at r2](https://reviewable.io/reviews/openmpf/openmpf/1041#-M0cXzpY29D-t9kWyfFi:-M0cXzpY29D-t9kWyfFj:b2h3y46) ([raw file](https://github.com/openmpf/openmpf/blob/c9d3f12a6a89dabe1e41e87cb45e2e01b8f0a7ee/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/padding/DetectionPaddingProcessor.java#L329)):*
> ```Java
>         int x = (int) rect.getX();
>         int y = (int) rect.getY();
>         int width = (int) Math.ceil(rect.getWidth());
> ```

Doing `ceil` here differs slightly in spirit from my original implementation, since this will always lean towards larger regions regardless if the user is expanding or shrinking, where my original implementation would lean towards smaller regions while shrinking. I think this is fine though. I like that you carry the double precision through more of the calculation, which is probably more important.

PR #1055, opened 2020-03-04
Add output object uri to callback

brosenberg42:
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1055)

### DIFF
--- a/trunk/interop/src/main/java/org/mitre/mpf/interop/JsonCallbackBody.java
+++ b/trunk/interop/src/main/java/org/mitre/mpf/interop/JsonCallbackBody.java
@@ -31,16 +31,23 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 
 public class JsonCallbackBody {
     /** The internal identifier assigned to this job by MPF. */
-    private long jobId;
-    public long getJobId() { return jobId; }
+    private final long _jobId;
+    public long getJobId() { return _jobId; }
 
     /** The ID that was provided to this job when it was initially submitted. */
-    private String externalid;
-    public String getExternalId() { return externalid; }
+    private final String _externalId;
+    public String getExternalId() { return _externalId; }
+
+    private final String _outputObjectUri;
+    public String getOutputObjectUri() { return _outputObjectUri; }
 
     @JsonCreator
-    public JsonCallbackBody(@JsonProperty("jobId") long jobId, @JsonProperty("externalId") String externalid) {
-        this.jobId = jobId;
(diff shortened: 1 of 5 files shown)


jrobble:
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/mvc/controller/JobController.java, line 127 at r1](https://reviewable.io/reviews/openmpf/openmpf/1055#-M1beG0I52m9LjXXc5DL:-M1ceicH6zmv3__a-uxN:b-cqg7oz) ([raw file](https://github.com/openmpf/openmpf/blob/029258eeb6ca90e36afc0a34b988ee2341991640/trunk/workflow-manager/src/main/java/org/mitre/mpf/mvc/controller/JobController.java#L127)):*

Note to self: Check if the slashes in the `outputobjecturi` are escaped.

Should we show the escaped forms here?
</blockquote></details>

Note to self: They are escaped when necessary:

```
GET /?jobid=125&externalid=88&outputobjecturi=file%3A%2F%2F%2Fopt%2Fmpf%2Fshare%2Foutput-objects%2F125%2Fdetection.json HTTP/1.1
```

```
{"jobId":127,"externalId":"88","outputObjectUri":"file:///opt/mpf/share/output-objects/127/detection.json"}
```

jrobble:
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/mvc/controller/JobController.java, line 127 at r1](https://reviewable.io/reviews/openmpf/openmpf/1055#-M1beG0I52m9LjXXc5DL:-M1cfUwm086mr-Ejuz33:b-urhbbu) ([raw file](https://github.com/openmpf/openmpf/blob/029258eeb6ca90e36afc0a34b988ee2341991640/trunk/workflow-manager/src/main/java/org/mitre/mpf/mvc/controller/JobController.java#L127)):*

Note to self: They are escaped when necessary:

```
GET /?jobid=125&externalid=88&outputobjecturi=file%3A%2F%2F%2Fopt%2Fmpf%2Fshare%2Foutput-objects%2F125%2Fdetection.json HTTP/1.1
```

```
{"jobId":127,"externalId":"88","outputObjectUri":"file:///opt/mpf/share/output-objects/127/detection.json"}
```
</blockquote></details>

I'd like to think this is self-explanatory, but I'm not so sure based on past experience: 

We may want to mention that the GET callback will be properly escaped. This not only includes the `outputobjecturi` parameter, but also the `externalid` parameter. I just tested the latter by adding some punctuation to it in my job creation request.

PR #1060, opened 2020-03-12
Add ability to restrict a service's media types.

brosenberg42:
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1060)


jrobble:
___
*[trunk/detection/executor/java/src/main/java/org/mitre/mpf/component/executor/detection/MPFDetectionMain.java, line 159 at r1](https://reviewable.io/reviews/openmpf/openmpf/1060#-M2tIOw27d_ORq-aqLlM:-M2tzVK8Ciygy4n7T76i:b-ejvdj) ([raw file](https://github.com/openmpf/openmpf/blob/186067d4539bbb33333a1e4b0e2ae85af051dcdb/trunk/detection/executor/java/src/main/java/org/mitre/mpf/component/executor/detection/MPFDetectionMain.java#L159)):*

Note to self: Test error condition.
</blockquote></details>

When this happens, the Docker service terminates, a new service is started, the error happens again, that service terminates, a new service is started, and the cycle continues.

Is that the behavior you expect?

It seems that's what should happen since the services are configured to restart by default. Specifically, `restart_policy` is set to `any` by default.

PR #1084, opened 2020-04-17
Update mpf.version.semantic property.

jrobble:
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1084)


jrobble:
Decided that this change does not need to land to R4.0.

PR #1121, opened 2020-06-26
Skip media inspection

hhuangMITRE:
Updated MediaInspectionProcessor to skip inspection when the user provides the proper media metadata via the job properties field. Also updated TestMediaInspectionProcessor to check for media inspection skips.

---
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1121)


jrobble:
@brosenberg42 , please do a sanity review of this PR. I ran a lot of manual media inspection skipping  tests with the WFM and they all performed as expected.

I do not have a corrupt Quicktime video to test the frame checking behavior, but I did recompile by replacing the read condition with `true` and the resulting behavior is as expected:
1. job immediately ends in `ERROR`
2. The `jobError` is `Error counting frames. Native error code -8702.`

I also verified that the final job status is `COMPLETE_WITH_WARNINGS` when media metadata is provided and media inspection skipping cannot be performed.

Side note:

More work needs to be done to get the status in the web UI to show `IN_PROGRESS_WITH_WARNINGS` properly (for this and detection padding overshrink). Also, the [GET] /rest/jobs/{id} endpoint shows `IN_PROGRESS` even after I add a warning for not being able to skip media inspection. I tested that by having a FACE --> MARKUP pipeline, not running a markup component, and getting the status after the first stage of the pipeline.

We do not need to address this as part of this PR, although if we don't the user will not be able to tell that media inspection skipping failed by looking at the job status. A warning message does appear in the WFM log though.

jrobble:
Please merge with develop. Still discussing if we want to land this to master or develop. Will let you know soon.

jrobble:
Please go ahead and land thisto develop.

PR #1166, opened 2020-08-26
Hf merge/feed forward track properties

jrobble:
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1166)


jrobble:
Will land on command line via fast-forward.

Issue #1211, opened 2020-11-24
Only generate FRAME_COUNT warning when the frame difference is > 1


jrobble:
In many cases OpenCV and ffmpeg differ on the frame count by 1, in which case we generate a `FRAME_COUNT` warning. This is not useful to the end user, so we should only generate this warning when the difference is > 1.

To future-proof this change, we should make the value a configurable system property.

jrobble:
Note that OpenCV's frame count was never intended to be accurate. See the accepted answer [here](https://pythonpedia.com/en/knowledge-base/31472155/python-opencv-cv2-cv-cv-cap-prop-frame-count-get-wrong-numbers).

PR #1213, opened 2020-11-25
Feature/preserve track end offsets

jrobble:
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1213)


jrobble (commented):
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1213)* status: 0 of 5 files reviewed, all discussions resolved (waiting on @jrobble)

___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/mediainspection/MediaInspectionProcessor.java, line 486 at r1](https://reviewable.io/reviews/openmpf/openmpf/1213#-MMxeQSlEhnFvU4gxFiL:-MMxeQSlEhnFvU4gxFiM:b-lthx5l) ([raw file](https://github.com/openmpf/openmpf/blob/d1d6c61e88e2344b754e5c28b2b9a3b11de0ff6e/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/mediainspection/MediaInspectionProcessor.java#L486)):*
> ```Java
>                                            "but FFmpeg reported it to be %s. %s will be used.",
>                                            openCvFrameCount, ffmpegFrameCount, frameCount);
>             if (Math.abs(ffmpegFrameCount - openCvFrameCount) >= _propertiesUtil.getWarningFrameCountDiff()) {
> ```

This addresses [this issue](https://github.com/openmpf/openmpf/issues/1211).
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/segmenting/VideoMediaSegmenter.java, line 138 at r1](https://reviewable.io/reviews/openmpf/openmpf/1213#-MMxbLEh-TLhX4b0vw9s:-MMxbLEh-TLhX4b0vw9t:b-twpikj) ([raw file](https://github.com/openmpf/openmpf/blob/d1d6c61e88e2344b754e5c28b2b9a3b11de0ff6e/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/segmenting/VideoMediaSegmenter.java#L138)):*
> ```Java
> 
>         DetectionProtobuf.VideoTrack.Builder videoTrackBuilder = DetectionProtobuf.VideoTrack.newBuilder()
>                 .setStartFrame(track.getStartOffsetFrameInclusive())
> ```

Prior to this change, a `SPEECH DETECTION --> KEYWORD TAGGING` pipeline would result in `SPEECH` tracks in the JSON output object with different `stopOffsetFrame` and `stopOffsetTime` values than the corresponding Keyword Tagging `TEXT` tracks. 

For example, a`SPEECH` track would have:

```
                        "startOffsetFrame": 6,
                        "stopOffsetFrame": 37,
                        "startOffsetTime": 250,
                        "stopOffsetTime": 1542,
```

While the `TEXT` track would have:

```
                        "startOffsetFrame": 6,
                        "stopOffsetFrame": 6,
                        "startOffsetTime": 250,
                        "stopOffsetTime": 250,
```

This is because of how the old code essentially shortens the length of the feed-forward track so that it's bounded at the min and max frames for the detections it contains. 

`SPEECH` tracks only contain one detection, which had an `offsetFrame` of 6 in this case. Thus, the feed-forward track starts at a detection at frame 6 and ends with the same detection at frame 6. Doing the math with video FPS, frame 6 corresponds to a time of 250 ms.

The problem with the old approach is that it did not consider the duration associated with any of the detections in the track. This is not an issue with video jobs since we treat each detection as essentially an instant in time, but detections in audio jobs have a duration that should be preserved.

jrobble:
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/segmenting/VideoMediaSegmenter.java, line 138 at r1](https://reviewable.io/reviews/openmpf/openmpf/1213#-MMxbLEh-TLhX4b0vw9s:-MMy0ZRG6EsuGbpkOzda:b-900bj6) ([raw file](https://github.com/openmpf/openmpf/blob/d1d6c61e88e2344b754e5c28b2b9a3b11de0ff6e/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/segmenting/VideoMediaSegmenter.java#L138)):*

Prior to this change, a `SPEECH DETECTION --> KEYWORD TAGGING` pipeline would result in `SPEECH` tracks in the JSON output object with different `stopOffsetFrame` and `stopOffsetTime` values than the corresponding Keyword Tagging `TEXT` tracks. 

For example, a`SPEECH` track would have:

```
                        "startOffsetFrame": 6,
                        "stopOffsetFrame": 37,
                        "startOffsetTime": 250,
                        "stopOffsetTime": 1542,
```

While the `TEXT` track would have:

```
                        "startOffsetFrame": 6,
                        "stopOffsetFrame": 6,
                        "startOffsetTime": 250,
                        "stopOffsetTime": 250,
```

This is because of how the old code essentially shortens the length of the feed-forward track so that it's bounded at the min and max frames for the detections it contains. 

`SPEECH` tracks only contain one detection, which had an `offsetFrame` of 6 in this case. Thus, the feed-forward track starts at a detection at frame 6 and ends with the same detection at frame 6. Doing the math with video FPS, frame 6 corresponds to a time of 250 ms.

The problem with the old approach is that it did not consider the duration associated with any of the detections in the track. This is not an issue with video jobs since we treat each detection as essentially an instant in time, but detections in audio jobs have a duration that should be preserved. 
</blockquote></details>

One way to rationalize this is that even if we reduce the number of detections that's part of a track by only selecting the detections according to some `FEED_FORWARD_TOP_CONFIDENCE_COUNT`, the object represented by the track still exists in all of the frames (and over the course of the entire duration) captured by the original unmodified track. As such, it makes sense to use the frame range (and the time range) represented by the original track in the modified feed-forward track.

clnowacki (approved):
> Reviewed 5 of 5 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1213)* status: :shipit: complete! all files reviewed, all discussions resolved

brosenberg42 (approved):
> Reviewed 5 of 5 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1213)* status: :shipit: complete! all files reviewed, all discussions resolved

jrobble (commented):
> Reviewed 5 of 5 files at r1, 1 of 1 files at r2, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1213)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jrobble)

PR #1214, opened 2020-11-30
Allow job properties to be set through environment variables.

### LINKED ISSUE #1206: Allow job properties that don't change during service lifetime to be set using environment variables
Modify the component executors to look for environment variables with the `MPF_PROP_` prefix and add them as job properties. Properties defined in this way override existing job properties.

Currently, some components allow particular job properties to be set through environment variables. For example, the Azure components allow `ACS_URL` and `ACS_SUBSCRIPTION_KEY` to be set as environment variables. The current behavior is undesirable because every component that needs this behavior must implement it themselves for each property.

brosenberg42:
Resolves #1206 

This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1214)

### DIFF
--- a/trunk/detection/executor/cpp/batch/BatchExecutorUtil.cpp
+++ b/trunk/detection/executor/cpp/batch/BatchExecutorUtil.cpp
@@ -1,0 +1,62 @@
+/******************************************************************************
+ * NOTICE                                                                     *
+ *                                                                            *
+ * This software (or technical data) was produced for the U.S. Government     *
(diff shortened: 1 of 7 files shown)


jrobble (approved):
> Reviewed 7 of 7 files at r1.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1214)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #1225, opened 2020-12-08
Prevent S3 storage service from using keys after they have been censored

brosenberg42:
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1225)

PR #1222, opened 2020-12-07
Add ability to censor property output

brosenberg42:
Issue: https://github.com/openmpf/openmpf/issues/1219

---
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1222)

PR #1228, opened 2020-12-09
hf-merge-master/fix-s3-storage

brosenberg42:
Merges https://github.com/openmpf/openmpf/pull/1225 and https://github.com/openmpf/openmpf/pull/1222 into master

---
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1228)


jrobble (approved):
As long as you've tested this using Minio, I approve. I was going to do that, but ran into build problems.

PR #1255, opened 2021-02-03
Improve markup

jrobble:
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1255)


brosenberg42 (changes requested):
I pushed a branch, br-review-markup, that contains many of the trivial changes I suggested.

> Reviewed 37 of 37 files at r1.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1255#-:-MTHFtUk1mQUbXlSWskX:b-mkifyv)* status: all files reviewed, 17 unresolved discussions (waiting on @jrobble)

___
*[trunk/markup/src/main/java/org/mitre/mpf/markup/MarkupRequestConsumer.java, line 99 at r1](https://reviewable.io/reviews/openmpf/openmpf/1255#-MTBUVuN5gyrfFRKjPAL:-MTBUVuN5gyrfFRKjPAM:b-dq6wrq) ([raw file](https://github.com/openmpf/openmpf/blob/1f4ac5aa10c81380e5f452ab21ee02de57893a7d/trunk/markup/src/main/java/org/mitre/mpf/markup/MarkupRequestConsumer.java#L99)):*
> ```Java
>         writer.setDestinationMedium(URI.create(markupRequest.getDestinationUri()));
> 
>         Map mediaMetadata = new HashMap<String, String>();
> ```

This line causes the non-generic version of Map to be used. It should either be:
`var mediaMetadata = new HashMap<String, String>();`
or
`Map<String, String> mediaMetadata = new HashMap<>()`
___
*[trunk/markup/src/main/java/org/mitre/mpf/markup/MarkupRequestConsumer.java, line 100 at r1](https://reviewable.io/reviews/openmpf/openmpf/1255#-MTBV9XD6dRci_uTPCSH:-MTBV9XD6dRci_uTPCSI:b87ugpl) ([raw file](https://github.com/openmpf/openmpf/blob/1f4ac5aa10c81380e5f452ab21ee02de57893a7d/trunk/markup/src/main/java/org/mitre/mpf/markup/MarkupRequestConsumer.java#L100)):*
> ```Java
> 
>         Map mediaMetadata = new HashMap<String, String>();
>         markupRequest.getMediaMetadataList().stream().forEach(e -> mediaMetadata.put(e.getKey(), e.getValue()) );
> ```

This is not how you create a map with streams. This should be: 
```java
var mediaMetadata = markupRequest.getMediaMetadataList()
        .stream()
        .collect(toMap(Markup.MetadataMap::getKey, Markup.MetadataMap::getValue))
```
___
*[trunk/video-overlay/src/main/cpp/BoundingBoxImageHandle.h, line 30 at r1](https://reviewable.io/reviews/openmpf/openmpf/1255#-MTBZHWkEW8y3Pjav--_:-MTBZHWkEW8y3Pjav--a:bitvbuj) ([raw file](https://github.com/openmpf/openmpf/blob/1f4ac5aa10c81380e5f452ab21ee02de57893a7d/trunk/video-overlay/src/main/cpp/BoundingBoxImageHandle.h#L30)):*
> ```objc
> #define MPF_BOUNDINGBOXIMAGEHANDLE_H
> 
> #include <MPFVideoCapture.h>
> ```

This is missing:
```
#include <string>

#include <opencv2/core.hpp>
```
___
*[trunk/video-overlay/src/main/cpp/BoundingBoxImageHandle.h, line 36 at r1](https://reviewable.io/reviews/openmpf/openmpf/1255#-MTBZMKJ4E6Vbdd4llai:-MTBZMKKBKVAFc8VJWPO:bmbm93s) ([raw file](https://github.com/openmpf/openmpf/blob/1f4ac5aa10c81380e5f452ab21ee02de57893a7d/trunk/video-overlay/src/main/cpp/BoundingBoxImageHandle.h#L36)):*
> ```objc
>     BoundingBoxImageHandle(std::string sourcePath, std::string destinationPath);
> 
>     cv::Size GetFrameSize();
> ```

This should be declared a const method.
___
*[trunk/video-overlay/src/main/cpp/BoundingBoxImageHandle.cpp, line 32 at r1](https://reviewable.io/reviews/openmpf/openmpf/1255#-MSmm3ow4HODUUhzMwIa:-MSmm3ow4HODUUhzMwIb:bbi8ahf) ([raw file](https://github.com/openmpf/openmpf/blob/1f4ac5aa10c81380e5f452ab21ee02de57893a7d/trunk/video-overlay/src/main/cpp/BoundingBoxImageHandle.cpp#L32)):*
> ```cpp
> BoundingBoxImageHandle::BoundingBoxImageHandle(std::string sourcePath, std::string destinationPath) :
>         sourcePath_(sourcePath),
>         destinationPath_(destinationPath),
> ```

These two strings should be `std::move()`ed in to their member variables.
___
*[trunk/video-overlay/src/main/cpp/BoundingBoxImageHandle.cpp, line 35 at r1](https://reviewable.io/reviews/openmpf/openmpf/1255#-MSmlfua3DV-wp-9W2fF:-MSmlfub5BjCft_IFiHj:bo2p7fu) ([raw file](https://github.com/openmpf/openmpf/blob/1f4ac5aa10c81380e5f452ab21ee02de57893a7d/trunk/video-overlay/src/main/cpp/BoundingBoxImageHandle.cpp#L35)):*
> ```cpp
>         videoCapture_(sourcePath_) {
>     if (!videoCapture_.IsOpened()) {
>         throw std::runtime_error("Unable to open source image: " + sourcePath_);
> ```

This is already checked in the MPFVideoCapture's constructor.
___
*[trunk/video-overlay/src/main/cpp/BoundingBoxVideoHandle.h, line 30 at r1](https://reviewable.io/reviews/openmpf/openmpf/1255#-MSms8SV9kbPNnmUNIGj:-MSms8SW7eikIv7Af73_:b-a9i3w7) ([raw file](https://github.com/openmpf/openmpf/blob/1f4ac5aa10c81380e5f452ab21ee02de57893a7d/trunk/video-overlay/src/main/cpp/BoundingBoxVideoHandle.h#L30)):*
> ```objc
> #define MPF_BOUNDINGBOXVIDEOHANDLE_H
> 
> #include <MPFVideoCapture.h>
> ```

This and its .cpp file are missing a few includes. It is bad to rely on transitive includes, except for when you have `somefile.cpp` it is okay to depend on includes from `somefile.h`.
___
*[trunk/video-overlay/src/main/cpp/BoundingBoxVideoHandle.h, line 38 at r1](https://reviewable.io/reviews/openmpf/openmpf/1255#-MSmzpRWBEQ9AcVawA1P:-MSmzpRWBEQ9AcVawA1Q:b-bwo0an) ([raw file](https://github.com/openmpf/openmpf/blob/1f4ac5aa10c81380e5f452ab21ee02de57893a7d/trunk/video-overlay/src/main/cpp/BoundingBoxVideoHandle.h#L38)):*
> ```objc
>     ~BoundingBoxVideoHandle();
> 
>     cv::Size GetFrameSize();
> ```

This method and few others here and in BoundingBoxImageHandle can be made const.
___
*[trunk/video-overlay/src/main/cpp/BoundingBoxVideoHandle.cpp, line 34 at r1](https://reviewable.io/reviews/openmpf/openmpf/1255#-MSmxEX6077uiKgohm6r:-MSmxEX6077uiKgohm6s:bnbzear) ([raw file](https://github.com/openmpf/openmpf/blob/1f4ac5aa10c81380e5f452ab21ee02de57893a7d/trunk/video-overlay/src/main/cpp/BoundingBoxVideoHandle.cpp#L34)):*
> ```cpp
> BoundingBoxVideoHandle::BoundingBoxVideoHandle(std::string sourcePath, std::string destinationPath, int crf,
>                                                int framePadding) :
>         destinationPath_(destinationPath), videoCapture_(sourcePath) {
> ```

`destinationPath_(destinationPath)` should be `destinationPath_(std::move(destinationPath))` because of  https://clang.llvm.org/extra/clang-tidy/checks/modernize-pass-by-value.html .  `std::string sourcePath` should be `const std::string& sourcePath` because `MPFVideoCapture`'s constructor parameter is a `const std::string&`.
___
*[trunk/video-overlay/src/main/cpp/BoundingBoxVideoHandle.cpp, line 75 at r1](https://reviewable.io/reviews/openmpf/openmpf/1255#-MTGkJJd39Sl1Wi4AmpD:-MTGkJJe11pwC-CPd7mI:bwaz2om) ([raw file](https://github.com/openmpf/openmpf/blob/1f4ac5aa10c81380e5f452ab21ee02de57893a7d/trunk/video-overlay/src/main/cpp/BoundingBoxVideoHandle.cpp#L75)):*
> ```cpp
> void BoundingBoxVideoHandle::HandleMarkedFrame(const cv::Mat& frame) {
>     size_t sizeInBytes = frame.step[0] * frame.rows; // https://stackoverflow.com/a/26441073
>     fwrite(frame.data, sizeof(unsigned char), sizeInBytes, pipe_);
> ```

We should properly handle non-continuous `cv::Mat`s. Cropping off the left or right side of an image will result in the matrix being non-continuous because cropping doesn't copy the matrix, it creates a submatrix pointing in to the original un-cropped frame. This will cause ffmpeg to be passed data with sections it was supposed to skip. Right now, the only reason this isn't an issue is that you make an unnecessary clone on this line: `Mat transformFrame = frame.clone();`. I think we should copy in the data row by row. This shouldn't affect performance since we can avoid a clone and I'm pretty sure the bottle neck is ffmpeg encoding, not copying the frames in to the pipe. I tested removing the clone and changing the lines to use `frame` instead of `transformFrame`. This resulted in a garbled video. After I made the change below, it produced the correct video.

```c++
for (int row = 0; row < frame.rows; ++row) {
    fwrite(frame.ptr(row), frame.elemSize(), frame.cols, pipe_);
}
```
I read the stack overflow link and `frame.elemSize()` is the correct thing to use in the code above. `frame.step[0]` is the width of the frame before cropping. A single row is always continuous so the stack overflow comment does not apply. `frame.elemSize()` is always 3 which is required because of `-pixel_format bgr24`.
___
*[trunk/video-overlay/src/main/cpp/BoundingBoxVideoHandle.cpp, line 88 at r1](https://reviewable.io/reviews/openmpf/openmpf/1255#-MSbxQzm5DrvZ_qX3wXF:-MSbxQzm5DrvZ_qX3wXG:b-6r1xj7) ([raw file](https://github.com/openmpf/openmpf/blob/1f4ac5aa10c81380e5f452ab21ee02de57893a7d/trunk/video-overlay/src/main/cpp/BoundingBoxVideoHandle.cpp#L88)):*
> ```cpp
> void BoundingBoxVideoHandle::Close() {
>     fflush(pipe_);
>     pclose(pipe_);
> ```

We need to check return value of pclose to see if ffmpeg returned a non-zero exit code. 
___
*[trunk/video-overlay/src/main/cpp/BoundingBoxWriter.cpp, line 124 at r1](https://reviewable.io/reviews/openmpf/openmpf/1255#-MSn19M77GSWTAE0qCUt:-MSn19M77GSWTAE0qCUu:b-sa7pz7) ([raw file](https://github.com/openmpf/openmpf/blob/1f4ac5aa10c81380e5f452ab21ee02de57893a7d/trunk/video-overlay/src/main/cpp/BoundingBoxWriter.cpp#L124)):*
> ```cpp
>         std::string destinationImagePath = jni.ToStdString(destinationImagePathJString);
> 
>         BoundingBoxImageHandle boundingBoxImageHandle(sourceImagePath, destinationImagePath);
> ```

If you change this to:
```cpp
BoundingBoxImageHandle boundingBoxImageHandle(
        jni.ToStdString(sourceImagePathJString),
        jni.ToStdString(destinationImagePathJString));
```
the `std::string`s will be moved instead of copied.
___
*[trunk/video-overlay/src/main/cpp/BoundingBoxWriter.cpp, line 496 at r1](https://reviewable.io/reviews/openmpf/openmpf/1255#-MTGs6615lvewoJWmiG3:-MTGs6615lvewoJWmiG4:b-pnwgt2) ([raw file](https://github.com/openmpf/openmpf/blob/1f4ac5aa10c81380e5f452ab21ee02de57893a7d/trunk/video-overlay/src/main/cpp/BoundingBoxWriter.cpp#L496)):*
> ```cpp
> 
>     cv::putText(labelMat, label, Point(labelBottomLeftX, labelBottomLeftY),
>         labelFont, labelScale, color, labelThickness, cv::LineTypes::LINE_8);
> ```

Using `cv::LineTypes::LINE_AA` make the edges of the text less jagged.




___
*[trunk/video-overlay/src/main/java/org/mitre/mpf/videooverlay/BoundingBox.java, line 180 at r1](https://reviewable.io/reviews/openmpf/openmpf/1255#-MTBxAkA49PK0VHWYZwp:-MTBxAkA49PK0VHWYZwq:b-dm7twc) ([raw file](https://github.com/openmpf/openmpf/blob/1f4ac5aa10c81380e5f452ab21ee02de57893a7d/trunk/video-overlay/src/main/java/org/mitre/mpf/videooverlay/BoundingBox.java#L180)):*
> ```Java
>                 && exemplar == casted.exemplar
>                 && confidence == casted.confidence
>                 && classification == casted.classification;
> ```

This should use .equals
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/markup/MarkupSplitter.java, line 211 at r1](https://reviewable.io/reviews/openmpf/openmpf/1255#-MT7WMLKC8qYlaKJ-Xdl:-MT7WMLKC8qYlaKJ-Xdm:b-xemotd) ([raw file](https://github.com/openmpf/openmpf/blob/1f4ac5aa10c81380e5f452ab21ee02de57893a7d/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/markup/MarkupSplitter.java#L211)):*
> ```Java
>             Optional<Boolean> detectionFlip = getFlip(detection.getDetectionProperties());
> 
>             Optional trackClassification = Optional.empty();
> ```

This should be: `Optional<String> trackClassification = Optional.empty();`
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/markup/MarkupSplitter.java, line 290 at r1](https://reviewable.io/reviews/openmpf/openmpf/1255#-MT7VLxtBZpZ6jJWVNoO:-MT7VLxtBZpZ6jJWVNoP:b-bjcuu7) ([raw file](https://github.com/openmpf/openmpf/blob/1f4ac5aa10c81380e5f452ab21ee02de57893a7d/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/markup/MarkupSplitter.java#L290)):*
> ```Java
>     /** Returns the appropriate markup extension for a given {@link MediaType}. */
>     private static String getMarkedUpMediaExtensionForMediaType(MediaType mediaType,
>                                                                 SystemPropertiesSnapshot snapshot) {
> ```

The snapshot parameter is unused.
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/IoUtils.java, line 287 at r1](https://reviewable.io/reviews/openmpf/openmpf/1255#-MTCaGI58fjosXFsZ2cE:-MTCaGI58fjosXFsZ2cF:b1dxc97) ([raw file](https://github.com/openmpf/openmpf/blob/1f4ac5aa10c81380e5f452ab21ee02de57893a7d/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/IoUtils.java#L287)):*
> ```Java
>     }
> 
>     public static void writeContentAsStream(
> ```

This name doesn't make sense. The content is already a stream.
___
*[trunk/workflow-manager/src/main/resources/properties/mpf.properties, line 302 at r1](https://reviewable.io/reviews/openmpf/openmpf/1255#-MTCabsl3SJ_XaJyGYNQ:-MTCabsm4neLP3LXQkyY:b-jkgmia) ([raw file](https://github.com/openmpf/openmpf/blob/1f4ac5aa10c81380e5f452ab21ee02de57893a7d/trunk/workflow-manager/src/main/resources/properties/mpf.properties#L302)):*
> ```INI
> ###################
> # Markup Settings #
> ###################
> ```

I'm pretty sure we are eventually going to have to make these settable through job properties, but this is fine for now.

jrobble (commented):
Thanks. Good catches. I'm have more updates planned for this PR.

> Reviewed 24 of 37 files at r1, 13 of 13 files at r2, 2 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1255#-:-MUAP2n10Pq-73Xow8YI:b-xl4ooh)* status: all files reviewed, 17 unresolved discussions (waiting on @brosenberg42 and @clnowacki)

___
*[trunk/markup/src/main/java/org/mitre/mpf/markup/MarkupRequestConsumer.java, line 99 at r1](https://reviewable.io/reviews/openmpf/openmpf/1255#-MTBUVuN5gyrfFRKjPAL:-MUAezP_6T77QwQ4ZafF:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/1f4ac5aa10c81380e5f452ab21ee02de57893a7d/trunk/markup/src/main/java/org/mitre/mpf/markup/MarkupRequestConsumer.java#L99)):*

This line causes the non-generic version of Map to be used. It should either be:
`var mediaMetadata = new HashMap<String, String>();`
or
`Map<String, String> mediaMetadata = new HashMap<>()`
</blockquote></details>

Done.
___
*[trunk/markup/src/main/java/org/mitre/mpf/markup/MarkupRequestConsumer.java, line 100 at r1](https://reviewable.io/reviews/openmpf/openmpf/1255#-MTBV9XD6dRci_uTPCSH:-MUAf7czECbJwG7NcGO9:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/1f4ac5aa10c81380e5f452ab21ee02de57893a7d/trunk/markup/src/main/java/org/mitre/mpf/markup/MarkupRequestConsumer.java#L100)):*

This is not how you create a map with streams. This should be: 
```java
var mediaMetadata = markupRequest.getMediaMetadataList()
        .stream()
        .collect(toMap(Markup.MetadataMap::getKey, Markup.MetadataMap::getValue))
```
</blockquote></details>

Done.
___
*[trunk/video-overlay/src/main/cpp/BoundingBoxImageHandle.h, line 30 at r1](https://reviewable.io/reviews/openmpf/openmpf/1255#-MTBZHWkEW8y3Pjav--_:-MUAfcB4CIOHhV9oKbvH:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/1f4ac5aa10c81380e5f452ab21ee02de57893a7d/trunk/video-overlay/src/main/cpp/BoundingBoxImageHandle.h#L30)):*

This is missing:
```
#include <string>

#include <opencv2/core.hpp>
```
</blockquote></details>

Done.
___
*[trunk/video-overlay/src/main/cpp/BoundingBoxImageHandle.h, line 36 at r1](https://reviewable.io/reviews/openmpf/openmpf/1255#-MTBZMKJ4E6Vbdd4llai:-MUAfdCL87V6K36WESZg:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/1f4ac5aa10c81380e5f452ab21ee02de57893a7d/trunk/video-overlay/src/main/cpp/BoundingBoxImageHandle.h#L36)):*

This should be declared a const method.
</blockquote></details>

Done.
___
*[trunk/video-overlay/src/main/cpp/BoundingBoxImageHandle.cpp, line 32 at r1](https://reviewable.io/reviews/openmpf/openmpf/1255#-MSmm3ow4HODUUhzMwIa:-MUAfeRU8B5ISWYnZo_F:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/1f4ac5aa10c81380e5f452ab21ee02de57893a7d/trunk/video-overlay/src/main/cpp/BoundingBoxImageHandle.cpp#L32)):*

These two strings should be `std::move()`ed in to their member variables.
</blockquote></details>

Done.
___
*[trunk/video-overlay/src/main/cpp/BoundingBoxImageHandle.cpp, line 35 at r1](https://reviewable.io/reviews/openmpf/openmpf/1255#-MSmlfua3DV-wp-9W2fF:-MUAffuT6PSFtxC4AG_D:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/1f4ac5aa10c81380e5f452ab21ee02de57893a7d/trunk/video-overlay/src/main/cpp/BoundingBoxImageHandle.cpp#L35)):*

This is already checked in the MPFVideoCapture's constructor.
</blockquote></details>

Done.
___
*[trunk/video-overlay/src/main/cpp/BoundingBoxVideoHandle.h, line 30 at r1](https://reviewable.io/reviews/openmpf/openmpf/1255#-MSms8SV9kbPNnmUNIGj:-MUAfhW47sK8kAp38Xwk:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/1f4ac5aa10c81380e5f452ab21ee02de57893a7d/trunk/video-overlay/src/main/cpp/BoundingBoxVideoHandle.h#L30)):*

This and its .cpp file are missing a few includes. It is bad to rely on transitive includes, except for when you have `somefile.cpp` it is okay to depend on includes from `somefile.h`.
</blockquote></details>

Done.
___
*[trunk/video-overlay/src/main/cpp/BoundingBoxVideoHandle.h, line 38 at r1](https://reviewable.io/reviews/openmpf/openmpf/1255#-MSmzpRWBEQ9AcVawA1P:-MUAfiUp3BQQ1D5UCySq:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/1f4ac5aa10c81380e5f452ab21ee02de57893a7d/trunk/video-overlay/src/main/cpp/BoundingBoxVideoHandle.h#L38)):*

This method and few others here and in BoundingBoxImageHandle can be made const.
</blockquote></details>

Done.
___
*[trunk/video-overlay/src/main/cpp/BoundingBoxVideoHandle.cpp, line 34 at r1](https://reviewable.io/reviews/openmpf/openmpf/1255#-MSmxEX6077uiKgohm6r:-MUAflxHCwTBMF9kVuTR:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/1f4ac5aa10c81380e5f452ab21ee02de57893a7d/trunk/video-overlay/src/main/cpp/BoundingBoxVideoHandle.cpp#L34)):*

`destinationPath_(destinationPath)` should be `destinationPath_(std::move(destinationPath))` because of  https://clang.llvm.org/extra/clang-tidy/checks/modernize-pass-by-value.html .  `std::string sourcePath` should be `const std::string& sourcePath` because `MPFVideoCapture`'s constructor parameter is a `const std::string&`.
</blockquote></details>

Done.
___
*[trunk/video-overlay/src/main/cpp/BoundingBoxVideoHandle.cpp, line 75 at r1](https://reviewable.io/reviews/openmpf/openmpf/1255#-MTGkJJd39Sl1Wi4AmpD:-MUAiJnH9aT57yCq2O9n:b-74vy1h) ([raw file](https://github.com/openmpf/openmpf/blob/1f4ac5aa10c81380e5f452ab21ee02de57893a7d/trunk/video-overlay/src/main/cpp/BoundingBoxVideoHandle.cpp#L75)):*

We should properly handle non-continuous `cv::Mat`s. Cropping off the left or right side of an image will result in the matrix being non-continuous because cropping doesn't copy the matrix, it creates a submatrix pointing in to the original un-cropped frame. This will cause ffmpeg to be passed data with sections it was supposed to skip. Right now, the only reason this isn't an issue is that you make an unnecessary clone on this line: `Mat transformFrame = frame.clone();`. I think we should copy in the data row by row. This shouldn't affect performance since we can avoid a clone and I'm pretty sure the bottle neck is ffmpeg encoding, not copying the frames in to the pipe. I tested removing the clone and changing the lines to use `frame` instead of `transformFrame`. This resulted in a garbled video. After I made the change below, it produced the correct video.

```c++
for (int row = 0; row < frame.rows; ++row) {
    fwrite(frame.ptr(row), frame.elemSize(), frame.cols, pipe_);
}
```
I read the stack overflow link and `frame.elemSize()` is the correct thing to use in the code above. `frame.step[0]` is the width of the frame before cropping. A single row is always continuous so the stack overflow comment does not apply. `frame.elemSize()` is always 3 which is required because of `-pixel_format bgr24`.

</blockquote></details>

Good catch. I updated the PR to do it your way.
___
*[trunk/video-overlay/src/main/cpp/BoundingBoxVideoHandle.cpp, line 88 at r1](https://reviewable.io/reviews/openmpf/openmpf/1255#-MSbxQzm5DrvZ_qX3wXF:-MUAkKS2ESzrTbkvTfe9:bxaed7x) ([raw file](https://github.com/openmpf/openmpf/blob/1f4ac5aa10c81380e5f452ab21ee02de57893a7d/trunk/video-overlay/src/main/cpp/BoundingBoxVideoHandle.cpp#L88)):*

We need to check return value of pclose to see if ffmpeg returned a non-zero exit code. 
</blockquote></details>

Ok. I based my update on what you do in `KeyFrameFilter.cpp`.
___
*[trunk/video-overlay/src/main/cpp/BoundingBoxWriter.cpp, line 124 at r1](https://reviewable.io/reviews/openmpf/openmpf/1255#-MSn19M77GSWTAE0qCUt:-MUAiT2a2yYE0B2kfYqw:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/1f4ac5aa10c81380e5f452ab21ee02de57893a7d/trunk/video-overlay/src/main/cpp/BoundingBoxWriter.cpp#L124)):*

If you change this to:
```cpp
BoundingBoxImageHandle boundingBoxImageHandle(
        jni.ToStdString(sourceImagePathJString),
        jni.ToStdString(destinationImagePathJString));
```
the `std::string`s will be moved instead of copied.
</blockquote></details>

Done.
___
*[trunk/video-overlay/src/main/cpp/BoundingBoxWriter.cpp, line 496 at r1](https://reviewable.io/reviews/openmpf/openmpf/1255#-MTGs6615lvewoJWmiG3:-MUAiUYs90-X-4mSCCWn:b-o8ytd8) ([raw file](https://github.com/openmpf/openmpf/blob/1f4ac5aa10c81380e5f452ab21ee02de57893a7d/trunk/video-overlay/src/main/cpp/BoundingBoxWriter.cpp#L496)):*

Using `cv::LineTypes::LINE_AA` make the edges of the text less jagged.





</blockquote></details>

Let's go with AA then.

I did some more testing with smaller values for `labelScale` and `labelThickness`. Reducing the former results in text that's a bit too close together to read (at least in the 320x240 video I was testing). Reducing the latter results in very then lines that are hard to read. Thus, I'm going to do what we discussed with @clnowacki and add an optional alpha setting.
___
*[trunk/video-overlay/src/main/java/org/mitre/mpf/videooverlay/BoundingBox.java, line 180 at r1](https://reviewable.io/reviews/openmpf/openmpf/1255#-MTBxAkA49PK0VHWYZwp:-MUAjPEN1VTGANOjfuMr:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/1f4ac5aa10c81380e5f452ab21ee02de57893a7d/trunk/video-overlay/src/main/java/org/mitre/mpf/videooverlay/BoundingBox.java#L180)):*

This should use .equals
</blockquote></details>

Done.
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/markup/MarkupSplitter.java, line 211 at r1](https://reviewable.io/reviews/openmpf/openmpf/1255#-MT7WMLKC8qYlaKJ-Xdl:-MUAjT79BA3sOxVwEih_:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/1f4ac5aa10c81380e5f452ab21ee02de57893a7d/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/markup/MarkupSplitter.java#L211)):*

This should be: `Optional<String> trackClassification = Optional.empty();`
</blockquote></details>

Done.
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/markup/MarkupSplitter.java, line 290 at r1](https://reviewable.io/reviews/openmpf/openmpf/1255#-MT7VLxtBZpZ6jJWVNoO:-MUAjUpE0TfX0UDsgSdM:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/1f4ac5aa10c81380e5f452ab21ee02de57893a7d/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/markup/MarkupSplitter.java#L290)):*

The snapshot parameter is unused.
</blockquote></details>

Done.
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/IoUtils.java, line 287 at r1](https://reviewable.io/reviews/openmpf/openmpf/1255#-MTCaGI58fjosXFsZ2cE:-MUAjWXCDzdpyp74iadx:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/1f4ac5aa10c81380e5f452ab21ee02de57893a7d/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/IoUtils.java#L287)):*

This name doesn't make sense. The content is already a stream.
</blockquote></details>

Done.

jrobble (commented):
> Reviewed 19 of 19 files at r4.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1255)* status: all files reviewed, 19 unresolved discussions (waiting on @brosenberg42 and @clnowacki)

___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/markup/MarkupSplitter.java, line 354 at r4](https://reviewable.io/reviews/openmpf/openmpf/1255#-MUYH1so2Zmhj9LIZ5nT:-MUYH1so2Zmhj9LIZ5nU:b-hrz1l6) ([raw file](https://github.com/openmpf/openmpf/blob/8f4c4c45bd78009784470971f8676667972cd3fb/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/markup/MarkupSplitter.java#L354)):*
> ```Java
>         String label = "";
>         if (textPart != null) {
>             label += String.format("%.10s", textPart).strip();
> ```

I decided to omit the "..." at the end of long labels to save space.
___
*[trunk/workflow-manager/src/main/resources/workflow-properties.json, line 35 at r4](https://reviewable.io/reviews/openmpf/openmpf/1255#-MUYHJXx4E3lQQHYM7_-:-MUYHJXx4E3lQQHYM7_0:b-ut13k8) ([raw file](https://github.com/openmpf/openmpf/blob/8f4c4c45bd78009784470971f8676667972cd3fb/trunk/workflow-manager/src/main/resources/workflow-properties.json#L35)):*
> ```JSON
>         "type": "BOOLEAN",
>         "propertiesKey": "mpf.output.objects.last.task.only",
>         "mediaTypes": ["VIDEO", "IMAGE", "AUDIO", "UNKNOWN"]
> ```

Resolves https://github.com/openmpf/openmpf/issues/1257.
___
*[trunk/workflow-manager/src/main/resources/properties/mpf.properties, line 302 at r1](https://reviewable.io/reviews/openmpf/openmpf/1255#-MTCabsl3SJ_XaJyGYNQ:-MUYGg-J2BBPmsMJG9LW:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/1f4ac5aa10c81380e5f452ab21ee02de57893a7d/trunk/workflow-manager/src/main/resources/properties/mpf.properties#L302)):*

I'm pretty sure we are eventually going to have to make these settable through job properties, but this is fine for now. 
</blockquote></details>

Done.

brosenberg42 (changes requested):
> Reviewed 6 of 13 files at r2, 8 of 19 files at r4, 16 of 16 files at r5.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1255)* status: all files reviewed, 26 unresolved discussions (waiting on @jrobble)

___
*[trunk/video-overlay/src/main/cpp/BoundingBoxVideoHandle.cpp, line 36 at r5](https://reviewable.io/reviews/openmpf/openmpf/1255#-MVWig9P65LXidRA_jTo:-MVWig9P65LXidRA_jTp:b3ogpdp) ([raw file](https://github.com/openmpf/openmpf/blob/f5392eb02631fe52099b51b7f6e994ac8467ee33/trunk/video-overlay/src/main/cpp/BoundingBoxVideoHandle.cpp#L36)):*
> ```cpp
> 
> 
> BoundingBoxVideoHandle::BoundingBoxVideoHandle(const std::string &sourcePath, std::string destinationPath,
> ```

`sourcePath` is unused.
___
*[trunk/video-overlay/src/main/cpp/BoundingBoxVideoHandle.cpp, line 37 at r5](https://reviewable.io/reviews/openmpf/openmpf/1255#-MVWjpTJBu033TniWE9A:-MVWjpTKALcSt0Eg_BFo:b-xj0yfi) ([raw file](https://github.com/openmpf/openmpf/blob/f5392eb02631fe52099b51b7f6e994ac8467ee33/trunk/video-overlay/src/main/cpp/BoundingBoxVideoHandle.cpp#L37)):*
> ```cpp
> 
> BoundingBoxVideoHandle::BoundingBoxVideoHandle(const std::string &sourcePath, std::string destinationPath,
>                                                std::string &encoder, int vp9Crf, bool border,
> ```

`encoder` should be const
___
*[trunk/video-overlay/src/main/cpp/BoundingBoxVideoHandle.cpp, line 40 at r5](https://reviewable.io/reviews/openmpf/openmpf/1255#-MVWhSGx0tOhVKeo02EJ:-MVWhSGy4owg3SMfHf8m:b-7d3q0b) ([raw file](https://github.com/openmpf/openmpf/blob/f5392eb02631fe52099b51b7f6e994ac8467ee33/trunk/video-overlay/src/main/cpp/BoundingBoxVideoHandle.cpp#L40)):*
> ```cpp
>                                                const ResolutionConfig &resCfg,
>                                                MPF::COMPONENT::MPFVideoCapture &videoCapture) :
>         destinationPath_(std::move(destinationPath)), videoCapture_(std::move(videoCapture)) {
> ```

Moving a parameter that was passed by reference is dangerous because the caller has no idea that the function moved the object and that it is no longer valid. `MPFVideoCapture` should be passed by value and moved in at the call site. So
this:
```cpp
BoundingBoxVideoHandle boundingBoxVideoHandle(sourceVideoPath, destinationVideoPath, encoder, vp9Crf, border,
                                              resCfg, videoCapture)
```
should be changed to:
```cpp
BoundingBoxVideoHandle boundingBoxVideoHandle(sourceVideoPath, destinationVideoPath, encoder, vp9Crf, border,
                                              resCfg, std::move(videoCapture))
```
___
*[trunk/video-overlay/src/main/cpp/BoundingBoxVideoHandle.cpp, line 57 at r5](https://reviewable.io/reviews/openmpf/openmpf/1255#-MVY4xrh-zRt8gXzeVq6:-MVY4xrh-zRt8gXzeVq7:b-whjg86) ([raw file](https://github.com/openmpf/openmpf/blob/f5392eb02631fe52099b51b7f6e994ac8467ee33/trunk/video-overlay/src/main/cpp/BoundingBoxVideoHandle.cpp#L57)):*
> ```cpp
>         " -f rawvideo" +
>         " -i -" +
>         " -pix_fmt yuv420p"; // https://trac.ffmpeg.org/ticket/5276
> ```

A very brief explanation with the link comments would be helpful since the link eventually not work.
___
*[trunk/video-overlay/src/main/cpp/BoundingBoxVideoHandle.cpp, line 59 at r5](https://reviewable.io/reviews/openmpf/openmpf/1255#-MVq0SbW-QI7bVbaIzNg:-MVq0SbW-QI7bVbaIzNh:blbx055) ([raw file](https://github.com/openmpf/openmpf/blob/f5392eb02631fe52099b51b7f6e994ac8467ee33/trunk/video-overlay/src/main/cpp/BoundingBoxVideoHandle.cpp#L59)):*
> ```cpp
>         " -pix_fmt yuv420p"; // https://trac.ffmpeg.org/ticket/5276
> 
>     if (std::string("vp9") == encoder) { // .webm
> ```

This should just be `if ("vp9" == encoder) { ` There is a `==` overload for `const char*` and `std::string`.
___
*[trunk/video-overlay/src/main/cpp/BoundingBoxWriter.cpp, line 60 at r5](https://reviewable.io/reviews/openmpf/openmpf/1255#-MVaie2y-3seBC_9Eq91:-MVWld31-af9NtyrX3wq:b-ft9bvi) ([raw file](https://github.com/openmpf/openmpf/blob/f5392eb02631fe52099b51b7f6e994ac8467ee33/trunk/video-overlay/src/main/cpp/BoundingBoxWriter.cpp#L60)):*
> ```cpp
> 
> // https://unicode.org/emoji/charts/full-emoji-list.html
> const std::string fastForwardEmoji = "\U000023E9";
> ```

These declaration all trigger https://clang.llvm.org/extra/clang-tidy/checks/cert-err58-cpp.html . These should all be declared like `constexpr const char * fastForwardEmoji = "\U000023E9";`
___
*[trunk/video-overlay/src/main/cpp/BoundingBoxWriter.cpp, line 67 at r5](https://reviewable.io/reviews/openmpf/openmpf/1255#-MVWnSMjAnVPk-yzICCD:-MVWnSMjAnVPk-yzICCE:bpwzp2a) ([raw file](https://github.com/openmpf/openmpf/blob/f5392eb02631fe52099b51b7f6e994ac8467ee33/trunk/video-overlay/src/main/cpp/BoundingBoxWriter.cpp#L67)):*
> ```cpp
> const std::string starEmoji        = "\U00002B50";
> 
> auto ft = cv::freetype::createFreeType2(); // for emojis
> ```

This shouldn't be a global variable and it should have a more descriptive name
___
*[trunk/video-overlay/src/main/cpp/BoundingBoxWriter.cpp, line 397 at r5](https://reviewable.io/reviews/openmpf/openmpf/1255#-MVb_C72-TJirbDnRvMh:-MVb_C72-TJirbDnRvMi:b405uq3) ([raw file](https://github.com/openmpf/openmpf/blob/f5392eb02631fe52099b51b7f6e994ac8467ee33/trunk/video-overlay/src/main/cpp/BoundingBoxWriter.cpp#L397)):*
> ```cpp
> 
> 
> const ResolutionConfig getResolutionConfig(int width, int height) {
> ```

Returning a const value is the same as just returning a value so it shouldn't be used when returning by value. https://clang.llvm.org/extra/clang-tidy/checks/readability-const-return-type.html
___
*[trunk/video-overlay/src/main/cpp/BoundingBoxWriter.cpp, line 398 at r5](https://reviewable.io/reviews/openmpf/openmpf/1255#-MVWttAg-N0ekQp21TGP:-MVWttAg-N0ekQp21TGQ:b7ceq0o) ([raw file](https://github.com/openmpf/openmpf/blob/f5392eb02631fe52099b51b7f6e994ac8467ee33/trunk/video-overlay/src/main/cpp/BoundingBoxWriter.cpp#L398)):*
> ```cpp
> 
> const ResolutionConfig getResolutionConfig(int width, int height) {
>     ft->loadFontData("/usr/share/fonts/google-noto-emoji/NotoEmoji-Regular.ttf", 0);
> ```

We should handle the case where the font doesn't exist. Currently, this is the error you get: `"Error writing bounding box(es). OpenCV(4.5.0) /apps/source/opencv4/opencv_contrib-4.5.0/modules/freetype/src/freetype.cpp:175: error: (-215:Assertion failed) !FT_New_Face( mLibrary, fontFileName.c_str(), idx, &(mFace) ) in function 'loadFontData'\n"`
___
*[trunk/video-overlay/src/main/cpp/BoundingBoxWriter.cpp, line 467 at r5](https://reviewable.io/reviews/openmpf/openmpf/1255#-MVWp2YcE6YQpSCMFaGZ:-MVWp2YcE6YQpSCMFaG_:b-4n2ytm) ([raw file](https://github.com/openmpf/openmpf/blob/f5392eb02631fe52099b51b7f6e994ac8467ee33/trunk/video-overlay/src/main/cpp/BoundingBoxWriter.cpp#L467)):*
> ```cpp
> 
> void drawBoundingBox(int x, int y, int width, int height, double boxRotation, bool boxFlip, double mediaRotation,
>                      bool mediaFlip, int red, int green, int blue, double alpha, BoundingBoxSource source,
> ```

`source` is unused
___
*[trunk/video-overlay/src/main/cpp/BoundingBoxWriter.cpp, line 518 at r5](https://reviewable.io/reviews/openmpf/openmpf/1255#-MVabDjY-piIAQswB6hX:-MVabDjY-piIAQswB6hY:b-qoeeni) ([raw file](https://github.com/openmpf/openmpf/blob/f5392eb02631fe52099b51b7f6e994ac8467ee33/trunk/video-overlay/src/main/cpp/BoundingBoxWriter.cpp#L518)):*
> ```cpp
> }
> 
> void drawLine(const Point2d &start, const Point2d &end, const Scalar &color, bool dashed,
> ```

`dashed` is always false.
___
*[trunk/video-overlay/src/main/cpp/BoundingBoxWriter.cpp, line 599 at r5](https://reviewable.io/reviews/openmpf/openmpf/1255#-MVat0cH8GgcCd2Y-Oo2:-MVat0cH8GgcCd2Y-Oo3:b-6ihf8e) ([raw file](https://github.com/openmpf/openmpf/blob/f5392eb02631fe52099b51b7f6e994ac8467ee33/trunk/video-overlay/src/main/cpp/BoundingBoxWriter.cpp#L599)):*
> ```cpp
>     int labelRectTopRightY = pt.y - textLabelSize.height - (2 * resCfg.labelPadding);
> 
>     int labelRectWidth = labelRectTopRightX - labelRectBottomLeftX;
> ```

You are subtracting something you just added. This appears to be equivalent to `int labelRectWidth = resCfg.labelIndent + emojiLabelSize.width + textLabelSize.width + resCfg.labelPadding;` and `labelRectTopRightX` is not used anywhere else.
___
*[trunk/video-overlay/src/main/cpp/BoundingBoxWriter.cpp, line 600 at r5](https://reviewable.io/reviews/openmpf/openmpf/1255#-MVatUUjCurHkunkKF_I:-MVatUUjCurHkunkKF_J:b2f1w5g) ([raw file](https://github.com/openmpf/openmpf/blob/f5392eb02631fe52099b51b7f6e994ac8467ee33/trunk/video-overlay/src/main/cpp/BoundingBoxWriter.cpp#L600)):*
> ```cpp
> 
>     int labelRectWidth = labelRectTopRightX - labelRectBottomLeftX;
>     int labelRectHeight = labelRectBottomLeftY - labelRectTopRightY;
> ```

You are adding (once you distribute the -) back something you just subtracted. This appears to be equivalent to `int labelRectHeight = textLabelSize.height + (2 * resCfg.labelPadding);`. `labelRectBottomLeftY` is not used anywhere else.
___
*[trunk/video-overlay/src/main/cpp/BoundingBoxWriter.cpp, line 670 at r5](https://reviewable.io/reviews/openmpf/openmpf/1255#-MVqKkd_1HmSCWvY1MU_:-MVqKkd_1HmSCWvY1MUa:b6khrk) ([raw file](https://github.com/openmpf/openmpf/blob/f5392eb02631fe52099b51b7f6e994ac8467ee33/trunk/video-overlay/src/main/cpp/BoundingBoxWriter.cpp#L670)):*
> ```cpp
>         // Depending on the position of the detection relative to the frame boundary, sometimes the label cannot be
>         // drawn within the viewable region. This is fine. Log and continue.
>         std::cerr << "Warning: Label outside of viewable region." << std::endl;
> ```

Doesn't this result in certain labels not showing up? Shouldn't we draw the visible part? I think if we do `cv::Rect(cv::Point(0, 0), image.size()) & addedLabelMatInsertRect` we'll get just the region that is in the image.
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/markup/MarkupSplitter.java, line 218 at r5](https://reviewable.io/reviews/openmpf/openmpf/1255#-MVXLg9r31jYH8ZbRRqu:-MVXLg9r31jYH8ZbRRqv:b5w74ny) ([raw file](https://github.com/openmpf/openmpf/blob/f5392eb02631fe52099b51b7f6e994ac8467ee33/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/markup/MarkupSplitter.java#L218)):*
> ```Java
>         if (!labelFromDetections) { // get track-level details
>             label = getLabel(track, labelTextPropToShow, labelNumericPropToShow);
>             if (track.getTrackProperties().containsKey("MOVING")) {
> ```

This check is not needed. `Boolean.parseBoolean()` handles nulls.
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/markup/MarkupSplitter.java, line 231 at r5](https://reviewable.io/reviews/openmpf/openmpf/1255#-MVXM8Ow9KGp5VXTLy7p:-MVXM8Ow9KGp5VXTLy7q:bvvlh7l) ([raw file](https://github.com/openmpf/openmpf/blob/f5392eb02631fe52099b51b7f6e994ac8467ee33/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/markup/MarkupSplitter.java#L231)):*
> ```Java
>             BoundingBoxSource detectionSource = BoundingBoxSource.DETECTION_ALGORITHM;
>             if (detection.getDetectionProperties().containsKey("FILLED_GAP")) {
>                 if (detection.getDetectionProperties().get("FILLED_GAP").equalsIgnoreCase("true")) {
> ```

This should use `Boolean.parseBoolean()` like the others. It will allow you to remove the `detection.getDetectionProperties().containsKey("FILLED_GAP")` check
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/markup/MarkupSplitter.java, line 241 at r5](https://reviewable.io/reviews/openmpf/openmpf/1255#-MVXN2LM5Rq_fZnONBZm:-MVXN2LM5Rq_fZnONBZn:b5w74ny) ([raw file](https://github.com/openmpf/openmpf/blob/f5392eb02631fe52099b51b7f6e994ac8467ee33/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/markup/MarkupSplitter.java#L241)):*
> ```Java
>             if (labelFromDetections) { // get detection-level details
>                 label = getLabel(detection, labelTextPropToShow, labelNumericPropToShow);
>                 if (detection.getDetectionProperties().containsKey("MOVING")) {
> ```

This check is not needed. `Boolean.parseBoolean()` handles nulls.
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/markup/MarkupSplitter.java, line 383 at r5](https://reviewable.io/reviews/openmpf/openmpf/1255#-MVXVIFLDeaWomMaA6hm:-MVXVIFLDeaWomMaA6hn:bqhhnu5) ([raw file](https://github.com/openmpf/openmpf/blob/f5392eb02631fe52099b51b7f6e994ac8467ee33/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/markup/MarkupSplitter.java#L383)):*
> ```Java
>     public static Optional<String> getLabel(Track track, String textProp, String numericProp) {
>         String textStr = null;
>         if (track.getTrackProperties().containsKey(textProp)) {
> ```

None of `.containsKey` checks in this method appear to be necessary. 
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/markup/MarkupSplitter.java, line 402 at r5](https://reviewable.io/reviews/openmpf/openmpf/1255#-MVXOxU71I62-8I7R1ku:-MVXOlm97l93JcevomMn:b-rl8l0x) ([raw file](https://github.com/openmpf/openmpf/blob/f5392eb02631fe52099b51b7f6e994ac8467ee33/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/markup/MarkupSplitter.java#L402)):*
> ```Java
>     public static Optional<String> getLabel(Detection detection, String textProp, String numericProp) {
>         String textStr = null;
>         if (detection.getDetectionProperties().containsKey(textProp)) {
> ```

This check is not needed. `String textStr = detection.getDetectionProperties().get(textProp);` is the same thing as what you have here.
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/markup/MarkupSplitter.java, line 406 at r5](https://reviewable.io/reviews/openmpf/openmpf/1255#-MVXPcSc5zVk87cvkPxE:-MVXPcSc5zVk87cvkPxF:buzy0lh) ([raw file](https://github.com/openmpf/openmpf/blob/f5392eb02631fe52099b51b7f6e994ac8467ee33/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/markup/MarkupSplitter.java#L406)):*
> ```Java
>         }
>         String numericStr = null;
>         if (detection.getDetectionProperties().containsKey(numericProp)) {
> ```

This check is not needed. `String numericStr = detection.getDetectionProperties().get(numericProp);` is the same thing as what you have here.
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/markup/MarkupSplitter.java, line 409 at r5](https://reviewable.io/reviews/openmpf/openmpf/1255#-MVXTo6W8FeXigYDoad3:-MVXTo6W8FeXigYDoad4:br8rmci) ([raw file](https://github.com/openmpf/openmpf/blob/f5392eb02631fe52099b51b7f6e994ac8467ee33/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/markup/MarkupSplitter.java#L409)):*
> ```Java
>             numericStr = detection.getDetectionProperties().get(numericProp);
>         }
>         if (numericStr == null && numericProp.equals("CONFIDENCE")) {
> ```

I think we should use `.equalsIgnoreCase` when comparing against "CONFIDENCE".
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/service/WorkflowPropertyService.java, line 52 at r5](https://reviewable.io/reviews/openmpf/openmpf/1255#-MVXaD27BM2EQ4i-hO_A:-MVXaD27BM2EQ4i-hO_B:bgkdpm7) ([raw file](https://github.com/openmpf/openmpf/blob/f5392eb02631fe52099b51b7f6e994ac8467ee33/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/service/WorkflowPropertyService.java#L52)):*
> ```Java
> public class WorkflowPropertyService {
> 
>     private static final String _propertiesFileKey = "workflow.properties.file";
> ```

This is unused
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/service/WorkflowPropertyService.java, line 66 at r5](https://reviewable.io/reviews/openmpf/openmpf/1255#-MVXahxl97ptJ194Fhjh:-MVXahxl97ptJ194Fhji:bnc4lj0) ([raw file](https://github.com/openmpf/openmpf/blob/f5392eb02631fe52099b51b7f6e994ac8467ee33/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/service/WorkflowPropertyService.java#L66)):*
> ```Java
>     }
> 
>     public WorkflowPropertyService(PropertiesUtil propertiesUtil, ObjectMapper objectMapper, Resource propertiesFile)
> ```

This can be declared `protected`
___
*[trunk/workflow-manager/src/main/resources/markup-properties.json, line 80 at r5](https://reviewable.io/reviews/openmpf/openmpf/1255#-MVXJKy4CSuZZTFg_ctr:-MVXJKy4CSuZZTFg_cts:b-dzpzes) ([raw file](https://github.com/openmpf/openmpf/blob/f5392eb02631fe52099b51b7f6e994ac8467ee33/trunk/workflow-manager/src/main/resources/markup-properties.json#L80)):*
> ```JSON
>     },
>     {
>         "name": "MARKUP_VIDEO_ENCODER",
> ```

I'm pretty sure these would typically be called codecs and ffmpeg would be the encoder.

jrobble (commented):
> Reviewed 12 of 16 files at r5, 5 of 5 files at r6.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1255)* status: all files reviewed, 25 unresolved discussions (waiting on @brosenberg42)

___
*[trunk/video-overlay/src/main/cpp/BoundingBoxVideoHandle.cpp, line 36 at r5](https://reviewable.io/reviews/openmpf/openmpf/1255#-MVWig9P65LXidRA_jTo:-MVwzSSK73A-IpvFgJmA:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/f5392eb02631fe52099b51b7f6e994ac8467ee33/trunk/video-overlay/src/main/cpp/BoundingBoxVideoHandle.cpp#L36)):*

`sourcePath` is unused.
</blockquote></details>

Done.
___
*[trunk/video-overlay/src/main/cpp/BoundingBoxVideoHandle.cpp, line 37 at r5](https://reviewable.io/reviews/openmpf/openmpf/1255#-MVWjpTJBu033TniWE9A:-MVwzSkRCnD1i5MO8JiF:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/f5392eb02631fe52099b51b7f6e994ac8467ee33/trunk/video-overlay/src/main/cpp/BoundingBoxVideoHandle.cpp#L37)):*

`encoder` should be const
</blockquote></details>

Done.
___
*[trunk/video-overlay/src/main/cpp/BoundingBoxVideoHandle.cpp, line 40 at r5](https://reviewable.io/reviews/openmpf/openmpf/1255#-MVWhSGx0tOhVKeo02EJ:-MW03rqS7wJX6Q751FuZ:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/f5392eb02631fe52099b51b7f6e994ac8467ee33/trunk/video-overlay/src/main/cpp/BoundingBoxVideoHandle.cpp#L40)):*

Moving a parameter that was passed by reference is dangerous because the caller has no idea that the function moved the object and that it is no longer valid. `MPFVideoCapture` should be passed by value and moved in at the call site. So
this:
```cpp
BoundingBoxVideoHandle boundingBoxVideoHandle(sourceVideoPath, destinationVideoPath, encoder, vp9Crf, border,
                                              resCfg, videoCapture)
```
should be changed to:
```cpp
BoundingBoxVideoHandle boundingBoxVideoHandle(sourceVideoPath, destinationVideoPath, encoder, vp9Crf, border,
                                              resCfg, std::move(videoCapture))
```
</blockquote></details>

Done.
___
*[trunk/video-overlay/src/main/cpp/BoundingBoxVideoHandle.cpp, line 57 at r5](https://reviewable.io/reviews/openmpf/openmpf/1255#-MVY4xrh-zRt8gXzeVq6:-MW0rWkH5Ka2Lx2HopyG:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/f5392eb02631fe52099b51b7f6e994ac8467ee33/trunk/video-overlay/src/main/cpp/BoundingBoxVideoHandle.cpp#L57)):*

A very brief explanation with the link comments would be helpful since the link eventually not work.
</blockquote></details>

Done.
___
*[trunk/video-overlay/src/main/cpp/BoundingBoxVideoHandle.cpp, line 59 at r5](https://reviewable.io/reviews/openmpf/openmpf/1255#-MVq0SbW-QI7bVbaIzNg:-MW0rY527EWCjPJmsNmF:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/f5392eb02631fe52099b51b7f6e994ac8467ee33/trunk/video-overlay/src/main/cpp/BoundingBoxVideoHandle.cpp#L59)):*

This should just be `if ("vp9" == encoder) { ` There is a `==` overload for `const char*` and `std::string`.
</blockquote></details>

Done.
___
*[trunk/video-overlay/src/main/cpp/BoundingBoxWriter.cpp, line 60 at r5](https://reviewable.io/reviews/openmpf/openmpf/1255#-MVaie2y-3seBC_9Eq91:-MW11zP04SzVO2gn9hZl:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/f5392eb02631fe52099b51b7f6e994ac8467ee33/trunk/video-overlay/src/main/cpp/BoundingBoxWriter.cpp#L60)):*

These declaration all trigger https://clang.llvm.org/extra/clang-tidy/checks/cert-err58-cpp.html . These should all be declared like `constexpr const char * fastForwardEmoji = "\U000023E9";`
</blockquote></details>

Done.
___
*[trunk/video-overlay/src/main/cpp/BoundingBoxWriter.cpp, line 67 at r5](https://reviewable.io/reviews/openmpf/openmpf/1255#-MVWnSMjAnVPk-yzICCD:-MWQi4Ks7VmVJvVku3zk:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/f5392eb02631fe52099b51b7f6e994ac8467ee33/trunk/video-overlay/src/main/cpp/BoundingBoxWriter.cpp#L67)):*

This shouldn't be a global variable and it should have a more descriptive name
</blockquote></details>

Done.
___
*[trunk/video-overlay/src/main/cpp/BoundingBoxWriter.cpp, line 397 at r5](https://reviewable.io/reviews/openmpf/openmpf/1255#-MVb_C72-TJirbDnRvMh:-MW1C9AC50rPt78ieWI3:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/f5392eb02631fe52099b51b7f6e994ac8467ee33/trunk/video-overlay/src/main/cpp/BoundingBoxWriter.cpp#L397)):*

Returning a const value is the same as just returning a value so it shouldn't be used when returning by value. https://clang.llvm.org/extra/clang-tidy/checks/readability-const-return-type.html
</blockquote></details>

Done.
___
*[trunk/video-overlay/src/main/cpp/BoundingBoxWriter.cpp, line 398 at r5](https://reviewable.io/reviews/openmpf/openmpf/1255#-MVWttAg-N0ekQp21TGP:-MWQ2D7l6FhgQB0Lg5p4:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/f5392eb02631fe52099b51b7f6e994ac8467ee33/trunk/video-overlay/src/main/cpp/BoundingBoxWriter.cpp#L398)):*

We should handle the case where the font doesn't exist. Currently, this is the error you get: `"Error writing bounding box(es). OpenCV(4.5.0) /apps/source/opencv4/opencv_contrib-4.5.0/modules/freetype/src/freetype.cpp:175: error: (-215:Assertion failed) !FT_New_Face( mLibrary, fontFileName.c_str(), idx, &(mFace) ) in function 'loadFontData'\n"`
</blockquote></details>

Done.
___
*[trunk/video-overlay/src/main/cpp/BoundingBoxWriter.cpp, line 467 at r5](https://reviewable.io/reviews/openmpf/openmpf/1255#-MVWp2YcE6YQpSCMFaGZ:-MWPqcf--giOP55S4DOe:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/f5392eb02631fe52099b51b7f6e994ac8467ee33/trunk/video-overlay/src/main/cpp/BoundingBoxWriter.cpp#L467)):*

`source` is unused
</blockquote></details>

Done.
___
*[trunk/video-overlay/src/main/cpp/BoundingBoxWriter.cpp, line 518 at r5](https://reviewable.io/reviews/openmpf/openmpf/1255#-MVabDjY-piIAQswB6hX:-MWPsntt8-jRvtlx0zap:b-d46dup) ([raw file](https://github.com/openmpf/openmpf/blob/f5392eb02631fe52099b51b7f6e994ac8467ee33/trunk/video-overlay/src/main/cpp/BoundingBoxWriter.cpp#L518)):*

`dashed` is always false.
</blockquote></details>

Removed parameter. Commented out dashed line code for potential future use.
___
*[trunk/video-overlay/src/main/cpp/BoundingBoxWriter.cpp, line 599 at r5](https://reviewable.io/reviews/openmpf/openmpf/1255#-MVat0cH8GgcCd2Y-Oo2:-MWPthPkBwC1h-XYB1RW:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/f5392eb02631fe52099b51b7f6e994ac8467ee33/trunk/video-overlay/src/main/cpp/BoundingBoxWriter.cpp#L599)):*

You are subtracting something you just added. This appears to be equivalent to `int labelRectWidth = resCfg.labelIndent + emojiLabelSize.width + textLabelSize.width + resCfg.labelPadding;` and `labelRectTopRightX` is not used anywhere else.
</blockquote></details>

Done.
___
*[trunk/video-overlay/src/main/cpp/BoundingBoxWriter.cpp, line 600 at r5](https://reviewable.io/reviews/openmpf/openmpf/1255#-MVatUUjCurHkunkKF_I:-MWPth9y1oERn2zgllsH:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/f5392eb02631fe52099b51b7f6e994ac8467ee33/trunk/video-overlay/src/main/cpp/BoundingBoxWriter.cpp#L600)):*

You are adding (once you distribute the -) back something you just subtracted. This appears to be equivalent to `int labelRectHeight = textLabelSize.height + (2 * resCfg.labelPadding);`. `labelRectBottomLeftY` is not used anywhere else.
</blockquote></details>

Done.
___
*[trunk/video-overlay/src/main/cpp/BoundingBoxWriter.cpp, line 670 at r5](https://reviewable.io/reviews/openmpf/openmpf/1255#-MVqKkd_1HmSCWvY1MU_:-MWQ2ERJBILQauCcAZgv:b-vaa6z4) ([raw file](https://github.com/openmpf/openmpf/blob/f5392eb02631fe52099b51b7f6e994ac8467ee33/trunk/video-overlay/src/main/cpp/BoundingBoxWriter.cpp#L670)):*

Doesn't this result in certain labels not showing up? Shouldn't we draw the visible part? I think if we do `cv::Rect(cv::Point(0, 0), image.size()) & addedLabelMatInsertRect` we'll get just the region that is in the image.
</blockquote></details>

Yes, when this error occurs no part of the label will show up. This is because`auto insertionRegion = image(paddedLabelMatInsertRect);` will generate an exception when the `paddedLabelMatInsertRect` extends even one pixel beyond the bounds of the image.

I implemented your idea. While doing so, I noticed that `insertionRegion.forEach()` will generate an exception when `insertionRegion.empty()`, instead of doing nothing, so I avoid that.

An exception is no longer anticipated or acceptable, so I removed the try/catch.

Also, I removed the part of the code where I was doubling the frame boundary. I was only doing that to provide more space for the entire padded label Mat to fit within the padded image.
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/markup/MarkupSplitter.java, line 218 at r5](https://reviewable.io/reviews/openmpf/openmpf/1255#-MVXLg9r31jYH8ZbRRqu:-MWQkTd96tAUBkOn0ICn:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/f5392eb02631fe52099b51b7f6e994ac8467ee33/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/markup/MarkupSplitter.java#L218)):*

This check is not needed. `Boolean.parseBoolean()` handles nulls.
</blockquote></details>

Done.
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/markup/MarkupSplitter.java, line 231 at r5](https://reviewable.io/reviews/openmpf/openmpf/1255#-MVXM8Ow9KGp5VXTLy7p:-MWQkwh26T0mYfMO8w8B:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/f5392eb02631fe52099b51b7f6e994ac8467ee33/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/markup/MarkupSplitter.java#L231)):*

This should use `Boolean.parseBoolean()` like the others. It will allow you to remove the `detection.getDetectionProperties().containsKey("FILLED_GAP")` check
</blockquote></details>

Done.
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/markup/MarkupSplitter.java, line 241 at r5](https://reviewable.io/reviews/openmpf/openmpf/1255#-MVXN2LM5Rq_fZnONBZm:-MWQl90r0WMS7IuQegYp:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/f5392eb02631fe52099b51b7f6e994ac8467ee33/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/markup/MarkupSplitter.java#L241)):*

This check is not needed. `Boolean.parseBoolean()` handles nulls.
</blockquote></details>

Done.
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/markup/MarkupSplitter.java, line 383 at r5](https://reviewable.io/reviews/openmpf/openmpf/1255#-MVXVIFLDeaWomMaA6hm:-MWQmlT42bqRRgzZvy31:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/f5392eb02631fe52099b51b7f6e994ac8467ee33/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/markup/MarkupSplitter.java#L383)):*

None of `.containsKey` checks in this method appear to be necessary. 
</blockquote></details>

Done.
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/markup/MarkupSplitter.java, line 402 at r5](https://reviewable.io/reviews/openmpf/openmpf/1255#-MVXOxU71I62-8I7R1ku:-MWQmCmZBKSmdlqKTXBB:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/f5392eb02631fe52099b51b7f6e994ac8467ee33/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/markup/MarkupSplitter.java#L402)):*

This check is not needed. `String textStr = detection.getDetectionProperties().get(textProp);` is the same thing as what you have here.
</blockquote></details>

Done.
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/markup/MarkupSplitter.java, line 406 at r5](https://reviewable.io/reviews/openmpf/openmpf/1255#-MVXPcSc5zVk87cvkPxE:-MWQmDEn16ZryM12kjD8:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/f5392eb02631fe52099b51b7f6e994ac8467ee33/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/markup/MarkupSplitter.java#L406)):*

This check is not needed. `String numericStr = detection.getDetectionProperties().get(numericProp);` is the same thing as what you have here.
</blockquote></details>

Done.
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/markup/MarkupSplitter.java, line 409 at r5](https://reviewable.io/reviews/openmpf/openmpf/1255#-MVXTo6W8FeXigYDoad3:-MWQmTdw81uudl_e58Oz:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/f5392eb02631fe52099b51b7f6e994ac8467ee33/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/markup/MarkupSplitter.java#L409)):*

I think we should use `.equalsIgnoreCase` when comparing against "CONFIDENCE".
</blockquote></details>

Done.
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/service/WorkflowPropertyService.java, line 52 at r5](https://reviewable.io/reviews/openmpf/openmpf/1255#-MVXaD27BM2EQ4i-hO_A:-MWQmXfL8XZwB7VM8Xja:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/f5392eb02631fe52099b51b7f6e994ac8467ee33/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/service/WorkflowPropertyService.java#L52)):*

This is unused
</blockquote></details>

Done.
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/service/WorkflowPropertyService.java, line 66 at r5](https://reviewable.io/reviews/openmpf/openmpf/1255#-MVXahxl97ptJ194Fhjh:-MWQmkQ575h1vKxnbHdG:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/f5392eb02631fe52099b51b7f6e994ac8467ee33/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/service/WorkflowPropertyService.java#L66)):*

This can be declared `protected`
</blockquote></details>

Done.
___
*[trunk/workflow-manager/src/main/resources/markup-properties.json, line 80 at r5](https://reviewable.io/reviews/openmpf/openmpf/1255#-MVXJKy4CSuZZTFg_ctr:-MWPxR3h7gu1zpBNzlzM:b-nozbe9) ([raw file](https://github.com/openmpf/openmpf/blob/f5392eb02631fe52099b51b7f6e994ac8467ee33/trunk/workflow-manager/src/main/resources/markup-properties.json#L80)):*

I'm pretty sure these would typically be called codecs and ffmpeg would be the encoder.
</blockquote></details>

I think "codec" is less descriptive since it can describe either encoding or decoding behavior. I want to make it clear that we're only talking about encoding.

On the FFmpeg codecs page they list [video encoders](https://ffmpeg.org/ffmpeg-codecs.html#toc-Video-Encoders). "mjpeg" is called a "Motion JPEG encoder". "x264 "H.264/MPEG-4 AVC encoder wrapper".

Honestly, I think the terms can be used interchangeably in this case.

jrobble (commented):
> Reviewed 1 of 1 files at r7.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1255)* status: all files reviewed, 25 unresolved discussions (waiting on @brosenberg42)

brosenberg42 (approved):
> Reviewed 4 of 5 files at r6.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1255)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jrobble)

brosenberg42 (approved):
> Reviewed 1 of 1 files at r7.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1255)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jrobble)

jrobble (commented):
> Reviewed 1 of 1 files at r8.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1255)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jrobble)

Issue #1254, opened 2021-01-27
Jobs table flickers when polling is enabled and the search box is used.


brosenberg42:
From the user: "something is still weird with the search in the ui, when i search something it returns results, reverts to all, returns results, etc. about 3 times"

I also saw this when polling on the jobs page was enabled and there were many complete jobs. As I was typing in the query and after I finished, I noticed that there were pending AJAX requests from before I started typing. I think we are seeing those old requests from before the query was entered are being processed before the request with the query.

As part of this task we should change the polling behavior. Currently, it is possible for there to be multiple polling requests active at the same time. This is because we don't wait for the request to complete before waiting for the specified interval. We should wait for the previous polling request to complete before waiting for the specified interval.

Issue #1237, opened 2020-12-21
jQuery DataTables bug: "int parameter 'draw' is present but cannot be translated into a null value"


jrobble:
From the end user

> I'm continually getting this error in the ui:
>
>`DataTables warning: table id=jobTable - Ajax error. For more information about this error, please see http://datatables.net/tn/7` 

In the WFM log:

```
openmpf5_workflow-manager.1.jrc0iukayhln@some.host    | 2020-12-16 21:58:11,958 ERROR [http-nio-8080-exec-103] o.m.m.m.ControllerUncaughtExceptionHandler - Request for http://some.swarm.host/workflow-manager/jobs-paged raised an uncaught exception
openmpf5_workflow-manager.1.jrc0iukayhln@some.host    | java.lang.IllegalStateException: Optional int parameter 'draw' is present but cannot be translated into a null value due to being declared as a primitive type. Consider declaring it as object wrapper for the corresponding primitive type.
```

This error seems to be occurring when using the Job Status web UI because of a POST to http://openmpf-workflow-manager.swarm.va.dcap.fbi.gov/workflow-manager/jobs-paged where the “draw” parameter is present has no value. I’ve read that Spring MVC will attempt to set undefined parameters to “null”. In our REST endpoint code it seems that we attempt set a primitive int variable to that “null”, which is not valid. Some people online recommend using the Integer class instead of the primitive int.
 
Long story short, it seems that we can handle parameters a little better in our code, although, I’m not sure why the “draw” parameter would ever be set to null. We’re using the jQuery DataTables library: https://datatables.net/manual/server-side#Sent-parameters.

PR #1344, opened 2021-06-09
Fix Datatables bugs

brosenberg42:
**Issues:**

- https://github.com/openmpf/openmpf/issues/1237
- https://github.com/openmpf/openmpf/issues/1254

### DIFF
--- a/trunk/mpf-rest-api/src/main/java/org/mitre/mpf/rest/api/JobPageListModel.java
+++ b/trunk/mpf-rest-api/src/main/java/org/mitre/mpf/rest/api/JobPageListModel.java
@@ -31,14 +31,15 @@ import java.util.List;
 
 public class JobPageListModel {
 	private List<JobPageModel> data = new ArrayList<JobPageModel>();
-	private String draw;
+	private Integer draw;
 	private int recordsTotal = 0;
 	private int  recordsFiltered = 0;
 	private String error  = null;
 
 	public JobPageListModel() {}
 
-	public JobPageListModel(String draw, int recordsTotal, int  recordsFiltered,String error,List<JobPageModel> jobs) {
+	public JobPageListModel(Integer draw, int recordsTotal, int recordsFiltered,String error,
+	                        List<JobPageModel> jobs) {
 		this.draw = draw;
 		this.recordsTotal = recordsTotal;
 		this.recordsFiltered = recordsFiltered;
@@ -58,11 +59,11 @@ public class JobPageListModel {
 		this.data = data;
 	}
 
-	public String getDraw() {
+	public Integer getDraw() {
 		return draw;
 	}
 
-	public void setDraw(String draw) {
+	public void setDraw(Integer draw) {
 		this.draw = draw;
 	}
 
--- a/trunk/workflow-manager/src/main/java/org/mitre/mpf/mvc/controller/JobController.java
+++ b/trunk/workflow-manager/src/main/java/org/mitre/mpf/mvc/controller/JobController.java
@@ -38,6 +38,7 @@ import org.mitre.mpf.wfm.data.access.JobRequestDao;
 import org.mitre.mpf.wfm.data.entities.persistent.BatchJob;
 import org.mitre.mpf.wfm.data.entities.persistent.JobRequest;
 import org.mitre.mpf.wfm.data.entities.persistent.Media;
+import org.mitre.mpf.wfm.enums.BatchJobStatusType;
 import org.mitre.mpf.wfm.event.JobProgress;
 import org.mitre.mpf.wfm.service.S3StorageBackend;
 import org.mitre.mpf.wfm.service.StorageException;
@@ -63,6 +64,8 @@ import java.nio.file.Files;
 import java.nio.file.NoSuchFileException;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
+import java.util.stream.Stream;
 
 import static java.util.stream.Collectors.toList;
 
@@ -163,15 +166,12 @@ public class JobController {
     @ResponseBody
     public List<SingleJobInfo> getJobStatus(@RequestParam(value = "useSession", required = false) boolean useSession) {
         if (useSession) {
-            return sessionModel.getSessionJobs().stream()
-                    .map(id -> convertJob(jobRequestDao.findById(id)))
-                    .collect(toList());
+            return convertJobs(sessionModel.getSessionJobs()
+                                       .stream()
+                                       .map(jobRequestDao::findById));
         }
         else {
-            return jobRequestDao.findAll()
-                    .stream()
-                    .map(this::convertJob)
-                    .collect(toList());
+            return convertJobs(jobRequestDao.findAll().stream());
         }
     }
 
@@ -192,7 +192,7 @@ public class JobController {
     @RequestMapping(value = {"/jobs-paged"}, method = RequestMethod.POST)
     @ResponseBody
     public JobPageListModel getJobStatusFiltered(
-            @RequestParam(value = "draw", required = false) int draw,
+            @RequestParam(value = "draw", required = false) Integer draw,
             @RequestParam(value = "start", required = false) int start,
             @RequestParam(value = "length", required = false) int length,
             @RequestParam(value = "search", required = false) String search,
@@ -204,14 +204,15 @@ public class JobController {
         String sortOrderDirection = orderDirection.equals("desc") ? orderDirection : "asc";
 
         //handle paging
-        List<SingleJobInfo> jobInfoModels = jobRequestDao
-                .findByPage(length, start, search, sortColumn, sortOrderDirection)
-                .stream()
-                .map(this::convertJob)
-                .collect(toList());
+        var jobInfoModels = convertJobs(
+                jobRequestDao
+                        .findByPage(length, start, search, sortColumn, sortOrderDirection)
+                        .stream());
+
         int recordsTotal = (int) jobRequestDao.countAll();
 
         JobPageListModel model = new JobPageListModel();
+        model.setDraw(draw);
         // Total records, before filtering (i.e. the total number of records in the database)
         model.setRecordsTotal(recordsTotal);
         // Total records, after filtering (i.e. the total number of records after filtering has been applied -
@@ -248,12 +249,10 @@ public class JobController {
             @ApiResponse(code = 400, message = "Invalid id") })
     @ResponseBody
     public ResponseEntity<SingleJobInfo> getJobStatusRest(@ApiParam(required = true, value = "Job Id") @PathVariable("id") long jobId) {
-        JobRequest jobRequest = jobRequestDao.findById(jobId);
-        if (jobRequest == null) {
-            log.error("getJobStatusRest: Error retrieving the SingleJobInfo model for the job with id '{}'", jobId);
-            return ResponseEntity.badRequest().body(null);
-        }
-        return ResponseEntity.ok(convertJob(jobRequest));
+        return Optional.ofNullable(jobRequestDao.findById(jobId))
+                .flatMap(this::convertJob)
+                .map(ResponseEntity::ok)
+                .orElseGet(() -> ResponseEntity.badRequest().body(null));
     }
 
     //INTERNAL
@@ -265,12 +264,9 @@ public class JobController {
             return null;
         }
 
-        JobRequest jobRequest = jobRequestDao.findById(jobId);
-        if (jobRequest == null) {
-            log.error("getJobStatus: Error retrieving the SingleJobInfo model for the job with id '{}'", jobId);
-            return null;
-        }
-        return convertJob(jobRequest);
+        return Optional.ofNullable(jobRequestDao.findById(jobId))
+                .flatMap(this::convertJob)
+                .orElse(null);
     }
 
     /*
@@ -436,19 +432,47 @@ public class JobController {
     }
 
 
-    private SingleJobInfo convertJob(JobRequest job) {
+    private Optional<SingleJobInfo> convertJob(JobRequest job) {
         float jobProgressVal = jobProgress.getJobProgress(job.getId())
                 .orElseGet(() -> job.getStatus().isTerminal() ? 100 : 0.0f);
 
-        var batchJob = jsonUtils.deserialize(job.getJob(), BatchJob.class);
-        var mediaUris = batchJob.getMedia().stream()
-                .map(Media::getUri)
-                .collect(toList());
+        List<String> mediaUris;
+        try {
+            if (job.getJob() == null) {
+                if (job.getStatus() == BatchJobStatusType.INITIALIZED) {
+                    return Optional.empty();
+                }
+                else {
+                    mediaUris = List.of();
+                }
+            }
+            else {
+                var batchJob = jsonUtils.deserialize(job.getJob(), BatchJob.class);
+                mediaUris = batchJob.getMedia().stream()
+                        .map(Media::getUri)
+                        .collect(toList());
+            }
 
-        return new SingleJobInfo(
+        }
+        catch (WfmProcessingException e) {
+            log.error(String.format(
+                    "Unable to determine mediaUris for job %s due to: %s", job.getId(), e), e);
+            mediaUris = List.of();
+        }
+
+        return Optional.of(new SingleJobInfo(
                 job.getId(), job.getPipeline(), job.getPriority(), job.getStatus().toString(), jobProgressVal,
                 job.getTimeReceived(), job.getTimeCompleted(), job.getOutputObjectPath(),
-                job.getStatus().isTerminal(), mediaUris);
+                job.getStatus().isTerminal(), mediaUris));
+    }
+
+
+    private List<SingleJobInfo> convertJobs(Stream<JobRequest> jobRequests) {
+        return jobRequests
+                .map(this::convertJob)
+                .filter(Optional::isPresent)
+                .map(Optional::get)
+                .collect(toList());
     }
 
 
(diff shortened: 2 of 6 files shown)


jrobble (changes requested):
> Reviewed 6 of 6 files at r1.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1344)* status: all files reviewed, 5 unresolved discussions (waiting on @brosenberg42)

___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/mvc/controller/JobController.java, line 253 at r1](https://reviewable.io/reviews/openmpf/openmpf/1344#-MbmF4OC18TKmekUi3sO:-MbmF4OC18TKmekUi3sP:bvldiqk) ([raw file](https://github.com/openmpf/openmpf/blob/f3ade59f4fc712e64ae928b46b90c9a399bbc041/trunk/workflow-manager/src/main/java/org/mitre/mpf/mvc/controller/JobController.java#L253)):*
> ```Java
>         JobRequest jobRequest = jobRequestDao.findById(jobId);
>         if (jobRequest == null) {
>             log.error("getJobStatusRest: Error retrieving the SingleJobInfo model for the job with id '{}'", jobId);
> ```

I think we should keep this error message here and in the method below.
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/mvc/controller/JobController.java, line 441 at r1](https://reviewable.io/reviews/openmpf/openmpf/1344#-MbmUmQg5wKCAF-2lQMU:-MbmUmQg5wKCAF-2lQMV:b-d582b5) ([raw file](https://github.com/openmpf/openmpf/blob/ffef24e61761db20569f34b940884eb9adff4192/trunk/workflow-manager/src/main/java/org/mitre/mpf/mvc/controller/JobController.java#L441)):*
> ```Java
>         List<String> mediaUris;
>         try {
>             if (job.getJob() == null) {
> ```

For testing purposes, I created a job and force-killed the WFM right after we write the job to the database, but before we call `setJob()` on it with the JSON.

I was able to create a NPE by executing the [POST] `/rest/jobs/{id}/resubmit` endpoint with that job id.

Also, I was able to create a NPE by clicking the "Media" button under "Actions" for that job on the Job Status page, which executes the [POST] `/markup/get-markup-results-filtered` endpoint.

Even if I could click the "Media" button, attempting to download the source media would also result in the NPE when the [POST] `/server/download` endpoint is executed.

These three actions all result in calls to `deserialize()` with a null parameter.

---

Instead of addressing each of these one by one, I propose we prevent the problem from happening by persisting the complete job to the database the first time we write it.

To do this, we first need to query the database for the sequence id, create the job object, and then persist it. I found [this post](https://stackoverflow.com/questions/6386888/get-next-sequence-value-from-database-using-hibernate/23267212#23267212) and tried out the code for myself.

I got it to work and pushed [this branch](https://github.com/openmpf/openmpf/compare/hotfix/datatables-bug...hotfix/datatables-bug-jrobble) for you to look at. I dropped the `getNextId()` call in `WfmStartup` for no reason other than it was a convenient place for testing. That call is not a peek. It will get you the next id and increment the sequence in the database.

To be clear, I think we should no longer use `@GeneratedValue` for the `id` in `JobRequest`, and instead populate the id value ourself using `getNextId()`.
___
*[trunk/workflow-manager/src/main/webapp/resources/mytheme/js/angular_app/services.js, line 695 at r1](https://reviewable.io/reviews/openmpf/openmpf/1344#-MbnqSaw7u1Y2G_h3WXN:-MbnqSax-gaJR8S2nqwE:b-9046el) ([raw file](https://github.com/openmpf/openmpf/blob/ffef24e61761db20569f34b940884eb9adff4192/trunk/workflow-manager/src/main/webapp/resources/mytheme/js/angular_app/services.js#L695)):*
> ```JavaScript
> 
>     var registrants = {},
>         internalInterval = 60000;
> ```

To confirm, this just controls the frequency that the stats page updates, right?

Looking at `AdminStatsCtrl.js` I see `var default_timer_counter = 15`. It seems that the timer counter decrements every interval, and since the interval is now one minute we only update that stats page every 15 minutes. That seems like a long time, but fine since a user can just manually refresh.

Please update the "Next refresh in N" to say "Next refresh in N minutes" for clarity.

Also, it will say "Next refresh in 0" and stay there for a minute. That's strange.
___
*[trunk/workflow-manager/src/main/webapp/resources/mytheme/js/controllers/JobsCtrl.js, line 158 at r1](https://reviewable.io/reviews/openmpf/openmpf/1344#-MbnpDeM5-ZXnLxzVd7-:-MbnpDeM5-ZXnLxzVd70:bib46m7) ([raw file](https://github.com/openmpf/openmpf/blob/ffef24e61761db20569f34b940884eb9adff4192/trunk/workflow-manager/src/main/webapp/resources/mytheme/js/controllers/JobsCtrl.js#L158)):*
> ```JavaScript
>             jobTable.on('error.dt', function(){
>                 console.error(
>                     "The most recent attempt to update the jobs table failed: %o",
> ```

The "Last checked at N" message is not updated on the Job Status page when the error banner is up and periodic updates are enabled (broadcasts are disabled).
___
*[trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/businessrules/TestJobRequestService.java, line 1 at r1](https://reviewable.io/reviews/openmpf/openmpf/1344#-MbnpTOp33foIpJeYKo4:-MbnpTOp33foIpJeYKo5:b-xe3ayv) ([raw file](https://github.com/openmpf/openmpf/blob/ffef24e61761db20569f34b940884eb9adff4192/trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/businessrules/TestJobRequestService.java#L1)):*
> ```Java
> package org.mitre.mpf.wfm.businessrules; /******************************************************************************
> ```

Please move the package below the header comment blocks.

brosenberg42 (commented):
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1344)* status: 2 of 12 files reviewed, 5 unresolved discussions (waiting on @brosenberg42 and @jrobble)

___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/mvc/controller/JobController.java, line 253 at r1](https://reviewable.io/reviews/openmpf/openmpf/1344#-MbmF4OC18TKmekUi3sO:-MbpuMOk7wBauVQaY1Pz:b-d6vi9w) ([raw file](https://github.com/openmpf/openmpf/blob/f3ade59f4fc712e64ae928b46b90c9a399bbc041/trunk/workflow-manager/src/main/java/org/mitre/mpf/mvc/controller/JobController.java#L253)):*

I think we should keep this error message here and in the method below.
</blockquote></details>

If I agreed, I wouldn't have removed it in the first place. The return value of the function already indicates that there was an error and the log message provides no additional information. Code clarity is my number 1 priority, so I can't justify keeping something reduces clarity and provides no value. Sometimes the best way to improve code is to remove useless parts. 
When considering my changes to `convertJob` that changes a 4 line method in to a 13 line method:
```java
    public ResponseEntity<SingleJobInfo> getJobStatusRest(@ApiParam(required = true, value = "Job Id") @PathVariable("id") long jobId) {
        var job = jobRequestDao.findById(jobId);
        if (job == null) {
            log.error("... ");
            return ResponseEntity.badRequest().body(null);
        }

        var optConvertedJob = convertJob(job);
        if (optConvertedJob.isPresent()) {
            return ResponseEntity.ok(optConvertedJob.get());
        }
        else {
            log.error("... ");
            return ResponseEntity.badRequest().body(null);
        }
    }
```
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/mvc/controller/JobController.java, line 441 at r1](https://reviewable.io/reviews/openmpf/openmpf/1344#-MbmUmQg5wKCAF-2lQMU:-MbqGX2A24PmTJ8ZRQ5b:bteztir) ([raw file](https://github.com/openmpf/openmpf/blob/ffef24e61761db20569f34b940884eb9adff4192/trunk/workflow-manager/src/main/java/org/mitre/mpf/mvc/controller/JobController.java#L441)):*

For testing purposes, I created a job and force-killed the WFM right after we write the job to the database, but before we call `setJob()` on it with the JSON.

I was able to create a NPE by executing the [POST] `/rest/jobs/{id}/resubmit` endpoint with that job id.

Also, I was able to create a NPE by clicking the "Media" button under "Actions" for that job on the Job Status page, which executes the [POST] `/markup/get-markup-results-filtered` endpoint.

Even if I could click the "Media" button, attempting to download the source media would also result in the NPE when the [POST] `/server/download` endpoint is executed.

These three actions all result in calls to `deserialize()` with a null parameter.

---

Instead of addressing each of these one by one, I propose we prevent the problem from happening by persisting the complete job to the database the first time we write it.

To do this, we first need to query the database for the sequence id, create the job object, and then persist it. I found [this post](https://stackoverflow.com/questions/6386888/get-next-sequence-value-from-database-using-hibernate/23267212#23267212) and tried out the code for myself.

I got it to work and pushed [this branch](https://github.com/openmpf/openmpf/compare/hotfix/datatables-bug...hotfix/datatables-bug-jrobble) for you to look at. I dropped the `getNextId()` call in `WfmStartup` for no reason other than it was a convenient place for testing. That call is not a peek. It will get you the next id and increment the sequence in the database.

To be clear, I think we should no longer use `@GeneratedValue` for the `id` in `JobRequest`, and instead populate the id value ourself using `getNextId()`.
</blockquote></details>

Good find. It does seem overly complicated, but that is hibernate's fault.
___
*[trunk/workflow-manager/src/main/webapp/resources/mytheme/js/angular_app/services.js, line 695 at r1](https://reviewable.io/reviews/openmpf/openmpf/1344#-MbnqSaw7u1Y2G_h3WXN:-Mbpyq1CDxY4FOBckJU_:b868pd6) ([raw file](https://github.com/openmpf/openmpf/blob/ffef24e61761db20569f34b940884eb9adff4192/trunk/workflow-manager/src/main/webapp/resources/mytheme/js/angular_app/services.js#L695)):*

To confirm, this just controls the frequency that the stats page updates, right?

Looking at `AdminStatsCtrl.js` I see `var default_timer_counter = 15`. It seems that the timer counter decrements every interval, and since the interval is now one minute we only update that stats page every 15 minutes. That seems like a long time, but fine since a user can just manually refresh.

Please update the "Next refresh in N" to say "Next refresh in N minutes" for clarity.

Also, it will say "Next refresh in 0" and stay there for a minute. That's strange.
</blockquote></details>

This was the wrong thing. I meant to change:
```javascript
	$interval( function() {
		MetadataService.getMetadataNoCache()
	}, 2000);
```
because it is causing a ton of network activity.
___
*[trunk/workflow-manager/src/main/webapp/resources/mytheme/js/controllers/JobsCtrl.js, line 158 at r1](https://reviewable.io/reviews/openmpf/openmpf/1344#-MbnpDeM5-ZXnLxzVd7-:-MbqRalJ6eJkK0iCOj6B:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/ffef24e61761db20569f34b940884eb9adff4192/trunk/workflow-manager/src/main/webapp/resources/mytheme/js/controllers/JobsCtrl.js#L158)):*

The "Last checked at N" message is not updated on the Job Status page when the error banner is up and periodic updates are enabled (broadcasts are disabled).
</blockquote></details>

Done.
___
*[trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/businessrules/TestJobRequestService.java, line 1 at r1](https://reviewable.io/reviews/openmpf/openmpf/1344#-MbnpTOp33foIpJeYKo4:-MbqBmMvEdhFPPCHnQR9:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/ffef24e61761db20569f34b940884eb9adff4192/trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/businessrules/TestJobRequestService.java#L1)):*

Please move the package below the header comment blocks.
</blockquote></details>

Done.

jrobble (changes requested):
> Reviewed 1 of 3 files at r2, 9 of 9 files at r3.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1344)* status: all files reviewed, 4 unresolved discussions (waiting on @brosenberg42)

___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/mvc/controller/JobController.java, line 253 at r1](https://reviewable.io/reviews/openmpf/openmpf/1344#-MbmF4OC18TKmekUi3sO:-Mbq_hQx-X1uUA2dNOYZ:b4yjhfx) ([raw file](https://github.com/openmpf/openmpf/blob/f3ade59f4fc712e64ae928b46b90c9a399bbc041/trunk/workflow-manager/src/main/java/org/mitre/mpf/mvc/controller/JobController.java#L253)):*

If I agreed, I wouldn't have removed it in the first place. The return value of the function already indicates that there was an error and the log message provides no additional information. Code clarity is my number 1 priority, so I can't justify keeping something reduces clarity and provides no value. Sometimes the best way to improve code is to remove useless parts. 
When considering my changes to `convertJob` that changes a 4 line method in to a 13 line method:
```java
    public ResponseEntity<SingleJobInfo> getJobStatusRest(@ApiParam(required = true, value = "Job Id") @PathVariable("id") long jobId) {
        var job = jobRequestDao.findById(jobId);
        if (job == null) {
            log.error("... ");
            return ResponseEntity.badRequest().body(null);
        }

        var optConvertedJob = convertJob(job);
        if (optConvertedJob.isPresent()) {
            return ResponseEntity.ok(optConvertedJob.get());
        }
        else {
            log.error("... ");
            return ResponseEntity.badRequest().body(null);
        }
    }
```
</blockquote></details>

I do like the brevity of the new code. 

My thought is that the person looking at the WFM logs may not have access to the logs of the system which executed the endpoint, or have any indication about the cause of the underlying issue - only that the system which made the call has some kind of error.

Really, it's the responsibility of the system which made the call to interpret the 400 error code for this endpoint correctly, which we have documented as "Invalid id". I guess it boils down to that I don't trust that information will somehow make its way to the end user and they will blame the WFM. They might not know what error code was returned.

Of course I'm probably just being paranoid, so I'm not going to push this anymore.
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/mvc/controller/JobController.java, line 441 at r3](https://reviewable.io/reviews/openmpf/openmpf/1344#-MbqxIgwCqqwmLWwgARz:-MbqxIgx2PRdMOvRQwNM:b-h2eb3f) ([raw file](https://github.com/openmpf/openmpf/blob/1ba2ba6b85519df8feba8170290abcb2a21acf81/trunk/workflow-manager/src/main/java/org/mitre/mpf/mvc/controller/JobController.java#L441)):*
> ```Java
>         List<String> mediaUris;
>         try {
>             if (job.getJob() == null) {
> ```

Can we remove this check now that the job data will always be set when the `JobRequest` is persisted the first time?
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/businessrules/impl/JobRequestServiceImpl.java, line 251 at r3](https://reviewable.io/reviews/openmpf/openmpf/1344#-MbqfvD7AGkjXN5hLddp:-MbqfvD7AGkjXN5hLddq:bapch0b) ([raw file](https://github.com/openmpf/openmpf/blob/1ba2ba6b85519df8feba8170290abcb2a21acf81/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/businessrules/impl/JobRequestServiceImpl.java#L251)):*
> ```Java
>             return jobRequestEntity;
>         }
>         catch (Exception e) {
> ```

Are you thinking that the only way an Exception will be generated is due to a database issue and that it doesn't make sense to set the status as `JOB_CREATION_ERROR` and try to persist to the database again?
___
*[trunk/workflow-manager/src/main/webapp/resources/mytheme/js/controllers/JobsCtrl.js, line 157 at r3](https://reviewable.io/reviews/openmpf/openmpf/1344#-MbqqUw91TsIlH0EB2tm:-MbqqUw91TsIlH0EB2tn:b-ld7f53) ([raw file](https://github.com/openmpf/openmpf/blob/1ba2ba6b85519df8feba8170290abcb2a21acf81/trunk/workflow-manager/src/main/webapp/resources/mytheme/js/controllers/JobsCtrl.js#L157)):*
> ```JavaScript
>             });
>             jobTable.on('error.dt', function(){
>                 tableLastUpdate = moment();
> ```

The "Last checked at" message is updated inside of `pollingUpdate()` only on a successful `jobTable.ajax.reload()`. If the jobs table fails to update then the value in the UI does not change. I'm asking you to please make sure that message changes to reflect that the page tried checking again despite running into another error.

I'm testing your changes using this debug code to produce a random failure:

```
    // INTERNAL
    // Parameters come from DataTables library: https://datatables.net/manual/server-side
    @RequestMapping(value = {"/jobs-paged"}, method = RequestMethod.POST)
    @ResponseBody
    public JobPageListModel getJobStatusFiltered(
            @RequestParam(value = "draw", required = false) Integer draw,
            @RequestParam(value = "start", required = false) int start,
            @RequestParam(value = "length", required = false) int length,
            @RequestParam(value = "search", required = false) String search,
            @RequestParam(value = "order[0][column]", defaultValue = "0") String orderByColumn,
            @RequestParam(value = "order[0][dir]", defaultValue = "desc") String orderDirection) {

        Random rd = new Random();
        if (rd.nextBoolean()) {
            throw new RuntimeException("BAD THING!"); // DEBUG
        }
```

brosenberg42:
___
*[trunk/workflow-manager/src/main/webapp/WEB-INF/html/index.html, line 81 at r4](https://reviewable.io/reviews/openmpf/openmpf/1344#-MbrJyxn0mLxq-2KxXou:-MbrJyxn0mLxq-2KxXov:b-pbpy3h) ([raw file](https://github.com/openmpf/openmpf/blob/f3ade59f4fc712e64ae928b46b90c9a399bbc041/trunk/workflow-manager/src/main/webapp/WEB-INF/html/index.html#L81)):*
> ```HTML
> 
>  
>

jrobble (approved):
> Reviewed 4 of 4 files at r4.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1344)* status: all files reviewed, 1 unresolved discussion (waiting on @brosenberg42)

Issue #1349, opened 2021-06-11
Improve reporting and recording job status


jrobble:
In general, every time the WFM updates the job status it should be persisted to the database as well as broadcast to the web UI. Currently, that is not done consistently and will result in undesirable behavior:

- Refreshing the web UI may show a different job status. This may be because what's currently shown is based on a broadcast while refreshing the page results in pulling the status from the database.

- Retrieving the job status through the REST endpoint may result in a different status than what's shown in the web UI. Again, this is because what's shown in the web UI is based on a broadcast but the REST endpoint gets the status from the database.

- Some job states are never shown in the web UI. In particular, `BUILDING_OUTPUT_OBJECT`.
    - **Update 6/15/2021:** We should remove the `BUILDING_OUTPUT_OBJECT` job status since it is short-lived and does not indicate if warnings or errors have occurred.

Related issues:

- The web UI does not show `IN_PROGRESS_WARNINGS` or `IN_PROGRESS_ERRORS`. It will instead show `IN_PROGRESS` until the job is complete and then show `COMPLETE_WITH_WARNINGS` and `COMPLETE_WITH_ERRORS`.
    - We should update the job status as soon as the WFM adds a warning or error message to a job. Related to https://github.com/openmpf/openmpf/issues/1151.

- Related to the above item, whenever the job has a warning or error, the job status label in the web UI should be orange or red, respectively.

- When the job status is `ERROR`, the JSON output object does not have any messages in the  `errors` field. We should ensure that a meaningful error message is added whenever we set the job status to `ERROR`.

- When resubmitting a job by clicking the `Resubmit` button in the web UI, the job status label will retain the color of the first job run. It should be reset to the neutral grey. Also, the `Start Date` will retain the date of the first job run until the job completes. If you refresh the web UI while the job is in progress the label color and `Start Date` will be correct.
 
- https://github.com/openmpf/openmpf/issues/1106 

- https://github.com/openmpf/openmpf/issues/1080

Issue #1151, opened 2020-08-07
Support IN_PROGRESS_WITH_WARNINGS status


jrobble:
More work needs to be done to get the status in the web UI to show `IN_PROGRESS_WITH_WARNINGS` properly (for failure to skip media inspection and detection padding overshrink). Also, the `[GET] /rest/jobs/{id}` endpoint shows `IN_PROGRESS` even after I add a warning for not being able to skip media inspection. I tested that by having a FACE --> MARKUP pipeline, not running a markup component, and getting the status after the first stage of the pipeline.

Similarly, ensure that the web UI and REST endpoint show `IN_PROGRESS_WITH_ERRORS` when an error is added.

Consider refactoring the WFM so that the job status is automatically set whenever an error or warning is added, and automatically broadcast when the status changes. 

Related to https://github.com/openmpf/openmpf/issues/1349.

Issue #1106, opened 2020-06-02
When a job ends in ERROR, the job status UI does not show an End Date


jrobble:
When a job ends in ERROR, the job status UI does not show an End Date. This is odd since the JSON output object has a valid `timeStop` field.

Issue #1080, opened 2020-04-13
Batch jobs prematurely set to 100% completion during artifact extraction


jrobble:
We may have a bug in our code that results in 100% progress  completion during artifact extraction before the progress gets set to 99% before building the output object.

In `BroadcastEnabledStringCountBasedWfmAggregator.java`:

```
int currentStage = 1 + job.getCurrentStage();
int totalStages = job.getPipeline().getStages().size();
float progressPerStage = 1 / (1f * totalStages) * 100f;

float taskProgress = (((float) aggregateCount) / ((float) splitSize));
float jobProgress = (currentStage - 1) * (progressPerStage) + (progressPerStage) * taskProgress;

...

//store the current job progress to prevent progress displaying as zero on manual refreshes
jobProgressStore.setJobProgress(jobId, jobProgress);
```

The `float jobProgress =` line seems as though it can calculate a value of 100%. 

Users have indicated that `[GET] /rest/jobs/{id}` can return a value of 100% when a job seems stuck in artifact extraction. The last line they see in the WFM is:

```
JmsConsumer[MPF.COMPLETED_DETECTIONS]] o.m.m.w.c.WfmSplitter - [Job 135765|*|*] WfmSplitter class: detectionExtractionSplitter|org.mitre.mpf.wfm.camel.operations.detection.artifactextraction.ArtifactExtractionSplitterImpl produced 1 work units with correlation id '135765:18e8bb23-b73a-4411-9fe9-83af2b9c89dd' (error = false).
``` 

Note that in `EndOfStageProcessor.java` we have:

```
if(!job.isOutputEnabled()) {
	jobStatusBroadcaster.broadcast(
			jobId,
			100,
			job.isCancelled() ? BatchJobStatusType.CANCELLED : BatchJobStatusType.COMPLETE,
			Instant.now());
	jobProgressStore.setJobProgress(jobId, 100.0f);
} else {
	jobStatusBroadcaster.broadcast(jobId, 99, BatchJobStatusType.BUILDING_OUTPUT_OBJECT, Instant.now());
	jobProgressStore.setJobProgress(jobId, 99.0f);
}		
```

This may set the progress value to 99%. Note that this code is executed after the code in the above block.

The solution may be to ensure that the first block of code above sets the progress to a max value of 99%.

PR #1364, opened 2021-07-06
Improve job status reporting

brosenberg42:
**Issues:**
- https://github.com/openmpf/openmpf/issues/1151
- https://github.com/openmpf/openmpf/issues/1349
- https://github.com/openmpf/openmpf/issues/1080
- https://github.com/openmpf/openmpf/issues/1106

**Related PRs:**
- https://github.com/openmpf/openmpf/pull/1364

---
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1364)


jrobble (changes requested):
> Reviewed 29 of 29 files at r1.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1364)* status: all files reviewed, 11 unresolved discussions (waiting on @brosenberg42)

*[a discussion](https://reviewable.io/reviews/openmpf/openmpf/1364#-Me6EeW_7zV4rW6Dk8ew:-Me6EeW_7zV4rW6Dk8ex:b-m4o3k3) (no related file):*
A job will appear as `COMPLETE` while TiesDb post / callback attempts are still going on in the background. Attempting to resubmit a job in this state will result in an exception and UI popup with the message:

> Failed to resubmit the job with id '1285'. Job with id 1285 already exists.

This is confusing to an end user, who doesn't know what's going on in the background. Once this happens the UI will stay as `RESUBMITTING` until the UI is refreshed.

I think we need some indication in the UI that there is still something going on. Minimally, disable the Resubmit button, but that alone might be a little confusing. Maybe change the tooltip / hint popover to say "Callbacks in progress" or something.

---

Also, the job will stay in `RESUBMITTING` if the media doesn't exist when you try to resubmit.

---
*[a discussion](https://reviewable.io/reviews/openmpf/openmpf/1364#-Me7vpJM9kw5JYnSquiJ:-Me7vpJN-16gh8zdonVH:b-yxb2oq) (no related file):*
In `DefaultJobErrorHandler.java` we have:

```
log.warn("An error was encountered while processing Job #{}. The Job will be marked as {}.", jobId, BatchJobStatusType.ERROR, throwable);
```

First, I think that should be `log.error()`. Second, where is the job status actually set to `ERROR` in response to this? I only see it set to `ERROR` in two places, and it's not clear that they are related to that message.

---
*[a discussion](https://reviewable.io/reviews/openmpf/openmpf/1364#-Me81918-c4kH1K67Y5a:-Me81918-c4kH1K67Y5b:banhpvt) (no related file):*
I put together this state machine diagram:



Primarily, my concern is that we might accidentally transition to an improper state, like `COMPLETE` or `CANCELLED` to `IN_PROGRESS_WARNINGS` (see other comment). My secondary concern is that even if there aren't many situations where this can happen now, we should protect ourselves from it happening in the future as the WFM gets more complex.

I think we should centralize the logic that validates the transitions between states. I like `JobStatusCalculator.completionStatus()`. I would feel more confident if we had something like that for non-terminal states. Maybe something like:

```
BatchJobStatusType  JobStatusCalculator.validateNextStatus(BatchJobStatusType initialStatus, BatchJobStatusType nextStatus)
```

- If status transition is invalid then log an error and return the initial status.
- This isn't intended to hide bugs in the code. For example, if trying to go from `COMPLETE` to `IN_PROGRESS_WARNINGS` we could return `COMPLETE_WITH_WARNINGS`, but that bug should really be handled in `InProgressBatchJobService.addWarning()` instead. We should treat situations like this as an invalid status transition and log an error.
- In many cases we can't transition out of a terminal status to something else, so return the initial terminal status. For example, if we're in an `ERROR` state we should never go to `COMPLETE_WITH_ERRORS`. Similarly, `CANCELLED` should never change to anything else, except `ERROR`. Everything should be allowed to transition to `ERROR`.

In most cases this should just be a safety net. I'm not asking you to change much of the existing code.

---
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/JobCompleteProcessorImpl.java, line 160 at r1](https://reviewable.io/reviews/openmpf/openmpf/1364#-MeRWKZz9w6iX9jBeEnK:-MeRWKZz9w6iX9jBeEnL:bnct2ox) ([raw file](https://github.com/openmpf/openmpf/blob/d11895adec3f7727d27a60f71aa6930749a43687/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/JobCompleteProcessorImpl.java#L160)):*
In `WfmSplitter` we have:

```
            log.error(String.format("Failed to complete the split operation for Job %s due to an exception.", jobId),
                      exception);
            inProgressJobs.setJobStatus(jobId, BatchJobStatusType.ERROR);
```

We should be adding an error message with the exception there.

As stated in the task:

> When the job status is `ERROR`, the JSON output object does not have any messages in the errors field. We should ensure that a meaningful error message is added whenever we set the job status to `ERROR`.
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/JobCompleteProcessorImpl.java, line 168 at r1](https://reviewable.io/reviews/openmpf/openmpf/1364#-Me6Df7f43DOJ0c5LQsc:-Me6Df7f43DOJ0c5LQsd:b-drdyld) ([raw file](https://github.com/openmpf/openmpf/blob/d11895adec3f7727d27a60f71aa6930749a43687/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/JobCompleteProcessorImpl.java#L168)):*
> ```Java
>             jobRequest.setJob(jsonUtils.serialize(job));
>             jobRequestDao.persist(jobRequest);
>             jobStatusBroadcaster.broadcast(jobId, 100, BatchJobStatusType.COMPLETE);
> ```

Submitting a Sphinx job with an image will result in first transitioning to `COMPLETE` before `COMPLETE_WITH_ERRORS`. While `COMPLETE` is shown in red, I think we should transition to the actual terminal status,`COMPLETE_WITH_ERRORS`, directly.

While debugging I confirmed that at this point in the code `jobStatus` is already set to `COMPLETE_WITH_ERRORS` so we should just use that. Doing so will ensure that what's in the UI, `inProgressBatchJobs`, and the DB is all the same.

If we do this, I think we can replace `setJobStatusNoBroadcast()` here with `setJobStatus()`:
```
inProgressBatchJobs.setJobStatusNoBroadcast(jobId, jobStatus.getValue());
```

In general, I'm wondering if we really need `setJobStatusNoBroadcast()` to be a separate public method or if we can just add the behavior to `setJobStatus()`.

---

Also, a few lines above we have `jobStatus.setValue(BatchJobStatusType.ERROR)`. We don't want to change the status back to `COMPLETE`.
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/artifactextraction/ArtifactExtractionProcessor.java, line 190 at r1](https://reviewable.io/reviews/openmpf/openmpf/1364#-MeRKgwI-YSrzTLc3JwA:-MeRKgwJ98UnwQeTfF7w:bbq4g70) ([raw file](https://github.com/openmpf/openmpf/blob/d11895adec3f7727d27a60f71aa6930749a43687/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/artifactextraction/ArtifactExtractionProcessor.java#L190)):*
> ```Java
>         var missingFrameString = createMissingFramesString(tracks, request)
>                 .map(err -> "Error extracting artifact(s). " + err)
>                 .orElseGet(() -> "Artifact extraction failed to due " + e);
> ```

I think you mean `failed due to: " + e);`
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/InProgressBatchJobsService.java, line 227 at r1](https://reviewable.io/reviews/openmpf/openmpf/1364#-Me6onKJ9jVL8183LnYg:-Me6onKK0m6fwY8y6LvJ:b-pykofb) ([raw file](https://github.com/openmpf/openmpf/blob/d11895adec3f7727d27a60f71aa6930749a43687/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/InProgressBatchJobsService.java#L227)):*
> ```Java
>         var job = getJobImpl(jobId);
>         job.addWarning(mediaId, IssueSources.toString(source), codeString, message);
>         if (job.getStatus() != BatchJobStatusType.IN_PROGRESS_ERRORS) {
> ```

It's not always appropriate to transition to an `IN_PROGRESS` state here.

Looking at this code:

```
        JobStatusCalculator.checkErrorMessages(jsonOutputObject, jobStatus);

        return storageService.store(jsonOutputObject, jobStatus, outputSha); // this may update the job status
```

That first line can set `jobStatus` to `COMPLETE`. `StorageService.store(JsonOutputObject ...)` can call `addJobWarning()`.

We don't want to transition back to `IN_PROGRESS_WARNINGS`.
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/access/JobRequestDao.java, line 45 at r1](https://reviewable.io/reviews/openmpf/openmpf/1364#-Me6m3OtEHIgPsGE3OUl:-Me6m3OtEHIgPsGE3OUm:bscogds) ([raw file](https://github.com/openmpf/openmpf/blob/d11895adec3f7727d27a60f71aa6930749a43687/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/access/JobRequestDao.java#L45)):*
> ```Java
>     public long getNextId();
> 
>     void updateStatus(long jobId, BatchJobStatusType status);
> ```

Add `public` to this.
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/access/hibernate/HibernateUserDaoImpl.java, line 55 at r1](https://reviewable.io/reviews/openmpf/openmpf/1364#-Me6jre6D_mUcolnVwc2:-Me6jre6D_mUcolnVwc3:bvbe4xb) ([raw file](https://github.com/openmpf/openmpf/blob/d11895adec3f7727d27a60f71aa6930749a43687/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/access/hibernate/HibernateUserDaoImpl.java#L55)):*
> ```Java
>                 .stream()
>                 .findFirst()
>                 .orElse(null);
> ```

`MarkupResultDao.findByJobIdAndMediaIndex()` returns an `Optional`. You could do that here instead of `null`. Just a suggestion.
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/enums/BatchJobStatusType.java, line 72 at r1](https://reviewable.io/reviews/openmpf/openmpf/1364#-Me6kGb5A_-95FoqKcoy:-Me6kGb66EcZbK-iT04q:bkcfdhx) ([raw file](https://github.com/openmpf/openmpf/blob/fca53dbea8ee0a4a7df8acc5213b9e87551238d7/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/enums/BatchJobStatusType.java#L72)):*
`BUILDING_OUTPUT_OBJECT` is still mentioned in the following files:
- `jobs/layout.html` in the `Status` column legend when you click the `i` icon.
- `REST-API.html`:

> A string that will be one of the following values: UNKNOWN, INITIALIZED (the job has been initialized but not started), JOB_CREATION_ERROR (indicates that a job was received, but a job could not be created from the contents of the request), IN_PROGRESS, BUILDING_OUTPUT_OBJECT, COMPLETE, CANCELLING, CANCELLED, ERROR

The possible statuses in that list need to be updated. It's missing the warning and error variants.
___
*[trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/camelOps/TestDetectionResponseProcessor.java, line 222 at r1](https://reviewable.io/reviews/openmpf/openmpf/1364#-MeROxdfDzj5JdSG343s:-MeROxdfDzj5JdSG343t:b7dvnqe) ([raw file](https://github.com/openmpf/openmpf/blob/fca53dbea8ee0a4a7df8acc5213b9e87551238d7/trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/camelOps/TestDetectionResponseProcessor.java#L222)):*
> ```Java
> 
>         verify(inProgressJobs, times(1))
>                 .setJobStatus(JOB_ID, BatchJobStatusType.IN_PROGRESS_ERRORS);
> ```

`addDetectionProcessingError` still has the logic for setting the job status:

```
        if (error.getErrorCode().equals(MpfConstants.REQUEST_CANCELLED)) {
            setJobStatus(error.getJobId(), BatchJobStatusType.CANCELLING);
        }
        else {
            setJobStatus(error.getJobId(), BatchJobStatusType.IN_PROGRESS_ERRORS);
        }
```

Why did you remove the test code that verifies that the status was set correctly?

brosenberg42 (commented):
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1364)* status: 20 of 49 files reviewed, 10 unresolved discussions (waiting on @brosenberg42 and @jrobble)

*[a discussion](https://reviewable.io/reviews/openmpf/openmpf/1364#-Me7vpJM9kw5JYnSquiJ:-Me_WyWJ6-yOkBFBtsYe:b-w3c6nz) (no related file):*

In `DefaultJobErrorHandler.java` we have:

```
log.warn("An error was encountered while processing Job #{}. The Job will be marked as {}.", jobId, BatchJobStatusType.ERROR, throwable);
```

First, I think that should be `log.error()`. Second, where is the job status actually set to `ERROR` in response to this? I only see it set to `ERROR` in two places, and it's not clear that they are related to that message.
</blockquote></details>

`DefaultJobErrorHandler` does not appear to be used.

---
*[a discussion](https://reviewable.io/reviews/openmpf/openmpf/1364#-Me81918-c4kH1K67Y5a:-MeaBxX-BMJA3vzf3KIw:bkrcy02) (no related file):*
> I think we should centralize the logic that validates the transitions between states. I like JobStatusCalculator.completionStatus(). I would feel more confident if we had something like that for non-terminal states. Maybe something like: `BatchJobStatusType  JobStatusCalculator.validateNextStatus(BatchJobStatusType initialStatus, BatchJobStatusType nextStatus)`

I think the reason we would need to validate the state transitions is that this is different from how a state machine typically works. Normally, you don't specify an initial state and the next state. You specify the initial state and the event that occurred. The state machine tells you the next state. So I think specifying the initial state as an enum and the event as an instance method on that enum is a better way to handle this.


> Similarly, CANCELLED should never change to anything else, except ERROR. 

I disagree with this part. Once something is cancelled the fact that there was an error is irrelevant. The user already told us they are not interested in the result when they cancelled the job.

---
*[a discussion](https://reviewable.io/reviews/openmpf/openmpf/1364#-Me6EeW_7zV4rW6Dk8ew:-MezaCvr1oAUB2uD1NIg:b-896fix) (no related file):*

A job will appear as `COMPLETE` while TiesDb post / callback attempts are still going on in the background. Attempting to resubmit a job in this state will result in an exception and UI popup with the message:

> Failed to resubmit the job with id '1285'. Job with id 1285 already exists.

This is confusing to an end user, who doesn't know what's going on in the background. Once this happens the UI will stay as `RESUBMITTING` until the UI is refreshed.

I think we need some indication in the UI that there is still something going on. Minimally, disable the Resubmit button, but that alone might be a little confusing. Maybe change the tooltip / hint popover to say "Callbacks in progress" or something.

---

Also, the job will stay in `RESUBMITTING` if the media doesn't exist when you try to resubmit.
</blockquote></details>

Done.

---
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/JobCompleteProcessorImpl.java, line 160 at r1](https://reviewable.io/reviews/openmpf/openmpf/1364#-MeRWKZz9w6iX9jBeEnK:-MeaCny04wgtJrKLUkbj:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/d11895adec3f7727d27a60f71aa6930749a43687/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/JobCompleteProcessorImpl.java#L160)):*

In `WfmSplitter` we have:

```
            log.error(String.format("Failed to complete the split operation for Job %s due to an exception.", jobId),
                      exception);
            inProgressJobs.setJobStatus(jobId, BatchJobStatusType.ERROR);
```

We should be adding an error message with the exception there.

As stated in the task:

> When the job status is `ERROR`, the JSON output object does not have any messages in the errors field. We should ensure that a meaningful error message is added whenever we set the job status to `ERROR`.
</blockquote></details>

Done.
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/JobCompleteProcessorImpl.java, line 168 at r1](https://reviewable.io/reviews/openmpf/openmpf/1364#-Me6Df7f43DOJ0c5LQsc:-MeaCdLz834IOkquGOse:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/d11895adec3f7727d27a60f71aa6930749a43687/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/JobCompleteProcessorImpl.java#L168)):*

Submitting a Sphinx job with an image will result in first transitioning to `COMPLETE` before `COMPLETE_WITH_ERRORS`. While `COMPLETE` is shown in red, I think we should transition to the actual terminal status,`COMPLETE_WITH_ERRORS`, directly.

While debugging I confirmed that at this point in the code `jobStatus` is already set to `COMPLETE_WITH_ERRORS` so we should just use that. Doing so will ensure that what's in the UI, `inProgressBatchJobs`, and the DB is all the same.

If we do this, I think we can replace `setJobStatusNoBroadcast()` here with `setJobStatus()`:
```
inProgressBatchJobs.setJobStatusNoBroadcast(jobId, jobStatus.getValue());
```

In general, I'm wondering if we really need `setJobStatusNoBroadcast()` to be a separate public method or if we can just add the behavior to `setJobStatus()`.

---

Also, a few lines above we have `jobStatus.setValue(BatchJobStatusType.ERROR)`. We don't want to change the status back to `COMPLETE`.
</blockquote></details>

Done.
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/artifactextraction/ArtifactExtractionProcessor.java, line 190 at r1](https://reviewable.io/reviews/openmpf/openmpf/1364#-MeRKgwI-YSrzTLc3JwA:-Me_XHMV7gJGt_WhzwgD:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/d11895adec3f7727d27a60f71aa6930749a43687/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/artifactextraction/ArtifactExtractionProcessor.java#L190)):*

I think you mean `failed due to: " + e);`
</blockquote></details>

Done.
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/InProgressBatchJobsService.java, line 227 at r1](https://reviewable.io/reviews/openmpf/openmpf/1364#-Me6onKJ9jVL8183LnYg:-MeaChzGCmjD8ohy5atA:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/d11895adec3f7727d27a60f71aa6930749a43687/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/InProgressBatchJobsService.java#L227)):*

It's not always appropriate to transition to an `IN_PROGRESS` state here.

Looking at this code:

```
        JobStatusCalculator.checkErrorMessages(jsonOutputObject, jobStatus);

        return storageService.store(jsonOutputObject, jobStatus, outputSha); // this may update the job status
```

That first line can set `jobStatus` to `COMPLETE`. `StorageService.store(JsonOutputObject ...)` can call `addJobWarning()`.

We don't want to transition back to `IN_PROGRESS_WARNINGS`.
</blockquote></details>

Done.
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/access/JobRequestDao.java, line 45 at r1](https://reviewable.io/reviews/openmpf/openmpf/1364#-Me6m3OtEHIgPsGE3OUl:-Me_XPA_-XxVOSE8L9oQ:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/d11895adec3f7727d27a60f71aa6930749a43687/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/access/JobRequestDao.java#L45)):*

Add `public` to this.
</blockquote></details>

Done.
___
*[trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/camelOps/TestDetectionResponseProcessor.java, line 222 at r1](https://reviewable.io/reviews/openmpf/openmpf/1364#-MeROxdfDzj5JdSG343s:-Me_aZh_0Aa2DJhWy9SI:b-n3oazg) ([raw file](https://github.com/openmpf/openmpf/blob/fca53dbea8ee0a4a7df8acc5213b9e87551238d7/trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/camelOps/TestDetectionResponseProcessor.java#L222)):*

`addDetectionProcessingError` still has the logic for setting the job status:

```
        if (error.getErrorCode().equals(MpfConstants.REQUEST_CANCELLED)) {
            setJobStatus(error.getJobId(), BatchJobStatusType.CANCELLING);
        }
        else {
            setJobStatus(error.getJobId(), BatchJobStatusType.IN_PROGRESS_ERRORS);
        }
```

Why did you remove the test code that verifies that the status was set correctly?
</blockquote></details>

Before, `DetectionResponseProcessor` called `InProgressBatchJobsService.setJobStatus` directly, that is no longer the case. That code snippet occurs within `InProgressBatchJobsService`, you can't test what a mocked object does internally because a mock doesn't actually do anything.

brosenberg42 (commented):
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1364)* status: 20 of 49 files reviewed, 10 unresolved discussions (waiting on @brosenberg42 and @jrobble)

___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/enums/BatchJobStatusType.java, line 72 at r1](https://reviewable.io/reviews/openmpf/openmpf/1364#-Me6kGb5A_-95FoqKcoy:-MezofAa3gvVbB-5W_zh:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/fca53dbea8ee0a4a7df8acc5213b9e87551238d7/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/enums/BatchJobStatusType.java#L72)):*

`BUILDING_OUTPUT_OBJECT` is still mentioned in the following files:
- `jobs/layout.html` in the `Status` column legend when you click the `i` icon.
- `REST-API.html`:

> A string that will be one of the following values: UNKNOWN, INITIALIZED (the job has been initialized but not started), JOB_CREATION_ERROR (indicates that a job was received, but a job could not be created from the contents of the request), IN_PROGRESS, BUILDING_OUTPUT_OBJECT, COMPLETE, CANCELLING, CANCELLED, ERROR

The possible statuses in that list need to be updated. It's missing the warning and error variants.

</blockquote></details>

Done.

jrobble (changes requested):
> Reviewed 29 of 29 files at r2.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1364)* status: all files reviewed, 15 unresolved discussions (waiting on @brosenberg42)

*[a discussion](https://reviewable.io/reviews/openmpf/openmpf/1364#-Me6EeW_7zV4rW6Dk8ew:-Mf-gGiYEWs-zobIlY8b:b-b55vmu) (no related file):*
Your updates address the first part. That works well. There is still this part:

> Also, the job will stay in RESUBMITTING if the media doesn't exist when you try to resubmit.

---

Related, please clear the "End Date" contents in the UI for that job after the user clicks the resubmit button. Currently, it shows the end date for the last run and doesn't get updated until the job is complete. The `[GET] /rest/jobs/id` response also shows the old end date until the job is complete.

---
*[a discussion](https://reviewable.io/reviews/openmpf/openmpf/1364#-Me7vpJM9kw5JYnSquiJ:-Mf-gewUEb0PfmAA8f0X:b-df1dwe) (no related file):*

`DefaultJobErrorHandler` does not appear to be used.
</blockquote></details>

Should we remove it?

---
*[a discussion](https://reviewable.io/reviews/openmpf/openmpf/1364#-Me81918-c4kH1K67Y5a:-Mf-h3B9CQGz79u8-p1M:b-yp6he3) (no related file):*

> I think we should centralize the logic that validates the transitions between states. I like JobStatusCalculator.completionStatus(). I would feel more confident if we had something like that for non-terminal states. Maybe something like: `BatchJobStatusType  JobStatusCalculator.validateNextStatus(BatchJobStatusType initialStatus, BatchJobStatusType nextStatus)`

I think the reason we would need to validate the state transitions is that this is different from how a state machine typically works. Normally, you don't specify an initial state and the next state. You specify the initial state and the event that occurred. The state machine tells you the next state. So I think specifying the initial state as an enum and the event as an instance method on that enum is a better way to handle this.


> Similarly, CANCELLED should never change to anything else, except ERROR. 

I disagree with this part. Once something is cancelled the fact that there was an error is irrelevant. The user already told us they are not interested in the result when they cancelled the job.
</blockquote></details>

I think the user would be interested if the job was unable to be cancelled due to an error. However, looking [here](https://github.com/openmpf/openmpf/blob/master/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/businessrules/impl/JobRequestServiceImpl.java#L372), we only log the exception anyway, so never mind.

---
___
*[trunk/mpf-rest-api/src/main/java/org/mitre/mpf/rest/api/SingleJobInfo.java, line 43 at r2](https://reviewable.io/reviews/openmpf/openmpf/1364#-Mf3d64j5BacNG_qHmTC:-Mf3d64j5BacNG_qHmTD:biyc357) ([raw file](https://github.com/openmpf/openmpf/blob/235303ee39b4179fcdea60ec264a32a24ac885d5/trunk/mpf-rest-api/src/main/java/org/mitre/mpf/rest/api/SingleJobInfo.java#L43)):*
> ```Java
>     private String outputObjectPath;
>     private boolean hasCallbacksInProgress;
>     //terminal if status is JOB_CREATION_ERROR, COMPLETE, CANCELLED, or ERROR.
> ```

Let's just remove this comment.
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/JobCompleteProcessorImpl.java, line 155 at r2](https://reviewable.io/reviews/openmpf/openmpf/1364#-Mf46aKb7OyuYMBsrZ9A:-Mf46aKeB2ehS5FUyJad:be6i4j5) ([raw file](https://github.com/openmpf/openmpf/blob/235303ee39b4179fcdea60ec264a32a24ac885d5/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/JobCompleteProcessorImpl.java#L155)):*
> ```Java
>                 jobRequest.setOutputObjectVersion(propertiesUtil.getOutputObjectVersion());
>             } catch (Exception exception) {
>                 inProgressBatchJobs.addFatalError(
> ```

Log the stack trace.
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/JobCompleteProcessorImpl.java, line 166 at r2](https://reviewable.io/reviews/openmpf/openmpf/1364#-Mf4KjcR-5GoBp2mqaEK:-Mf4KjcR-5GoBp2mqaEL:b-p16lmz) ([raw file](https://github.com/openmpf/openmpf/blob/235303ee39b4179fcdea60ec264a32a24ac885d5/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/JobCompleteProcessorImpl.java#L166)):*
> ```Java
>             jobRequest.setJob(jsonUtils.serialize(job));
>             jobRequestDao.persist(jobRequest);
>             jobStatusBroadcaster.broadcast(jobId, 100, completionStatus);
> ```

Two popups appear on the job status page when a job completes. One is due to this `jobStatusBroadcaster.broadcast(jobId, 100, completionStatus);` call. The other is due to the `-> completeJob(job)` calls below. This is because `completeJob()` results in another broadcast.
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/markup/MarkupResponseProcessor.java, line 102 at r2](https://reviewable.io/reviews/openmpf/openmpf/1364#-Mf4MXHN3RgJYTeB_eyo:-Mf4MXHO2DLBC9d3QFed:b-6j50o9) ([raw file](https://github.com/openmpf/openmpf/blob/235303ee39b4179fcdea60ec264a32a24ac885d5/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/markup/MarkupResponseProcessor.java#L102)):*
> ```Java
>                                       warningMessage, IssueSources.MARKUP);
>         }
>         return null;
> ```

I tested a FACE --> MARKUP pipeline where I stopped the markup service so that the job could only complete the first stage. Then I cancelled the job through the UI. The job ended in `COMPLETE_WITH_ERRORS` with this in the JSON:

```
   "errors": [
      {
         "mediaId": 24,
         "details": [
            {
               "source": "MARKUP",
               "code": "MARKUP",
               "message": "This request was cancelled."
            }
         ]
      }
   ],
```

The job should have ended in `CANCELLED`.
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/mediainspection/MediaInspectionProcessor.java, line 139 at r2](https://reviewable.io/reviews/openmpf/openmpf/1364#-Mf3kb352Sk1-KxxdLx7:-Mf3kb352Sk1-KxxdLx8:bgi38ec) ([raw file](https://github.com/openmpf/openmpf/blob/235303ee39b4179fcdea60ec264a32a24ac885d5/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/mediainspection/MediaInspectionProcessor.java#L139)):*
> ```Java
>                     String errorMessage = "Could not calculate the SHA-256 hash for the file due to IOException: "
>                             + ioe;
>                     _inProgressJobs.addError(jobId, mediaId, IssueCodes.ARTIFACT_EXTRACTION, errorMessage);
> ```

I think we want `MEDIA_INSPECTION` here instead of `ARTIFACT_EXTRACTION`.
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/mediaretrieval/RemoteMediaProcessor.java, line 113 at r2](https://reviewable.io/reviews/openmpf/openmpf/1364#-Mf3guFH1PBz71vlmowc:-Mf3guFH1PBz71vlmowd:b13vpwz) ([raw file](https://github.com/openmpf/openmpf/blob/235303ee39b4179fcdea60ec264a32a24ac885d5/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/mediaretrieval/RemoteMediaProcessor.java#L113)):*
> ```Java
>                 break;
>             default:
>                 log.warn("The UriScheme '{}' was not expected at this time.", media.getUriScheme());
> ```

Since we're adding an error, please change this to `log.error()`.
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/InProgressBatchJobsService.java, line 241 at r2](https://reviewable.io/reviews/openmpf/openmpf/1364#-Mf3hMW_5BpDIB7xnSrO:-Mf3hMWa8MXfGJ-kwV-F:bcskbgi) ([raw file](https://github.com/openmpf/openmpf/blob/235303ee39b4179fcdea60ec264a32a24ac885d5/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/InProgressBatchJobsService.java#L241)):*
> ```Java
> 
> 
>     public synchronized void addJobError(long jobId, IssueCodes code, String message) {
> ```

This can be removed. It's only used by a test to determine that it's never called.

`addJobWarning()` is used, so maybe you want to keep `addJobError()` for consistency and future use, or comment it out for now. Your decision.
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/InProgressBatchJobsService.java, line 272 at r2](https://reviewable.io/reviews/openmpf/openmpf/1364#-Mf3X-VtC_5oiAhAs2mj:-Mf3X-VtC_5oiAhAs2mk:bb37hcf) ([raw file](https://github.com/openmpf/openmpf/blob/235303ee39b4179fcdea60ec264a32a24ac885d5/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/InProgressBatchJobsService.java#L272)):*
> ```Java
> 
>     public synchronized void addDetectionProcessingError(DetectionProcessingError error) {
>         LOG.info("Adding detection processing error for job {}'s media {}: {} - {}",
> ```

Please change this to `LOG.error()`.
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/InProgressBatchJobsService.java, line 301 at r2](https://reviewable.io/reviews/openmpf/openmpf/1364#-Mf49Fey9G5-fSabU5uw:-Mf-mCce-wVzO7lNHD04:b-jze56j) ([raw file](https://github.com/openmpf/openmpf/blob/235303ee39b4179fcdea60ec264a32a24ac885d5/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/InProgressBatchJobsService.java#L301)):*
> ```Java
>         LOG.info("Setting status of job {} to {}", jobId, batchJobStatus);
>         job.setStatus(batchJobStatus);
>         if (!batchJobStatus.isTerminal()) {
> ```

Why are we performing this terminal check?

For example, in theory it's possible to go from `COMPLETE` to `COMPLETE_WITH_ERRORS`. Don't we want that to be reflected in the DB and UI?

Is your thought that `JobCompleteProcessorImpl.wfmProcess()` is already handling DB persistence and the broadcast?
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/enums/BatchJobStatusType.java, line 72 at r1](https://reviewable.io/reviews/openmpf/openmpf/1364#-Me6kGb5A_-95FoqKcoy:-Mf-khqX5omeryZPUV9U:b-ulzopv) ([raw file](https://github.com/openmpf/openmpf/blob/fca53dbea8ee0a4a7df8acc5213b9e87551238d7/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/enums/BatchJobStatusType.java#L72)):*

Done.
</blockquote></details>

The UI has one additional status that is not reflected on the backend: `RESUBMITTING`. Please add that to the legend on the job status page. 
___
*[trunk/workflow-manager/src/main/webapp/resources/mytheme/js/controllers/JobsCtrl.js, line 138 at r2](https://reviewable.io/reviews/openmpf/openmpf/1364#-Mf3dMJY7kbEZesbRK7y:-Mf3dMJZ-LNKjIgOnmIq:b-r1qyj5) ([raw file](https://github.com/openmpf/openmpf/blob/235303ee39b4179fcdea60ec264a32a24ac885d5/trunk/workflow-manager/src/main/webapp/resources/mytheme/js/controllers/JobsCtrl.js#L138)):*
> ```JavaScript
>                             var cancel_disabled = "";
>                             if (job.terminal || job.jobStatus === 'CANCELLING' ||
>                                     job.jobStatus === 'COMPLETE' || job.hasCallbacksInProgress) {
> ```

`COMPLETE` is a terminal status, so do we need a separate check for it here? I would think it's covered by `if (job.terminal`.
___
*[trunk/workflow-manager/src/main/webapp/resources/mytheme/js/controllers/JobsCtrl.js, line 141 at r2](https://reviewable.io/reviews/openmpf/openmpf/1364#-Mf3cXK_-lB4S-fszWS8:-Mf3cXKaA_p24aalCFpM:b2scgjv) ([raw file](https://github.com/openmpf/openmpf/blob/235303ee39b4179fcdea60ec264a32a24ac885d5/trunk/workflow-manager/src/main/webapp/resources/mytheme/js/controllers/JobsCtrl.js#L141)):*
> ```JavaScript
>                                 cancel_disabled = "disabled=disabled";
>                             }
>                             var isterminal = "";
> ```

I think it be clearer to call this `resubmit_disabled`.
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/BroadcastEnabledCountBasedWfmAggregator.java, line 70 at r2](https://reviewable.io/reviews/openmpf/openmpf/1364#-Mf3S3TW9tschmLn20cO:-Mf3S3TXC0p0li-Jyowb:b4c9lyr) ([raw file](https://github.com/openmpf/openmpf/blob/235303ee39b4179fcdea60ec264a32a24ac885d5/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/BroadcastEnabledCountBasedWfmAggregator.java#L70)):*
> ```Java
>             int totalTasks = job.getPipelineElements().getTaskCount();
>             float progressInCurrentTask = (float) aggregateCount / splitSize;
>             // Make sure job progress never gets to 100% before the job is actually complete.
> ```

:thumbsup:

brosenberg42 (commented):
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1364)* status: 39 of 51 files reviewed, 14 unresolved discussions (waiting on @brosenberg42 and @jrobble)

*[a discussion](https://reviewable.io/reviews/openmpf/openmpf/1364#-Me7vpJM9kw5JYnSquiJ:-Mf8WCX_4Clj2_5Q2wXr:b-xm1zg5) (no related file):*

Should we remove it?
</blockquote></details>

Yes, I wanted to make sure I wasn't missing something.

---
*[a discussion](https://reviewable.io/reviews/openmpf/openmpf/1364#-Me6EeW_7zV4rW6Dk8ew:-Mf9WWLE6tqLRTcMzu9X:b-896fix) (no related file):*

Your updates address the first part. That works well. There is still this part:

> Also, the job will stay in RESUBMITTING if the media doesn't exist when you try to resubmit.

---

Related, please clear the "End Date" contents in the UI for that job after the user clicks the resubmit button. Currently, it shows the end date for the last run and doesn't get updated until the job is complete. The `[GET] /rest/jobs/id` response also shows the old end date until the job is complete.
</blockquote></details>

Done.

---
___
*[trunk/mpf-rest-api/src/main/java/org/mitre/mpf/rest/api/SingleJobInfo.java, line 43 at r2](https://reviewable.io/reviews/openmpf/openmpf/1364#-Mf3d64j5BacNG_qHmTC:-Mf8Wbq37McaRw-ckGzD:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/235303ee39b4179fcdea60ec264a32a24ac885d5/trunk/mpf-rest-api/src/main/java/org/mitre/mpf/rest/api/SingleJobInfo.java#L43)):*

Let's just remove this comment.
</blockquote></details>

Done.
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/JobCompleteProcessorImpl.java, line 155 at r2](https://reviewable.io/reviews/openmpf/openmpf/1364#-Mf46aKb7OyuYMBsrZ9A:-Mf8WrXuDwAU65Ypzr4u:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/235303ee39b4179fcdea60ec264a32a24ac885d5/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/JobCompleteProcessorImpl.java#L155)):*

Log the stack trace.
</blockquote></details>

Done.
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/JobCompleteProcessorImpl.java, line 166 at r2](https://reviewable.io/reviews/openmpf/openmpf/1364#-Mf4KjcR-5GoBp2mqaEK:-Mf8Wz1c4fyA5dU-SMwU:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/235303ee39b4179fcdea60ec264a32a24ac885d5/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/JobCompleteProcessorImpl.java#L166)):*

Two popups appear on the job status page when a job completes. One is due to this `jobStatusBroadcaster.broadcast(jobId, 100, completionStatus);` call. The other is due to the `-> completeJob(job)` calls below. This is because `completeJob()` results in another broadcast.
</blockquote></details>

Done.
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/markup/MarkupResponseProcessor.java, line 102 at r2](https://reviewable.io/reviews/openmpf/openmpf/1364#-Mf4MXHN3RgJYTeB_eyo:-Mf99SnX-91MXCot6z_M:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/235303ee39b4179fcdea60ec264a32a24ac885d5/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/markup/MarkupResponseProcessor.java#L102)):*

I tested a FACE --> MARKUP pipeline where I stopped the markup service so that the job could only complete the first stage. Then I cancelled the job through the UI. The job ended in `COMPLETE_WITH_ERRORS` with this in the JSON:

```
   "errors": [
      {
         "mediaId": 24,
         "details": [
            {
               "source": "MARKUP",
               "code": "MARKUP",
               "message": "This request was cancelled."
            }
         ]
      }
   ],
```

The job should have ended in `CANCELLED`.
</blockquote></details>

Done.
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/mediainspection/MediaInspectionProcessor.java, line 139 at r2](https://reviewable.io/reviews/openmpf/openmpf/1364#-Mf3kb352Sk1-KxxdLx7:-Mf8X6Uj6dpHN8iOuX0y:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/235303ee39b4179fcdea60ec264a32a24ac885d5/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/mediainspection/MediaInspectionProcessor.java#L139)):*

I think we want `MEDIA_INSPECTION` here instead of `ARTIFACT_EXTRACTION`.
</blockquote></details>

Done.
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/mediaretrieval/RemoteMediaProcessor.java, line 113 at r2](https://reviewable.io/reviews/openmpf/openmpf/1364#-Mf3guFH1PBz71vlmowc:-Mf8XFxoCp_2u_5TEkka:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/235303ee39b4179fcdea60ec264a32a24ac885d5/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/mediaretrieval/RemoteMediaProcessor.java#L113)):*

Since we're adding an error, please change this to `log.error()`.
</blockquote></details>

Done.
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/InProgressBatchJobsService.java, line 241 at r2](https://reviewable.io/reviews/openmpf/openmpf/1364#-Mf3hMW_5BpDIB7xnSrO:-Mf8YU-m6jLYAteVRBRw:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/235303ee39b4179fcdea60ec264a32a24ac885d5/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/InProgressBatchJobsService.java#L241)):*

This can be removed. It's only used by a test to determine that it's never called.

`addJobWarning()` is used, so maybe you want to keep `addJobError()` for consistency and future use, or comment it out for now. Your decision.
</blockquote></details>

Done.
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/InProgressBatchJobsService.java, line 272 at r2](https://reviewable.io/reviews/openmpf/openmpf/1364#-Mf3X-VtC_5oiAhAs2mj:-Mf8XNNsD_O-602p1kkt:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/235303ee39b4179fcdea60ec264a32a24ac885d5/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/InProgressBatchJobsService.java#L272)):*

Please change this to `LOG.error()`.
</blockquote></details>

Done.
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/InProgressBatchJobsService.java, line 301 at r2](https://reviewable.io/reviews/openmpf/openmpf/1364#-Mf49Fey9G5-fSabU5uw:-Mf8VDQl69cKPZ_OyJY6:b-by84t3) ([raw file](https://github.com/openmpf/openmpf/blob/235303ee39b4179fcdea60ec264a32a24ac885d5/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/InProgressBatchJobsService.java#L301)):*
This change was your suggestion:
> In general, I'm wondering if we really need setJobStatusNoBroadcast() to be a separate public method or if we can just add the behavior to setJobStatus().
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/enums/BatchJobStatusType.java, line 72 at r1](https://reviewable.io/reviews/openmpf/openmpf/1364#-Me6kGb5A_-95FoqKcoy:-Mf8fsc32JdCAp4lqG-v:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/fca53dbea8ee0a4a7df8acc5213b9e87551238d7/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/enums/BatchJobStatusType.java#L72)):*

The UI has one additional status that is not reflected on the backend: `RESUBMITTING`. Please add that to the legend on the job status page. 
</blockquote></details>

Done.
___
*[trunk/workflow-manager/src/main/webapp/resources/mytheme/js/controllers/JobsCtrl.js, line 138 at r2](https://reviewable.io/reviews/openmpf/openmpf/1364#-Mf3dMJY7kbEZesbRK7y:-Mf8Xs7K4kdDCwcvkLQd:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/235303ee39b4179fcdea60ec264a32a24ac885d5/trunk/workflow-manager/src/main/webapp/resources/mytheme/js/controllers/JobsCtrl.js#L138)):*

`COMPLETE` is a terminal status, so do we need a separate check for it here? I would think it's covered by `if (job.terminal`.
</blockquote></details>

Done.
___
*[trunk/workflow-manager/src/main/webapp/resources/mytheme/js/controllers/JobsCtrl.js, line 141 at r2](https://reviewable.io/reviews/openmpf/openmpf/1364#-Mf3cXK_-lB4S-fszWS8:-Mf8XkRc6U-Tg2fkWaEe:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/235303ee39b4179fcdea60ec264a32a24ac885d5/trunk/workflow-manager/src/main/webapp/resources/mytheme/js/controllers/JobsCtrl.js#L141)):*

I think it be clearer to call this `resubmit_disabled`.
</blockquote></details>

Done.

jrobble (changes requested):
> Reviewed 11 of 12 files at r3, 1 of 1 files at r4.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1364)* status: all files reviewed, 3 unresolved discussions (waiting on @brosenberg42)

___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/WfmSplitter.java, line 88 at r4](https://reviewable.io/reviews/openmpf/openmpf/1364#-MfEJX2A6JiikOeQSWzS:-MfEJX2A6JiikOeQSWzT:bb6mye8) ([raw file](https://github.com/openmpf/openmpf/blob/3f50c80bd96c50558895fef4d9d9d9b82c47fe05/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/WfmSplitter.java#L88)):*
> ```Java
>         catch (Exception exception) {
>             // The split operation failed. The job cannot continue.
>             // Print out the stack trace since no details will be reported in the JSON output.
> ```

Please print out the stack trace for this fatal error too. This is what I get in the logs:

```
2021-07-22 13:43:30,880 INFO [http-nio-8080-exec-3] o.m.m.w.b.i.JobRequestServiceImpl - [Job 1366|*|*] Job has started and is running at priority 0.
2021-07-22 13:43:31,092 ERROR [Camel (camelContext) thread #1 - JmsConsumer[MPF.MEDIA_RETRIEVAL]] o.m.m.w.d.InProgressBatchJobsService - Adding the following error to job 1366: OTHER - Failed to complete the split operation for Job 1366 due to: java.lang.RuntimeException: BAD THING!
2021-07-22 13:43:31,119 INFO [Camel (camelContext) thread #1 - JmsConsumer[MPF.MEDIA_RETRIEVAL]] o.m.m.w.d.InProgressBatchJobsService - Setting status of job 1366 to ERROR
```
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/InProgressBatchJobsService.java, line 301 at r2](https://reviewable.io/reviews/openmpf/openmpf/1364#-Mf49Fey9G5-fSabU5uw:-MfEkC7K7fKHnSSTVXHW:bvflyyj) ([raw file](https://github.com/openmpf/openmpf/blob/235303ee39b4179fcdea60ec264a32a24ac885d5/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/InProgressBatchJobsService.java#L301)):*
I found where you persist and broadcast the terminal job in other places in the code. I'm assuming this check is to prevent duplication of those behaviors. At first glance it seems odd. Please add a comment explaining as much. Something like:

> // Terminal jobs are persisted, and status is broadcast, in other code.
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/InProgressBatchJobsService.java, line 284 at r4](https://reviewable.io/reviews/openmpf/openmpf/1364#-MfEefhDEdesgXmekwRJ:-MfEefhDEdesgXmekwRK:be887hj) ([raw file](https://github.com/openmpf/openmpf/blob/3f50c80bd96c50558895fef4d9d9d9b82c47fe05/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/InProgressBatchJobsService.java#L284)):*
> ```Java
> 
>         if (error.getErrorCode().equals(MpfConstants.REQUEST_CANCELLED)) {
>             cancelJob(error.getJobId());
> ```

I get the "Marking job X as cancelled" twice in the logs:

```
2021-07-22 14:00:54,977 INFO [http-nio-8080-exec-6] o.m.m.w.b.i.JobRequestServiceImpl - [Job 1367:*:*] Cancelling job.
2021-07-22 14:00:54,977 INFO [http-nio-8080-exec-6] o.m.m.w.d.InProgressBatchJobsService - Marking job 1367 as cancelled.
2021-07-22 14:00:54,977 INFO [http-nio-8080-exec-6] o.m.m.w.d.InProgressBatchJobsService - Setting status of job 1367 to CANCELLING
2021-07-22 14:00:55,133 INFO [Camel (camelContext) thread #10 - JmsConsumer[MPF.COMPLETED_MARKUP]] o.m.m.w.d.InProgressBatchJobsService - Marking job 1367 as cancelled.
```

Maybe check `job.isCancelled()` first inside `cancelJob()` before executing the lines of code again. In general, running the same code twice that's in there now isn't much of a concern (unless users are parsing the "Marking job X" line). It might be more of a concern if it's eventually  updated and no longer idempotent.

brosenberg42 (commented):
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1364)* status: 49 of 52 files reviewed, 3 unresolved discussions (waiting on @brosenberg42 and @jrobble)

___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/WfmSplitter.java, line 88 at r4](https://reviewable.io/reviews/openmpf/openmpf/1364#-MfEJX2A6JiikOeQSWzS:-MfJ1JMM1xTWFczCzoXp:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/3f50c80bd96c50558895fef4d9d9d9b82c47fe05/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/WfmSplitter.java#L88)):*

Please print out the stack trace for this fatal error too. This is what I get in the logs:

```
2021-07-22 13:43:30,880 INFO [http-nio-8080-exec-3] o.m.m.w.b.i.JobRequestServiceImpl - [Job 1366|*|*] Job has started and is running at priority 0.
2021-07-22 13:43:31,092 ERROR [Camel (camelContext) thread #1 - JmsConsumer[MPF.MEDIA_RETRIEVAL]] o.m.m.w.d.InProgressBatchJobsService - Adding the following error to job 1366: OTHER - Failed to complete the split operation for Job 1366 due to: java.lang.RuntimeException: BAD THING!
2021-07-22 13:43:31,119 INFO [Camel (camelContext) thread #1 - JmsConsumer[MPF.MEDIA_RETRIEVAL]] o.m.m.w.d.InProgressBatchJobsService - Setting status of job 1366 to ERROR
```
</blockquote></details>

Done.
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/InProgressBatchJobsService.java, line 301 at r2](https://reviewable.io/reviews/openmpf/openmpf/1364#-Mf49Fey9G5-fSabU5uw:-MfJ5I0DEe9i8bnjyhNC:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/235303ee39b4179fcdea60ec264a32a24ac885d5/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/InProgressBatchJobsService.java#L301)):*

I found where you persist and broadcast the terminal job in other places in the code. I'm assuming this check is to prevent duplication of those behaviors. At first glance it seems odd. Please add a comment explaining as much. Something like:

> // Terminal jobs are persisted, and status is broadcast, in other code.
</blockquote></details>

Done.
___
*[trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/InProgressBatchJobsService.java, line 284 at r4](https://reviewable.io/reviews/openmpf/openmpf/1364#-MfEefhDEdesgXmekwRJ:-MfJ4vTy3aIAdF9ebsmJ:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/3f50c80bd96c50558895fef4d9d9d9b82c47fe05/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/InProgressBatchJobsService.java#L284)):*

I get the "Marking job X as cancelled" twice in the logs:

```
2021-07-22 14:00:54,977 INFO [http-nio-8080-exec-6] o.m.m.w.b.i.JobRequestServiceImpl - [Job 1367:*:*] Cancelling job.
2021-07-22 14:00:54,977 INFO [http-nio-8080-exec-6] o.m.m.w.d.InProgressBatchJobsService - Marking job 1367 as cancelled.
2021-07-22 14:00:54,977 INFO [http-nio-8080-exec-6] o.m.m.w.d.InProgressBatchJobsService - Setting status of job 1367 to CANCELLING
2021-07-22 14:00:55,133 INFO [Camel (camelContext) thread #10 - JmsConsumer[MPF.COMPLETED_MARKUP]] o.m.m.w.d.InProgressBatchJobsService - Marking job 1367 as cancelled.
```

Maybe check `job.isCancelled()` first inside `cancelJob()` before executing the lines of code again. In general, running the same code twice that's in there now isn't much of a concern (unless users are parsing the "Marking job X" line). It might be more of a concern if it's eventually  updated and no longer idempotent.
</blockquote></details>

Done.

jrobble (approved):
> Reviewed 3 of 3 files at r5.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1364)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @brosenberg42)

jrobble (approved):
> Reviewed 4 of 4 files at r6.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1364)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @brosenberg42)

Issue #1010, opened 2019-12-03
`mpf.output.objects.enabled` does not behave as expected


jrobble:
I've been looking at how `mpf.output.objects.enabled` is used because I was trying to test it in conjunction with the other `mpf.output.objects.*` properties. In general, it's used in two places:

**1. EndOfStageProcessor checks mpf.output.objects.enabled here:**
```
			//notify of completion - use
			if(!job.isOutputEnabled()) {
				jobStatusBroadcaster.broadcast(
						jobId,
						100,
						job.isCancelled() ? BatchJobStatusType.CANCELLED : BatchJobStatusType.COMPLETE,
						new Date());
				jobProgressStore.setJobProgress(jobId, 100.0f);
			} else {
				jobStatusBroadcaster.broadcast(jobId, 99, BatchJobStatusType.BUILDING_OUTPUT_OBJECT, new Date());
				jobProgressStore.setJobProgress(jobId, 99.0f);
			}	
```
This doesn't really make sense because the output object is always created in JobCompleteProcessorImpl:
```
			try {
				markJobStatus(jobId, BatchJobStatusType.BUILDING_OUTPUT_OBJECT);

				// NOTE: jobStatus is mutable - it __may__ be modified in createOutputObject!
				createOutputObject(jobId, jobStatus);
```
Looking back, it seems that the last time mpf.output.objects.enabled was properly used for batch jobs was in R2.1.0 in OutputObjectSenderProcessor:
```
			if (jsonJobRequest.isOutputObjectEnabled()) {
				JsonOutputObjectSummary jsonOutputObjectSummary = new JsonOutputObjectSummary(jobRequest.getOutputObjectPath());
				exchange.getOut().setHeader(MpfHeaders.SEND_OUTPUT_OBJECT, Boolean.TRUE);
				exchange.getOut().setHeader(MpfHeaders.JOB_ID, jobId);
				exchange.getOut().setBody(jsonUtils.serializeAsText(jsonOutputObjectSummary));
			}
```
Which resulted in placing a message in the old MPF_OUTPUT queue. From JobCompletedRouteBuilder:
```
			.choice()
				.when(header(MpfHeaders.SEND_OUTPUT_OBJECT).isEqualTo(Boolean.TRUE))
					.to(String.format("output:queue:%s", propertiesUtil.getOutputQueueName()), exitPoint)
```
[...]

PR #1394, opened 2021-08-13
Remove use of output.objects.enabled for batch jobs and add streaming.output.objects.to.disk.enabled property

clnowacki:
**Issues:**

- https://github.com/openmpf/openmpf/issues/1010

---
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1394)


jrobble (changes requested):
> Reviewed 20 of 20 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1394)* status: all files reviewed, 1 unresolved discussion (waiting on @clnowacki)

___
*[trunk/workflow-manager/src/main/resources/properties/mpf.properties, line 314 at r1](https://reviewable.io/reviews/openmpf/openmpf/1394#-MjW3IIu3dLdmAqIYvKs:-MjW3IIu3dLdmAqIYvKt:b-bx0iqd) ([raw file](https://github.com/openmpf/openmpf/blob/81cdc15f5ebd12082cf87a3c350f19fcff7b2a87/trunk/workflow-manager/src/main/resources/properties/mpf.properties#L314)):*
Let's reword:

> Normally, these are retrieved via a callback, but the user also may choose to write them to disk.

to:

> Normally, these are received via a callback, but if this setting is enabled the user may retrieve them from disk.

Note that I replaced the first "retrieved" with "received".

The bit about "choose to write them to disk" is a little misleading since if the user chooses to do nothing that's the default setting.

clnowacki (commented):
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1394)* status: all files reviewed, 1 unresolved discussion (waiting on @jrobble)

___
*[trunk/workflow-manager/src/main/resources/properties/mpf.properties, line 314 at r1](https://reviewable.io/reviews/openmpf/openmpf/1394#-MjW3IIu3dLdmAqIYvKs:-MjZVt638OMAgoaj23Do:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/c90166b3ff537ee794c395cc7f0591f2af3c570d/trunk/workflow-manager/src/main/resources/properties/mpf.properties#L314)):*

Let's reword:

> Normally, these are retrieved via a callback, but the user also may choose to write them to disk.

to:

> Normally, these are received via a callback, but if this setting is enabled the user may retrieve them from disk.

Note that I replaced the first "retrieved" with "received".

The bit about "choose to write them to disk" is a little misleading since if the user chooses to do nothing that's the default setting.
</blockquote></details>

Done.

jrobble (approved):
> Reviewed 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1394)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @clnowacki)

Issue #1601, opened 2022-10-06
Remove "Showing 1 to 25 of 102 entries" from the jobs table.


brosenberg42:
According to the simon timings, getting those counts takes a lot longer than getting the job data.
HibernateJobRequestDaoImpl.findByPage average: 32ms
HibernateJobRequestDaoImpl.countAll average: 11,315 ms

PR #1618, opened 2022-11-15
Feat/job counts

brosenberg42:
**Issues:**
- https://github.com/openmpf/openmpf/issues/1601

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1618)

### DIFF
--- a/trunk/mpf-rest-api/src/main/java/org/mitre/mpf/rest/api/JobPageListModel.java
+++ b/trunk/mpf-rest-api/src/main/java/org/mitre/mpf/rest/api/JobPageListModel.java
@@ -25,70 +25,9 @@
  ******************************************************************************/
 
 package org.mitre.mpf.rest.api;
-import java.util.ArrayList;
+
 import java.util.List;
 
-
-public class JobPageListModel {
-	private List<JobPageModel> data = new ArrayList<JobPageModel>();
-	private Integer draw;
-	private int recordsTotal = 0;
-	private int  recordsFiltered = 0;
-	private String error  = null;
-
-	public JobPageListModel() {}
-
-	public JobPageListModel(Integer draw, int recordsTotal, int recordsFiltered,String error,
-	                        List<JobPageModel> jobs) {
-		this.draw = draw;
-		this.recordsTotal = recordsTotal;
-		this.recordsFiltered = recordsFiltered;
-		this.error = error;
-		this.data = jobs;
-	}
-
-	public List<JobPageModel> getData() {
-		return data;
-	}
-
-	public void addData(JobPageModel job) {
-		this.data.add(job);
-	}
-
-	public void setData(List<JobPageModel> data) {
-		this.data = data;
-	}
-
-	public Integer getDraw() {
-		return draw;
-	}
-
-	public void setDraw(Integer draw) {
-		this.draw = draw;
-	}
-
-	public int getRecordsTotal() {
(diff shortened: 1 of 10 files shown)


jrobble (changes requested):
> Reviewed 10 of 10 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1618)* status: all files reviewed, 1 unresolved discussion (waiting on @brosenberg42)

___
*[`trunk/workflow-manager/src/main/webapp/resources/mytheme/js/angular_app/app.js` line 69 at r1](https://reviewable.io/reviews/openmpf/openmpf/1618#-NH0qwNrAvpn8Z001uRH:-NH0qwNsCr9e0tIO4aum:b-9ugzy8) ([raw file](https://github.com/openmpf/openmpf/blob/706a6cab3795def549e94ed228fe686815b78364/trunk/workflow-manager/src/main/webapp/resources/mytheme/js/angular_app/app.js#L69)):*
> ```JavaScript
> 	  $urlRouterProvider.otherwise("/jobs");
> 
> 	  // Matches URLs with no fragment. After a user initally logs in, in the URL will not have a
> ```

I think you have an extra "in" in "in, in the URL".

brosenberg42 (commented):
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1618)* status: 8 of 10 files reviewed, 2 unresolved discussions (waiting on @jrobble)

___
*[`trunk/workflow-manager/src/main/webapp/resources/mytheme/js/angular_app/app.js` line 69 at r1](https://reviewable.io/reviews/openmpf/openmpf/1618#-NH0qwNrAvpn8Z001uRH:-NH4Y8E3AwAodIKac0cK:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/706a6cab3795def549e94ed228fe686815b78364/trunk/workflow-manager/src/main/webapp/resources/mytheme/js/angular_app/app.js#L69)):*

I think you have an extra "in" in "in, in the URL".
</blockquote></details>

Done.
___
*[`trunk/workflow-manager/src/main/webapp/resources/mytheme/js/controllers/JobsCtrl.js` line 317 at r1](https://reviewable.io/reviews/openmpf/openmpf/1618#-NH1bRyk2VipXFYLhgAY:-NH4_iee7FMJn6_dTv8o:b-stzwjm) ([raw file](https://github.com/openmpf/openmpf/blob/706a6cab3795def549e94ed228fe686815b78364/trunk/workflow-manager/src/main/webapp/resources/mytheme/js/controllers/JobsCtrl.js#L317)):*

It seems a little strange to say "Showing 51 to 59 of 114 total entries" when there are really only 59 filtered results. Can we just say "Showing X to Y entries"?

Also seems a little strange to click on page 12 and then end up on page 6 when looking at filtered results.

I wonder if it's better to just not show the total number of entries when doing a search and just have the user click "Next" through the pages one-by-one. That means not showing the final page after the "..." since we don't know what the right number should be (unless you can call it "Last" instead of a number).
</blockquote></details>

The old message used to mention the total number of entries, so I tried to keep that behavior. The message does say "total".

One issue is that we can't change the pagination type specifically for when a filter is being applied. It is set when the table is initialized and can't be changed afterwards. Also, even when a filter is applied, being able to jump forward a few pages is useful.

PR #1655, opened 2023-03-02
Use startup.auto.registration.skip.spring=false.

jrobble:
**Issues:**

- https://github.com/openmpf/openmpf/issues/1658

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1655)

### DIFF
--- a/trunk/workflow-manager/src/main/resources/properties/mpf-private-example.properties
+++ b/trunk/workflow-manager/src/main/resources/properties/mpf-private-example.properties
@@ -41,4 +41,5 @@ mpf.ansible.role.child.path=/home/mpf/openmpf-projects/openmpf/trunk/ansible/ins
 mpf.ansible.local-only=true
 node.auto.config.enabled=true
 node.auto.config.num.services.per.component=1
+startup.auto.registration.skip=false
 startup.auto.registration.skip.spring=false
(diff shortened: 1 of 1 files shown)


jrobble:
I always used to run `mvn install` with `-Dstartup.auto.registration.skip=false`, which just took care of setting `startup.auto.registration.skip.spring=false` in `mpf.properties` due to `startup.auto.registration.skip.spring=${startup.auto.registration.skip}`.

When I switched to running the WFM in VScode it was using the default value of `startup.auto.registration.skip` (true), to set `startup.auto.registration.skip.spring`. Manually setting `startup.auto.registration.skip.spring=false` in `mpf-private.properties` overrides that.

I have not found a way to specify `-Dstartup.auto.registration.skip=false` to the maven build in VScode. I looked at [these](https://github.com/Microsoft/vscode-maven) options, but none seem suitable. I tried `maven.executable.options`, but it didn't work. No need to look into it, setting `mpf-private.properties` is fine. 

All in all, I'm not going to submit a PR or change anything. This was all due to my lack of understanding and switching to VScode.

jrobble:
Reopening. Setting `startup.auto.registration.skip.spring=false` (in addition to the current `startup.auto.registration.skip=false` setting) will be useful for developers running the WFM through VScode.

brosenberg42 (approved):

Issue #1564, opened 2022-08-12
Use ffprobe's JSON output instead of regexes during media inspection


brosenberg42:
A command like `ffprobe -hide_banner -show_streams -print_format json video.mp4` can be used to get the media inspection information. Using the JSON output will less brittle than the [regexes we currently use](https://github.com/openmpf/openmpf/blob/master/trunk/workflow-manager/src/main/resources/tika-external-parsers.xml). 
- The JSON output formats the frame rate as a fraction, so https://github.com/openmpf/openmpf/issues/1342 will need to be done at the same time as this task.
- We don't use all of the keys in [tika-external-parsers.xml](https://github.com/openmpf/openmpf/blob/master/trunk/workflow-manager/src/main/resources/tika-external-parsers.xml). All of the ones that we actually use are in the JSON output. 
- The JSON output will have different objects for the video stream and audio stream, so make sure that the properties are being taken from the correct stream.

Issue #1342, opened 2021-06-01
Use ffprobe to get FPS during media inspection


jrobble:
We should consider using this to get the frame rate:

````
[mpf@openmpf-master models]$ ffprobe -hide_banner -select_streams v -show_entries stream=r_frame_rate -print_format default=noprint_wrappers=1:nokey=1 /home/mpf/Desktop/TMP/DELETEME/zero-sized-detections.mp4 2>/dev/null
412000/179989
```

412000/179989 = 2.28902877398

Currently, we use:

```
[mpf@openmpf-master models]$ ffmpeg -nostats -i /home/mpf/Desktop/TMP/DELETEME/zero-sized-detections.mp4 -map 0:v:0 -c copy -f null -
Stream #0:0(und): Video: h264 (Constrained Baseline) (avc1 / 0x31637661), yuv420p, 1920x1080, 777 kb/s, 2.29 fps, 2.29 tbr, 412k tbn, 824k tbc (default)
```

The difference between 2.28902877398 and 2.29 is small, but more meaningful the longer the video. If this media 4 hours long, the drift would be 3 seconds.

We need to figure out if `r_frame_rate` is the most appropriate value to use, and that whatever we choose works on a variety of video types, including CFR and VFR. 

More generally, this also begs the question of whether we should be using the ffmpeg Tika external parser for anything.

PR #1660, opened 2023-03-07
Feat/ffprobe json

brosenberg42:
**Issues:**
- https://github.com/openmpf/openmpf/issues/1564
- https://github.com/openmpf/openmpf/issues/1342

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1660)


jrobble (changes requested):
> Reviewed 34 of 34 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1660)* status: all files reviewed, 4 unresolved discussions (waiting on @brosenberg42)

___
*[`trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/mediainspection/Fraction.java` line 38 at r1](https://reviewable.io/reviews/openmpf/openmpf/1660#-NSHzvRZBf9UQXS9cBcr:-NSHzvRj55YQBDN_N_G8:bsw15vm) ([raw file](https://github.com/openmpf/openmpf/blob/4574eb81c12e0c7be62ad47064fba8ef86037f96/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/mediainspection/Fraction.java#L38)):*
> ```Java
> 
>     public Fraction {
>         if (denominator < 0) {
> ```

Why would the denominator be a negative number? In what cases would we get one?
___
*[`trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/FrameTimeInfo.java` line 125 at r1](https://reviewable.io/reviews/openmpf/openmpf/1660#-NSLtXC62jhvSeXNnlWV:-NSLtXCA1NVI92vU51-S:b-2esndw) ([raw file](https://github.com/openmpf/openmpf/blob/4574eb81c12e0c7be62ad47064fba8ef86037f96/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/FrameTimeInfo.java#L125)):*
> ```Java
> 
> 
>     private static FrameTimeInfo fpsFrameTimeInfo(
> ```

Regarding both `private static FrameTimeInfo fpsFrameTimeInfo()` methods, it seems a little strange that they have both a `startTime` and `requiresTimeEstimation` parameter. Is there ever a time where `requiresTimeEstimation=true` and `startTime` is not 0?

It seems like `startTime` could be optional and when it's empty you can infer that time estimation is required. Just a suggestion.
___
*[`trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/MediaTypeUtils.java` line 82 at r1](https://reviewable.io/reviews/openmpf/openmpf/1660#-NSLxnNk7g2I6yl1SJT2:-NSLxnNrEbO8vfjrSSV6:bkbwaj2) ([raw file](https://github.com/openmpf/openmpf/blob/4574eb81c12e0c7be62ad47064fba8ef86037f96/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/MediaTypeUtils.java#L82)):*
> ```Java
>             catch (IllegalArgumentException e) {
>                 LOG.error(
>                         "The \"{}\" properties from the media type white list file contained the invalid value of \"{}\".",
> ```

I think  you want "property" instead of "properties" in this log message.
___
*[`trunk/workflow-manager/src/test/resources/samples/test4.mkv` line 0 at r1](https://reviewable.io/reviews/openmpf/openmpf/1660#-NSHrSS0ESP8OQGeHjZA:-NSHrSS5BU3Akm5mbhPY:bkc1n17) ([raw file](https://github.com/openmpf/openmpf/blob/4574eb81c12e0c7be62ad47064fba8ef86037f96/trunk/workflow-manager/src/test/resources/samples/test4.mkv#L0)):*
This is a large file for a unit test, but I guess it's fine if we can't find anything smaller. It's an official piece of test media, so that is an advantage to using it.
___
*[`trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/mediainspection/FfprobeMetadata.java` line 36 at r1](https://reviewable.io/reviews/openmpf/openmpf/1660#-NSHqCHi4YM6tJeXDRc3:-NSHqCHl5rubcfjubnpz:b-1pnh7e) ([raw file](https://github.com/openmpf/openmpf/blob/4574eb81c12e0c7be62ad47064fba8ef86037f96/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/mediainspection/FfprobeMetadata.java#L36)):*
> ```Java
> public record FfprobeMetadata(Optional<Video> video, Optional<Audio> audio) {
> 
>     public record Image(OptionalInt width, OptionalInt height, OptionalInt exifOrientation) {}
> ```

When we create these records we should know the values for each metadata field, right? Do they need to be Optional?

Width and height are required for an image. I wouldn't consider them optional. Similarly, frame count and duration are required for video streams, and duration is required for audio streams.

brosenberg42 (commented):
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1660)* status: 25 of 36 files reviewed, 4 unresolved discussions (waiting on @brosenberg42 and @jrobble)

___
*[`trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/mediainspection/Fraction.java` line 38 at r1](https://reviewable.io/reviews/openmpf/openmpf/1660#-NSHzvRZBf9UQXS9cBcr:-NSQUT6W-_iGTwGEMEc6:b-160g4r) ([raw file](https://github.com/openmpf/openmpf/blob/4574eb81c12e0c7be62ad47064fba8ef86037f96/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/mediainspection/Fraction.java#L38)):*

Why would the denominator be a negative number? In what cases would we get one?
</blockquote></details>

Normalizing a fraction so that the denominator is always positive is a typical thing to do. `2/3 * -1/-2 = -1/-3` looks weird. There is nothing stopping someone from doing `new Fraction(1, -3)`. It would be strange for a fraction class to only behave correctly when the parameters come from ffprobe.
___
*[`trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/FrameTimeInfo.java` line 125 at r1](https://reviewable.io/reviews/openmpf/openmpf/1660#-NSLtXC62jhvSeXNnlWV:-NSRKuSzBxwI6QD0tr1A:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/4574eb81c12e0c7be62ad47064fba8ef86037f96/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/FrameTimeInfo.java#L125)):*

Regarding both `private static FrameTimeInfo fpsFrameTimeInfo()` methods, it seems a little strange that they have both a `startTime` and `requiresTimeEstimation` parameter. Is there ever a time where `requiresTimeEstimation=true` and `startTime` is not 0?

It seems like `startTime` could be optional and when it's empty you can infer that time estimation is required. Just a suggestion.
</blockquote></details>

Done.
___
*[`trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/MediaTypeUtils.java` line 82 at r1](https://reviewable.io/reviews/openmpf/openmpf/1660#-NSLxnNk7g2I6yl1SJT2:-NSQosc526au8kkZFT_W:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/4574eb81c12e0c7be62ad47064fba8ef86037f96/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/MediaTypeUtils.java#L82)):*

I think  you want "property" instead of "properties" in this log message.
</blockquote></details>

Done.
___
*[`trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/mediainspection/FfprobeMetadata.java` line 36 at r1](https://reviewable.io/reviews/openmpf/openmpf/1660#-NSHqCHi4YM6tJeXDRc3:-NSQs-uBEERNZbdLpb4o:bylprst) ([raw file](https://github.com/openmpf/openmpf/blob/4574eb81c12e0c7be62ad47064fba8ef86037f96/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/mediainspection/FfprobeMetadata.java#L36)):*

When we create these records we should know the values for each metadata field, right? Do they need to be Optional?

Width and height are required for an image. I wouldn't consider them optional. Similarly, frame count and duration are required for video streams, and duration is required for audio streams. 
</blockquote></details>

Optional just means that the value may not be present, it does not mean "not required". The exact output of ffprobe varies based on the content of the media file, so certain properties that we access may not be present. The code that accesses these fields is what determines whether or not it considers them "required". Using OptionalInt forces the code that accesses it to decide how to handle the case where it is empty. It is `MediaInspectionHelper`'s job to determine whether or not it can collect enough information for a job to proceed. The duration for audio files isn't required because we don't segment them.

jrobble (approved):
> Reviewed 11 of 11 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1660)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @brosenberg42)

___
*[`trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/mediainspection/Fraction.java` line 38 at r1](https://reviewable.io/reviews/openmpf/openmpf/1660#-NSHzvRZBf9UQXS9cBcr:-NSgdAaF-rMxN-5A8c5p:b-941iew) ([raw file](https://github.com/openmpf/openmpf/blob/4574eb81c12e0c7be62ad47064fba8ef86037f96/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/mediainspection/Fraction.java#L38)):*

Normalizing a fraction so that the denominator is always positive is a typical thing to do. `2/3 * -1/-2 = -1/-3` looks weird. There is nothing stopping someone from doing `new Fraction(1, -3)`. It would be strange for a fraction class to only behave correctly when the parameters come from ffprobe.
</blockquote></details>

Ok. Just curious if there were some strange situation in which you saw a negative denominator when working with ffprobe.
___
*[`trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/FrameTimeInfo.java` line 125 at r1](https://reviewable.io/reviews/openmpf/openmpf/1660#-NSLtXC62jhvSeXNnlWV:-NSghJesArQkQY_88NPj:bt1edu6) ([raw file](https://github.com/openmpf/openmpf/blob/4574eb81c12e0c7be62ad47064fba8ef86037f96/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/FrameTimeInfo.java#L125)):*

Done.
</blockquote></details>

Nice refactor.
___
*[`trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/mediainspection/FfprobeMetadata.java` line 36 at r1](https://reviewable.io/reviews/openmpf/openmpf/1660#-NSHqCHi4YM6tJeXDRc3:-NSgaSJG8abqT4YDPGfM:b-tchy8j) ([raw file](https://github.com/openmpf/openmpf/blob/4574eb81c12e0c7be62ad47064fba8ef86037f96/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/mediainspection/FfprobeMetadata.java#L36)):*

Optional just means that the value may not be present, it does not mean "not required". The exact output of ffprobe varies based on the content of the media file, so certain properties that we access may not be present. The code that accesses these fields is what determines whether or not it considers them "required". Using OptionalInt forces the code that accesses it to decide how to handle the case where it is empty. It is `MediaInspectionHelper`'s job to determine whether or not it can collect enough information for a job to proceed. The duration for audio files isn't required because we don't segment them. 
</blockquote></details>

Ok. I think what's throwing me off is that `width` and `height` aren't `OptionalInt` for videos. I think the difference is that for images we fall back to the BufferedImage approach when we can't otherwise determine the width and height from ffprobe. In your approach `FfprobeMetadata.Image` needs a way to represent that those values could not be determined. That's fine. I was just trying to understand it better.

Issue #1650, opened 2023-02-23
Create TiesDb records for thumbnail jobs under the parent media


jrobble:
Job producers may submit thumbnails to OpenMPF when there is an issue with the original media. These would be high quality thumbnails that are close to the original media resolution. The advantage of submitting the thumbnail is that it's converted into a format that can be processed.

When the thumbnails are processed by OpenMPF we need to create the TiesDb record (supplemental) for the original media. That way when other job producers submit jobs to OpenMPF using the original media we can check TiesDb and pull the record for the job run on the thumbnail.

Whenever a job producer submits a thumbnail in place of the original media they should provide an additional job / media metadata property, such as `LINKED_MEDIA_HASH` (or maybe `PRIMARY_MEDIA_HASH`), that refers to the hash of the original media. When we create the TiesDb record for the original media, we need to add fields for the file URL and hash of the thumbnail media. That way someone looking at the record can tell the results weren't generated using the original media, which could help with debugging.

Note that this assumes that all job producers will know the original media hash when submitting the thumbnail and that they won't submit the thumbnail without specifying the original media hash, or else that would require a separate entry in TiesDb for the thumbnail media. (We don't want separate entries in TiesDb for thumbnail media because that would be a lot of entries.)

Related to https://github.com/openmpf/openmpf/issues/1613.

PR #1673, opened 2023-04-19
Add support for linked media

brosenberg42:
**Issues:**
- https://github.com/openmpf/openmpf/issues/1650

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1673)

### DIFF
--- a/trunk/interop/src/main/java/org/mitre/mpf/interop/JsonMediaOutputObject.java
+++ b/trunk/interop/src/main/java/org/mitre/mpf/interop/JsonMediaOutputObject.java
@@ -35,8 +35,9 @@ import static org.mitre.mpf.interop.util.CompareUtils.stringCompare;
 
 @JsonTypeName("MediaOutputObject")
 @JsonPropertyOrder({
-		"mediaId", "parentMediaId", "path", "sha256", "mimeType", "mediaType", "length", "frameRanges", "timeRanges",
-		"mediaMetadata", "mediaProperties", "status", "detectionProcessingErrors", "markupResult", "output"})
+		"mediaId", "parentMediaId", "path", "tiesDbSourceMediaPath", "sha256", "mimeType",
+		"mediaType", "length", "frameRanges", "timeRanges", "mediaMetadata", "mediaProperties",
+		"status", "detectionProcessingErrors", "markupResult", "output"})
 public class JsonMediaOutputObject implements Comparable<JsonMediaOutputObject> {
 
 	@JsonProperty("mediaId")
@@ -53,6 +54,12 @@ public class JsonMediaOutputObject implements Comparable<JsonMediaOutputObject> 
 	@JsonPropertyDescription("The URI to this media file.")
 	private String path;
 	public String getPath() { return path; }
+
+	@JsonInclude(JsonInclude.Include.NON_DEFAULT)
+    @JsonPropertyDescription("If the job didn't run because it was possible to get the results "
+        + "from TiesDb, this will be set to the path of the media used to generate those results.")
+	private String tiesDbSourceMediaPath;
+	public String getTiesDbSourceMediaPath() { return tiesDbSourceMediaPath; }
 
 	@JsonProperty("mediaType")
 	@JsonPropertyDescription("The type associated with this media file. For example, \"VIDEO\".")
@@ -118,11 +125,13 @@ public class JsonMediaOutputObject implements Comparable<JsonMediaOutputObject> 
 	private SortedMap<String, SortedSet<JsonDetectionProcessingError>> detectionProcessingErrors;
 	public SortedMap<String, SortedSet<JsonDetectionProcessingError>> getDetectionProcessingErrors() { return detectionProcessingErrors; }
 
-	public JsonMediaOutputObject(long mediaId, long parentMediaId, String path, String mediaType, String mimeType,
-								 int length, String sha256, String status) {
+	public JsonMediaOutputObject(
+			long mediaId, long parentMediaId, String path, String tiesDbSourceMediaPath,
+			String mediaType, String mimeType, int length, String sha256, String status) {
 		this.mediaId = mediaId;
 		this.parentMediaId = parentMediaId;
 		this.path = path;
+		this.tiesDbSourceMediaPath = tiesDbSourceMediaPath;
 		this.type = mediaType;
 		this.mimeType = mimeType;
 		this.length = length;
@@ -143,6 +152,7 @@ public class JsonMediaOutputObject implements Comparable<JsonMediaOutputObject> 
 			@JsonProperty("mediaId") long mediaId,
 			@JsonProperty("parentMediaId") long parentMediaId,
 			@JsonProperty("path") String path,
+			@JsonProperty("tiesDbSourceMediaPath") String tiesDbSourceMediaPath,
 			@JsonProperty("mediaType") String mediaType,
 			@JsonProperty("mimeType") String mimeType,
 			@JsonProperty("length") int length,
@@ -155,8 +165,9 @@ public class JsonMediaOutputObject implements Comparable<JsonMediaOutputObject> 
 			@JsonProperty("markupResult") JsonMarkupOutputObject markupResult,
 			@JsonProperty("output") SortedMap<String, SortedSet<JsonActionOutputObject>> detectionTypes,
 			@JsonProperty("detectionProcessingErrors") SortedMap<String, SortedSet<JsonDetectionProcessingError>> detectionProcessingErrors) {
-		JsonMediaOutputObject jsonMediaOutputObject =
-				new JsonMediaOutputObject(mediaId, parentMediaId, path, mediaType, mimeType, length, sha256, status);
+		JsonMediaOutputObject jsonMediaOutputObject = new JsonMediaOutputObject(
+				mediaId, parentMediaId, path, tiesDbSourceMediaPath, mediaType, mimeType, length,
+				sha256, status);
 		jsonMediaOutputObject.markupResult = markupResult;
 
 		if (frameRanges != null) {
--- a/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/JobCompleteProcessorImpl.java
+++ b/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/JobCompleteProcessorImpl.java
@@ -425,6 +425,7 @@ public class JobCompleteProcessorImpl extends WfmProcessor implements JobComplet
                     media.getId(),
                     media.getParentId(),
                     media.getPersistentUri(),
+                    null,
                     media.getType().map(Enum::toString).orElse(null),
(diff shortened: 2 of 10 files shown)


jrobble (changes requested):
Why can't they all be this simple? :)

Please update the TiesDb Guide to mention how the `LINKED_MEDIA_HASH` property can be used.

> Reviewed 8 of 8 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1673#-:-NTPvcUs7nniHGFtiT9x:bp1zg2j)* status: all files reviewed, 1 unresolved discussion (waiting on @brosenberg42)

___
*[`trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/persistent/MediaImpl.java` line 243 at r1](https://reviewable.io/reviews/openmpf/openmpf/1673#-NTQ2Ex6BE02JH1hoiyi:-NTQ2Ex72tOZoZc2Hrkh:blwk8iu) ([raw file](https://github.com/openmpf/openmpf/blob/670a4cf4a7ea97420e53da2516555c9a7139d225/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/persistent/MediaImpl.java#L243)):*
> ```Java
>     @Override
>     @JsonIgnore
>     public Optional<String> getLinkedHash() {
> ```

I ran a job with S3 copy enabled. It found a record in TiesDb. This is the JSON output:

```json
   "media": [
      {
         "mediaId": 17,
         "parentMediaId": -1,
         "path": "file:///home/mpf/git/openmpf-projects/openmpf/trunk/install/share/remote-media/at3_1m4_01.tif",
         "sha256": "820258067d4211af126d2852f444e0d84bf78696ccffe1869f71236ef035fc72",
         "mimeType": "image/jpeg",
         "mediaType": "IMAGE",
         "length": 1,
         "frameRanges": [],
         "timeRanges": [],
         "mediaMetadata": {
            "EXIF_ORIENTATION": "1",
            "FRAME_HEIGHT": "360",
            "FRAME_WIDTH": "480",
            "HORIZONTAL_FLIP": "FALSE",
            "MIME_TYPE": "image/jpeg",
            "ROTATION": "0"
         },
         "mediaProperties": {
            "LINKED_MEDIA_HASH": "820258067d4211af126d2852f444e0d84bf78696ccffe1869f71236ef035fc72"
         },
```

The `path` shows the image for the most recent job I submitted, but all of the other information in the `media` section, minus `mediaProperties`, is for the image associated with the original job.

I think that `path` should refer to the original media path.

---

Something similar happens when creating a job using a non-existent (bogus) media file in S3. A user can provide a `MEDIA_HASH` that corresponds to valid piece of media that has a record in TiesDb, a `MIME_TYPE` with a valid prefix like `image/wow`, and a bogus `mediaUri`, and the job will complete.

In that case the `media` section in the JSON will use the bogus `mediaUri` but use the rest of the info for the media in the original job.

I'm not saying this is a problem. I just wanted to point it out. If the user does something like that they should expect strange results.

The difference between this bogus media case and the "LINKED_MEDIA_HASH" one I explained above is that in the above case the media in the second job is understandably a different piece of media than in the original job, whereas in the bogus media case it's supposed to the same piece of media.

jrobble:
*[`trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/persistent/MediaImpl.java` line 243 at r1](https://reviewable.io/reviews/openmpf/openmpf/1673#-NTQ2Ex6BE02JH1hoiyi:-NTR7RerEw8gPQqQCecB:b-t77b0x) ([raw file](https://github.com/openmpf/openmpf/blob/670a4cf4a7ea97420e53da2516555c9a7139d225/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/persistent/MediaImpl.java#L243)):*

I ran a job with S3 copy enabled. It found a record in TiesDb. This is the JSON output:

```json
   "media": [
      {
         "mediaId": 17,
         "parentMediaId": -1,
         "path": "file:///home/mpf/git/openmpf-projects/openmpf/trunk/install/share/remote-media/at3_1m4_01.tif",
         "sha256": "820258067d4211af126d2852f444e0d84bf78696ccffe1869f71236ef035fc72",
         "mimeType": "image/jpeg",
         "mediaType": "IMAGE",
         "length": 1,
         "frameRanges": [],
         "timeRanges": [],
         "mediaMetadata": {
            "EXIF_ORIENTATION": "1",
            "FRAME_HEIGHT": "360",
            "FRAME_WIDTH": "480",
            "HORIZONTAL_FLIP": "FALSE",
            "MIME_TYPE": "image/jpeg",
            "ROTATION": "0"
         },
         "mediaProperties": {
            "LINKED_MEDIA_HASH": "820258067d4211af126d2852f444e0d84bf78696ccffe1869f71236ef035fc72"
         },
```

The `path` shows the image for the most recent job I submitted, but all of the other information in the `media` section, minus `mediaProperties`, is for the image associated with the original job.

I think that `path` should refer to the original media path.

---

Something similar happens when creating a job using a non-existent (bogus) media file in S3. A user can provide a `MEDIA_HASH` that corresponds to valid piece of media that has a record in TiesDb, a `MIME_TYPE` with a valid prefix like `image/wow`, and a bogus `mediaUri`, and the job will complete.

In that case the `media` section in the JSON will use the bogus `mediaUri` but use the rest of the info for the media in the original job.

I'm not saying this is a problem. I just wanted to point it out. If the user does something like that they should expect strange results.

The difference between this bogus media case and the "LINKED_MEDIA_HASH" one I explained above is that in the above case the media in the second job is understandably a different piece of media than in the original job, whereas in the bogus media case it's supposed to the same piece of media.

</blockquote></details>

TBH, I'm not sure what the best thing to do is for the "LINKED_MEDIA_HASH" case. A few thoughts:

- `path` seems like it should refer to the media described by the other info in the `media` section.
- We want to include the path of the media associated with the new job somewhere in the JSON output, even if that media wasn't used for anything.

One idea is to include a new `jobRequestMediaPath` in the media section that refers to the media path associated with the job request whenever we decide to use a record in TiesDb. When we use a record we always populate `path` with the value of the media path used to generate the original JSON output in that record.

I like this approach because a user looking at the JSON will know exactly what media was used to generate the JSON regardless if "LINKED_MEDIA_HASH" was provided or "MEDIA_HASH" + "MIME_TYPE" metadata was provided.

jrobble:
*[`trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/persistent/MediaImpl.java` line 243 at r1](https://reviewable.io/reviews/openmpf/openmpf/1673#-NTQ2Ex6BE02JH1hoiyi:-NTREBgu9nik4yBk61JG:b-kumhw7) ([raw file](https://github.com/openmpf/openmpf/blob/670a4cf4a7ea97420e53da2516555c9a7139d225/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/persistent/MediaImpl.java#L243)):*

TBH, I'm not sure what the best thing to do is for the "LINKED_MEDIA_HASH" case. A few thoughts:

- `path` seems like it should refer to the media described by the other info in the `media` section.
- We want to include the path of the media associated with the new job somewhere in the JSON output, even if that media wasn't used for anything.

One idea is to include a new `jobRequestMediaPath` in the media section that refers to the media path associated with the job request whenever we decide to use a record in TiesDb. When we use a record we always populate `path` with the value of the media path used to generate the original JSON output in that record.

I like this approach because a user looking at the JSON will know exactly what media was used to generate the JSON regardless if "LINKED_MEDIA_HASH" was provided or "MEDIA_HASH" + "MIME_TYPE" metadata was provided.



</blockquote></details>

Also, I'm wondering if we should call the property `TIES_DB_LINKED_MEDIA_HASH` because it only makes sense for TiesDb. Originally I was thinking it would be specified as media metadata, but now that you've made it a media property it seems like it should have the `TIES_DB_` prefix like the other properties.

brosenberg42 (commented):
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1673)* status: all files reviewed, 1 unresolved discussion (waiting on @jrobble)

___
*[`trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/persistent/MediaImpl.java` line 243 at r1](https://reviewable.io/reviews/openmpf/openmpf/1673#-NTQ2Ex6BE02JH1hoiyi:-NTTHl7V95WR9Hotv-S-:bpil3rq) ([raw file](https://github.com/openmpf/openmpf/blob/670a4cf4a7ea97420e53da2516555c9a7139d225/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/persistent/MediaImpl.java#L243)):*

Also, I'm wondering if we should call the property `TIES_DB_LINKED_MEDIA_HASH` because it only makes sense for TiesDb. Originally I was thinking it would be specified as media metadata, but now that you've made it a media property it seems like it should have the `TIES_DB_` prefix like the other properties.
</blockquote></details>

I think `path` should always be the path the job was submitted with. If a job is submitted with multiple media entries, using the same path as the submission makes it easier to figure out which media object in the output is associated with the media input. When a user is providing the linked media hash, they are saying that the provided media and the linked media are interchangeable. Setting path to the value from TiesDb would be inconsistent with how the job id refers to the new job id.

Regarding the name of the property, I think keeping the name short in the job request is better, but it probably makes sense to rename this method.

jrobble:
*[`trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/persistent/MediaImpl.java` line 243 at r1](https://reviewable.io/reviews/openmpf/openmpf/1673#-NTQ2Ex6BE02JH1hoiyi:-NTZT5HU55l-lZkkahu4:b-v21ygr) ([raw file](https://github.com/openmpf/openmpf/blob/670a4cf4a7ea97420e53da2516555c9a7139d225/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/persistent/MediaImpl.java#L243)):*

I think `path` should always be the path the job was submitted with. If a job is submitted with multiple media entries, using the same path as the submission makes it easier to figure out which media object in the output is associated with the media input. When a user is providing the linked media hash, they are saying that the provided media and the linked media are interchangeable. Setting path to the value from TiesDb would be inconsistent with how the job id refers to the new job id.

Regarding the name of the property, I think keeping the name short in the job request is better, but it probably makes sense to rename this method.
</blockquote></details>

As discussed, let's add a `tiesDbSourceMediaPath` underneath the `path` field in the media entry whenever we use a record from TiesDb. It will be consistent with the use of `tiesDbSourceJobId`.

Also, as discussed, let's continue using `LINKED_MEDIA_HASH` as the property name.

jrobble (approved):
> Reviewed 4 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1673)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @brosenberg42)

Issue #1678, opened 2023-05-03
ffmpeg hangs when processing some kinds of unsupported/corrupted media


jrobble:
**Update 5/30/2023:**

Related to https://github.com/openmpf/openmpf/issues/1687.

The ffprobe command mentioned below doesn't hang in the OpenMPF 7.1.12 and 7.2.0 Docker images because it uses a more recent version of ffprobe/ffmpeg than what I had installed on my dev. VM when I created this issue.

Now what's happening is that the Workflow Manager in 7.1.12 is locking up in `FrameCounter.countNative()`:

```
# DEBUG: cpu time is 4.29 hours?
"Camel (camelContext) thread #142 - JmsConsumer[MPF.MEDIA_INSPECTION_WORK_QUEUE]" #296 daemon prio=5 os_prio=0 cpu=15439603.43ms elapsed=15614.40s tid=0x00007efe78013eb0 nid=0x5ea runnable  [0x00007efd724ee000]
   java.lang.Thread.State: RUNNABLE
    at org.mitre.mpf.framecounter.FrameCounter.countNative(Native Method)
    at org.mitre.mpf.framecounter.FrameCounter.count(FrameCounter.java:52)
```

Even if the media were to get past media inspection it will get locked up in the component when it uses OpenCV to read the video frames. I verified this using the CLI Runner. Thus, this issue still affects 7.2.0 even through we no longer use `FrameCounter.countNative()` in that version of OpenMPF. The reason this happens is because OpenCV is compiled with ffmpeg dependencies than are older than the dependencies used to install the command-line version of ffprobe/ffmpeg we use during media inspection.

To address this issue, we should run the ffprobe command with `-loglevel error`. If it generates log messages on the err stream we then generate an OpenMPF media error so that the job fails. We should limit the number of error log lines we capture when generating the OpenMPF error message. We should consider 5 lines and make it a configurable system property.

--- 

The issue appears to be with the Tika parser. Specifically, this ffmpeg command being used by Tika hangs:

```
ffmpeg -nostats -i b1xxx.mp4 -map 0:v:0 -c copy -f null -
ffmpeg version 4.2.7-0ubuntu0.1 Copyright (c) 2000-2022 the FFmpeg developers
  built with gcc 9 (Ubuntu 9.4.0-1ubuntu1~20.04.1)
[...]

PR #1688, opened 2023-05-31
Check ffprobe's error logging

brosenberg42:
**Issues:**
- https://github.com/openmpf/openmpf/issues/1678

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1688)


jrobble (changes requested):
> Reviewed 8 of 8 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1688)* status: all files reviewed, 2 unresolved discussions (waiting on @brosenberg42)

___
*[`trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/mediainspection/FfprobeMetadataExtractor.java` line 468 at r1](https://reviewable.io/reviews/openmpf/openmpf/1688#-NWoZ_XE8S0-gS8w24G_:-NWoZ_XF87qivG4cxQt-:b-98vsqs) ([raw file](https://github.com/openmpf/openmpf/blob/9dd1fd0aa52bb72c538075508c98f8f41aefe759/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/mediainspection/FfprobeMetadataExtractor.java#L468)):*
> ```Java
>     private int getNumStdErrLines(BatchJob job, Media media) {
>         try {
>             return Integer.parseInt(_aggregateJobPropertiesUtil.getValue(
> ```

Check that the value is > 0.
___
*[`trunk/workflow-manager/src/main/resources/workflow-properties.json` line 486 at r1](https://reviewable.io/reviews/openmpf/openmpf/1688#-NWotMTm0t5X2COJmh4M:-NWotMTn4a0AXfxP180I:b-xubhl) ([raw file](https://github.com/openmpf/openmpf/blob/9dd1fd0aa52bb72c538075508c98f8f41aefe759/trunk/workflow-manager/src/main/resources/workflow-properties.json#L486)):*
> ```JSON
>     },
>     {
>         "name": "FFPROBE_STDERR_NUM_LINES",
> ```

Add this to `ties-db-check-ignorable-props.json`.

jrobble (changes requested):
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1688)* status: all files reviewed, 3 unresolved discussions (waiting on @brosenberg42)

___
*[`trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/mediainspection/FfprobeMetadataExtractor.java` line 79 at r1](https://reviewable.io/reviews/openmpf/openmpf/1688#-NWnuHjnBMfCsk_WyHz6:-NWnuHjnBMfCsk_WyHz7:b-x8jf7k) ([raw file](https://github.com/openmpf/openmpf/blob/9dd1fd0aa52bb72c538075508c98f8f41aefe759/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/mediainspection/FfprobeMetadataExtractor.java#L79)):*
> ```Java
>         }
>         catch (MediaInspectionException e) {
>             if (shouldIgnoreStdErr(job, media)) {
> ```

A MediaInspectionException will be thrown if:

1. We can't start the ffprobe process
2. We can't read the output stream
3. The exit code is not zero
4. There is content in the error stream and `ffprobe.ignore.stderr` is false

Even if `ffprobe.ignore.stderr` is true, I think we should at least log a warning message if any of these things occurs. That includes logging the decoder error lines that would otherwise have shown up in the JSON output object. 

However, I then ran some more tests which lead me to believe we should always treat 1, 2, and 3 as critical errors for audio and video files. Read below.

I modified the code to return a non-zero ffprobe exit code. When running an image job it ends in `COMPLETE_WITH_ERRORS` and the exit code error shows up on the JSON output, as expected. Then happens when `ffprobe.ignore.stderr` is true or false. Right now the behavior is inconsistent between images and videos. For videos, when `ffprobe.ignore.stderr` is true, the job ends in `COMPLETE_WITH_ERRORS` due to:

```
"code": "UNSUPPORTED_DATA_TYPE",
"message": "The detection component does not support detection data type of UNKNOWN"
```

I know that end users find that confusing. TBH, this line of code seems a bit odd:

```
return new FfprobeMetadata(Optional.empty(), Optional.empty());
```

When ffprobe fails to run, meaning we don't get any meaningful JSON output, it seems like we should treat that as a critical error, at least for audio and video files. Image inspection has Tika and BufferedImage as a fallback. There are no fallback options for audio and video files.

jrobble (approved):
> Reviewed 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1688)* status: all files reviewed, 1 unresolved discussion (waiting on @brosenberg42)

jrobble (commented):
> Reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1688)* status: 7 of 9 files reviewed, all discussions resolved (waiting on @brosenberg42)

___
*[`trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/camelOps/TestMediaInspectionProcessor.java` line 362 at r3](https://reviewable.io/reviews/openmpf/openmpf/1688#-NWrwASL-wo_0GS-ubQ5:-NWrwASL-wo_0GS-ubQ6:buufyja) ([raw file](https://github.com/openmpf/openmpf/blob/521bf3535fc55a0c2ff4abc0ffb2f4262728f00e/trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/camelOps/TestMediaInspectionProcessor.java#L362)):*
Please rename this to `testInvalidVideo` and the log line below to:

> "Starting media inspection test with invalid video."

---

Also, please add a separate test called `testVideoToUnknownFallback` that uses a video container format that has no audio or video stream and actually results in the media being treated as the UNKNOWN type.

jrobble (changes requested):
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1688)* status: 7 of 9 files reviewed, 1 unresolved discussion (waiting on @brosenberg42)

jrobble (approved):
> Reviewed 1 of 2 files at r3, 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1688)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @brosenberg42)

PR #1753, opened 2024-01-23
Feat/component timing

brosenberg42:
**Issues:**
- https://github.com/openmpf/openmpf/issues/1754

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1753)


jrobble:
*[`trunk/interop/src/main/java/org/mitre/mpf/interop/JsonTiming.java` line 43 at r1](https://reviewable.io/reviews/openmpf/openmpf/1753#-Np0mRCdE3U2Bnw2NLuv:-Np0mRCdE3U2Bnw2NLuw:bo9sgce) ([raw file](https://github.com/openmpf/openmpf/blob/d7f0fe6dd1e77c592e68f5c614976da70ed36c68/trunk/interop/src/main/java/org/mitre/mpf/interop/JsonTiming.java#L43)):*
> ```Java
> 
> @JsonPropertyOrder({"processingTime", "actions"})
> public class JsonTiming implements Comparable<JsonTiming> {
> ```

Please add this timing information to the TiesDB record as well.

brosenberg42 (commented):
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1753)* status: 0 of 23 files reviewed, 1 unresolved discussion (waiting on @jrobble)

___
*[`trunk/interop/src/main/java/org/mitre/mpf/interop/JsonTiming.java` line 43 at r1](https://reviewable.io/reviews/openmpf/openmpf/1753#-Np0mRCdE3U2Bnw2NLuv:-NpLWbjg--eot4ctM7q3:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/d7f0fe6dd1e77c592e68f5c614976da70ed36c68/trunk/interop/src/main/java/org/mitre/mpf/interop/JsonTiming.java#L43)):*

Please add this timing information to the TiesDB record as well.
</blockquote></details>

Done.

jrobble (changes requested):
> Reviewed 19 of 20 files at r1, 4 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1753)* status: all files reviewed, 8 unresolved discussions (waiting on @brosenberg42)

*[a discussion](https://reviewable.io/reviews/openmpf/openmpf/1753#-Nr6Wu9j5isuiOxQiQDU:-Nr6Wu9j5isuiOxQiQDV:b-oj5ldo) (no related file):*
`detection.proto` has:

```
    // Profiling metrics.
    optional MetricsMessage metrics = 405;
```

We could set `processing_time` in `MetricsMessage` to be consistent with how we're using `time_processing` in `markup.proto`.

Seems inconsistent to use the "ProcessingTime" ActiveMQ message header for components, but a protobuf field for markup.

___
*[a discussion](https://reviewable.io/reviews/openmpf/openmpf/1753#-NrB3jUOBpTqoSQq621k:-NrB3jUOBpTqoSQq621l:b-hrnp2s) (no related file):*
The reported times do not include time spent being processed by the WFM doing things like media inspection. That could take a long time for variable frame rate videos.

The issue you linked to specifically says "component" time. Unless it's easy to also capture the WFM time (not including callbacks) then let's create a new issue for WFM CPU time. Also, please create an issue for GPU times.

___
*[a discussion](https://reviewable.io/reviews/openmpf/openmpf/1753#-NrBF9kz38lpO5Xs_I8N:-NrBF9kz38lpO5Xs_I8O:b2lri4k) (no related file):*
I discovered this bug while testing: https://github.com/openmpf/openmpf/issues/1765

It can be resolved separately, unless you want to tackle it now.

___
___
*[`trunk/detection/executor/cpp/batch/main.cpp` line 245 at r2](https://reviewable.io/reviews/openmpf/openmpf/1753#-Nr6LdmfDQw0giI7MVqT:-Nr6LdmfDQw0giI7MVqU:b-ynsfhp) ([raw file](https://github.com/openmpf/openmpf/blob/1d3b5f9ec698b8dc6ce66238b11113e78c1be940/trunk/detection/executor/cpp/batch/main.cpp#L245)):*
> ```cpp
>         logger.Info("Waiting for next job.");
>         auto job_context = job_receiver.GetJob();
>         if (!component.Supports(job_context.job_type)) {
> ```

Why bother with the health check if the job is going to be rejected? Wouldn't it be better to wait until we get a valid job and run the health check as close to that job as possible?

Just thinking that the more timely the health checks are, the more valuable they are when used to detect env. issues.
___
*[`trunk/interop/src/main/java/org/mitre/mpf/interop/JsonTiming.java` line 51 at r2](https://reviewable.io/reviews/openmpf/openmpf/1753#-Nr6N24f6Gxv5eDFDgSg:-Nr6N24gAkGYN8oXQQWL:b-6t2neo) ([raw file](https://github.com/openmpf/openmpf/blob/5cacd4f00bf4738874f6d4999ec0758a241b1fd5/trunk/interop/src/main/java/org/mitre/mpf/interop/JsonTiming.java#L51)):*
> ```Java
> 
>     @JsonPropertyDescription("The processing time for each action.")
>     private final SortedSet<JsonActionTiming> _actions;
> ```

It would be nice to store the actions based on execution order so that when you're reading the timing info in the JSON top to bottom everything is in the same order it's executed in the pipeline. Right now the actions are ordered lexicographically based on their name.

This is a nitpick and not required, put if it's simple to do then I think we should do it.
___
*[`trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/persistent/BatchJobImpl.java` line 279 at r2](https://reviewable.io/reviews/openmpf/openmpf/1753#-Nr6OtH6-frgJ31nqFbG:-Nr6OtH74bjGzGwYEPx6:b-pef5z4) ([raw file](https://github.com/openmpf/openmpf/blob/5cacd4f00bf4738874f6d4999ec0758a241b1fd5/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/persistent/BatchJobImpl.java#L279)):*
Please leave a comment that explains our motivation. Something like:

> If any component returns a time < 0 then the overall time will be -1. We do this to reflect that we have incomplete information rather than tallying up what we can.
___
*[`trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/service/TestTiesDbService.java` line 551 at r2](https://reviewable.io/reviews/openmpf/openmpf/1753#-Nr6mnh7-LhLoziW4Jts:-Nr6mnh7-LhLoziW4Jtt:b320jr2) ([raw file](https://github.com/openmpf/openmpf/blob/5cacd4f00bf4738874f6d4999ec0758a241b1fd5/trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/service/TestTiesDbService.java#L551)):*
> ```Java
>     private static TiesDbInfo createExpectedTiesDbInfo() {
>         var timing = new JsonTiming(
>                 1349,
> ```

Nitpick: `createAction()` names actions with `ACTION_`.
___
*[`trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/service/TestTiesDbService.java` line 578 at r2](https://reviewable.io/reviews/openmpf/openmpf/1753#-Nr6nGeY9JcnEZvrHZ7G:-Nr6nGeY9JcnEZvrHZ7H:b320jr2) ([raw file](https://github.com/openmpf/openmpf/blob/5cacd4f00bf4738874f6d4999ec0758a241b1fd5/trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/service/TestTiesDbService.java#L578)):*
> ```Java
>         var timing = new JsonTiming(
>                 1349,
>                 List.of(new JsonActionTiming("ACTION", 876), new JsonActionTiming("ACTION2", 473)));
> ```

Nitpick: `createAction()` names actions with `ACTION_`.

brosenberg42 (commented):
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1753)* status: all files reviewed, 8 unresolved discussions (waiting on @jrobble)

*[a discussion](https://reviewable.io/reviews/openmpf/openmpf/1753#-Nr6Wu9j5isuiOxQiQDU:-NrC5H2m7jbvrIOvJecd:bmlfuzr) (no related file):*

`detection.proto` has:

```
    // Profiling metrics.
    optional MetricsMessage metrics = 405;
```

We could set `processing_time` in `MetricsMessage` to be consistent with how we're using `time_processing` in `markup.proto`.

Seems inconsistent to use the "ProcessingTime" ActiveMQ message header for components, but a protobuf field for markup.
</blockquote></details>

It will be inconsistent either way because markup.proto doesn't use MetricsMessage. Using the header allows us to include protobuf serialization in the timing.

___
___
*[`trunk/detection/executor/cpp/batch/main.cpp` line 245 at r2](https://reviewable.io/reviews/openmpf/openmpf/1753#-Nr6LdmfDQw0giI7MVqT:-NrC3uRaC9BRZQ05zwMQ:bqj5hw) ([raw file](https://github.com/openmpf/openmpf/blob/1d3b5f9ec698b8dc6ce66238b11113e78c1be940/trunk/detection/executor/cpp/batch/main.cpp#L245)):*

Why bother with the health check if the job is going to be rejected? Wouldn't it be better to wait until we get a valid job and run the health check as close to that job as possible?

Just thinking that the more timely the health checks are, the more valuable they are when used to detect env. issues.
</blockquote></details>

I was trying to capture as much of the job processing time as possible. I can switch them, but then the call to `component.Support(...)` won't be included in the timing.

jrobble (changes requested):
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1753)* status: all files reviewed, 8 unresolved discussions (waiting on @brosenberg42)

*[a discussion](https://reviewable.io/reviews/openmpf/openmpf/1753#-Nr6Wu9j5isuiOxQiQDU:-NrCGYQN2SM7pqmO43Lw:bcetnbw) (no related file):*

It will be inconsistent either way because markup.proto doesn't use MetricsMessage. Using the header allows us to include protobuf serialization in the timing.
</blockquote></details>

Should we switch over to using the header for markup and remove the timing fields from the protobuf messages if we're not going to use them?

When we handle GPU timing do you expect to use a different header, one for CPU time and one for GPU time? Just thinking that the more complicated the timing information we need to pass back is, the more reason it should be in the protobuf.

___
___
*[`trunk/detection/executor/cpp/batch/main.cpp` line 245 at r2](https://reviewable.io/reviews/openmpf/openmpf/1753#-Nr6LdmfDQw0giI7MVqT:-NrCH9Rj6BSAqWUAhX6B:bha10f7) ([raw file](https://github.com/openmpf/openmpf/blob/1d3b5f9ec698b8dc6ce66238b11113e78c1be940/trunk/detection/executor/cpp/batch/main.cpp#L245)):*

I was trying to capture as much of the job processing time as possible. I can switch them, but then the call to `component.Support(...)` won't be included in the timing.
</blockquote></details>

Ok. I understand. I think the time to run that call is so minimal that it's not worth worrying about. I think the benefit of running the health checks before a job outweighs that benefit. Please add a comment to code explaining that when you switch it back.

brosenberg42 (commented):
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1753)* status: 14 of 25 files reviewed, 8 unresolved discussions (waiting on @jrobble)

*[a discussion](https://reviewable.io/reviews/openmpf/openmpf/1753#-NrB3jUOBpTqoSQq621k:-NrGHY6u0yOv9OhQZ7BB:b-896fix) (no related file):*

The reported times do not include time spent being processed by the WFM doing things like media inspection. That could take a long time for variable frame rate videos.

The issue you linked to specifically says "component" time. Unless it's easy to also capture the WFM time (not including callbacks) then let's create a new issue for WFM CPU time. Also, please create an issue for GPU times.
</blockquote></details>

Done.

___
___
*[`trunk/detection/executor/cpp/batch/main.cpp` line 245 at r2](https://reviewable.io/reviews/openmpf/openmpf/1753#-Nr6LdmfDQw0giI7MVqT:-NrH4u_AAh9kGWXWdhUc:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/1d3b5f9ec698b8dc6ce66238b11113e78c1be940/trunk/detection/executor/cpp/batch/main.cpp#L245)):*

Ok. I understand. I think the time to run that call is so minimal that it's not worth worrying about. I think the benefit of running the health checks before a job outweighs that benefit. Please add a comment to code explaining that when you switch it back.
</blockquote></details>

Done.
___
*[`trunk/interop/src/main/java/org/mitre/mpf/interop/JsonTiming.java` line 51 at r2](https://reviewable.io/reviews/openmpf/openmpf/1753#-Nr6N24f6Gxv5eDFDgSg:-NrGHfWhBCdEh1nJjCa1:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/5cacd4f00bf4738874f6d4999ec0758a241b1fd5/trunk/interop/src/main/java/org/mitre/mpf/interop/JsonTiming.java#L51)):*

It would be nice to store the actions based on execution order so that when you're reading the timing info in the JSON top to bottom everything is in the same order it's executed in the pipeline. Right now the actions are ordered lexicographically based on their name.

This is a nitpick and not required, put if it's simple to do then I think we should do it.
</blockquote></details>

Done.
___
*[`trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/persistent/BatchJobImpl.java` line 279 at r2](https://reviewable.io/reviews/openmpf/openmpf/1753#-Nr6OtH6-frgJ31nqFbG:-NrG1SV16uvn85PVe5mz:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/5cacd4f00bf4738874f6d4999ec0758a241b1fd5/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/persistent/BatchJobImpl.java#L279)):*

Please leave a comment that explains our motivation. Something like:

> If any component returns a time < 0 then the overall time will be -1. We do this to reflect that we have incomplete information rather than tallying up what we can.
</blockquote></details>

Done.
___
*[`trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/service/TestTiesDbService.java` line 551 at r2](https://reviewable.io/reviews/openmpf/openmpf/1753#-Nr6mnh7-LhLoziW4Jts:-NrGBikP938pCJlDNJY9:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/5cacd4f00bf4738874f6d4999ec0758a241b1fd5/trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/service/TestTiesDbService.java#L551)):*

Nitpick: `createAction()` names actions with `ACTION_`.
</blockquote></details>

Done.
___
*[`trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/service/TestTiesDbService.java` line 578 at r2](https://reviewable.io/reviews/openmpf/openmpf/1753#-Nr6nGeY9JcnEZvrHZ7G:-NrGBqlb9XOB0bviYHo7:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/5cacd4f00bf4738874f6d4999ec0758a241b1fd5/trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/service/TestTiesDbService.java#L578)):*

Nitpick: `createAction()` names actions with `ACTION_`.
</blockquote></details>

Done.

jrobble (changes requested):
> Reviewed 8 of 11 files at r3, 10 of 10 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1753)* status: all files reviewed, 3 unresolved discussions (waiting on @brosenberg42)

*[a discussion](https://reviewable.io/reviews/openmpf/openmpf/1753#-Nr6Wu9j5isuiOxQiQDU:-NrqOCaXCx8ktsdMySbB:b8gshhg) (no related file):*

Should we switch over to using the header for markup and remove the timing fields from the protobuf messages if we're not going to use them?

When we handle GPU timing do you expect to use a different header, one for CPU time and one for GPU time? Just thinking that the more complicated the timing information we need to pass back is, the more reason it should be in the protobuf.
</blockquote></details>

I see that you created https://github.com/openmpf/openmpf/issues/1769 for refactoring the protobufs.

As discussed elsewhere, we plan to a separate header for GPU timing when we support that behavior.

___
*[a discussion](https://reviewable.io/reviews/openmpf/openmpf/1753#-NrB3jUOBpTqoSQq621k:-NrqYB6mBR7tsm-cHc3Q:b6w58wf) (no related file):*

Done.
</blockquote></details>

I see the issues:

WFM time: https://github.com/openmpf/openmpf/issues/1766

GPU time: https://github.com/openmpf/openmpf/issues/1767

___
___
*[`trunk/detection/executor/cpp/batch/HealthCheck.h` line 79 at r4](https://reviewable.io/reviews/openmpf/openmpf/1753#-NrqEWez95nlYbTuNrWU:-NrqEWf-0XcIUBRJ5zX7:b-9dbwfs) ([raw file](https://github.com/openmpf/openmpf/blob/d152222ae1f2eaceb698ba4dc495a849de9759cc/trunk/detection/executor/cpp/batch/HealthCheck.h#L79)):*
> ```objc
>     HealthCheckJob health_check_job_;
> 
>     using clock_t = std::chrono::steady_clock;
> ```

Why did you switch to using steady_clock? 

From what I read, that's the preferred approach if you don't need to know the wall clock time, so it makes sense to me. I'm wondering if you have other reasons or if there was a bug.
___
*[`trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/JobCompleteProcessorImpl.java` line 770 at r4](https://reviewable.io/reviews/openmpf/openmpf/1753#-NrqJNYl7X2A5RF0j5zY:-NrqJNYnC22BTNLsv-Q8:b-7l5vje) ([raw file](https://github.com/openmpf/openmpf/blob/d152222ae1f2eaceb698ba4dc495a849de9759cc/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/JobCompleteProcessorImpl.java#L770)):*
> ```Java
>                 // Use distinct to prevent duplicate entries in the unlikely case of the same
>                 // action being used by more than one stage.
>                 .distinct()
> ```

I can think of one possible use case where the same action may be used by more than one stage. That's when testing if face detection can find the face in face regions passed from upstream. This is done to determine if re-detecting the face before re-templating is a viable approach for external systems.

In cases where the same action is used more than once can we add the processing times together?
___
*[`trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/service/TestTiesDbService.java` line 551 at r2](https://reviewable.io/reviews/openmpf/openmpf/1753#-Nr6mnh7-LhLoziW4Jts:-NrqNBCzB9jmbCM84uP_:bkbxdqz) ([raw file](https://github.com/openmpf/openmpf/blob/5cacd4f00bf4738874f6d4999ec0758a241b1fd5/trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/service/TestTiesDbService.java#L551)):*

Done.
</blockquote></details>

`new JsonActionTiming("ACTION", 876)` should be `new JsonActionTiming("ACTION_0", 876)` since the index is always included in the name.

brosenberg42 (commented):
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1753)* status: all files reviewed, 3 unresolved discussions (waiting on @jrobble)

___
*[`trunk/detection/executor/cpp/batch/HealthCheck.h` line 79 at r4](https://reviewable.io/reviews/openmpf/openmpf/1753#-NrqEWez95nlYbTuNrWU:-Ns88wYB0ilaybulKgDa:bwkqa49) ([raw file](https://github.com/openmpf/openmpf/blob/d152222ae1f2eaceb698ba4dc495a849de9759cc/trunk/detection/executor/cpp/batch/HealthCheck.h#L79)):*

Why did you switch to using steady_clock? 

From what I read, that's the preferred approach if you don't need to know the wall clock time, so it makes sense to me. I'm wondering if you have other reasons or if there was a bug.
</blockquote></details>

I was looking at something in std::chrono and realized that we should be using steady_clock. If the machine's system time is changed while the component is running, then the timing of the health checks will be wrong. 
___
*[`trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/JobCompleteProcessorImpl.java` line 770 at r4](https://reviewable.io/reviews/openmpf/openmpf/1753#-NrqJNYl7X2A5RF0j5zY:-Ns8Cr_A5yHcamnnLwZb:bjpd0uo) ([raw file](https://github.com/openmpf/openmpf/blob/d152222ae1f2eaceb698ba4dc495a849de9759cc/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/JobCompleteProcessorImpl.java#L770)):*

I can think of one possible use case where the same action may be used by more than one stage. That's when testing if face detection can find the face in face regions passed from upstream. This is done to determine if re-detecting the face before re-templating is a viable approach for external systems.

In cases where the same action is used more than once can we add the processing times together?
</blockquote></details>

I think that situation qualifies as "unlikely". In that situation, they would be using the same algorithm multiple times, but it seems unlikely that both stages would have the exact same configuration and thus the same action.

There is nothing to add together here. The times are already currently per action, not per stage. 
___
*[`trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/service/TestTiesDbService.java` line 551 at r2](https://reviewable.io/reviews/openmpf/openmpf/1753#-Nr6mnh7-LhLoziW4Jts:-Ns8HSg85bl2PKKZSY-C:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/5cacd4f00bf4738874f6d4999ec0758a241b1fd5/trunk/workflow-manager/src/test/java/org/mitre/mpf/wfm/service/TestTiesDbService.java#L551)):*

`new JsonActionTiming("ACTION", 876)` should be `new JsonActionTiming("ACTION_0", 876)` since the index is always included in the name.
</blockquote></details>

Done.

jrobble (approved):
> Reviewed 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1753)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @brosenberg42)

___
*[`trunk/detection/executor/cpp/batch/HealthCheck.h` line 79 at r4](https://reviewable.io/reviews/openmpf/openmpf/1753#-NrqEWez95nlYbTuNrWU:-Ns9OZqeAXP2AEDphXnU:b-oxgazt) ([raw file](https://github.com/openmpf/openmpf/blob/d152222ae1f2eaceb698ba4dc495a849de9759cc/trunk/detection/executor/cpp/batch/HealthCheck.h#L79)):*

I was looking at something in std::chrono and realized that we should be using steady_clock. If the machine's system time is changed while the component is running, then the timing of the health checks will be wrong.
</blockquote></details>

Good find.
___
*[`trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/JobCompleteProcessorImpl.java` line 770 at r4](https://reviewable.io/reviews/openmpf/openmpf/1753#-NrqJNYl7X2A5RF0j5zY:-Ns9On59CTjUd5Lz5x_V:bzcs6id) ([raw file](https://github.com/openmpf/openmpf/blob/d152222ae1f2eaceb698ba4dc495a849de9759cc/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/JobCompleteProcessorImpl.java#L770)):*

I think that situation qualifies as "unlikely". In that situation, they would be using the same algorithm multiple times, but it seems unlikely that both stages would have the exact same configuration and thus the same action.

There is nothing to add together here. The times are already currently per action, not per stage.
</blockquote></details>

I was thinking of the case where it is the same action. For example:

```
MOTION --> FACE WITH FEED-FORWARD REGION --> FACE WITH FEED-FORWARD REGION
```

The purpose would be to determine how many faces we failed to detect between stages 2 and 3. A person may run this kind of job to determine the impact of an external system re-templating the face crops from OpenMPF.

I agree this example is an outlier. I'm playing devil's advocate. 

How hard would it be to support this kind of pipeline? If we choose not to, I would feel more comfortable generating a warning rather than silently doing the wrong thing.

jrobble (changes requested):
Accidentally approved.

jrobble (changes requested):
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1753)* status: all files reviewed, 1 unresolved discussion (waiting on @brosenberg42)

jrobble (approved):
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1753)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @brosenberg42)

___
*[`trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/JobCompleteProcessorImpl.java` line 770 at r4](https://reviewable.io/reviews/openmpf/openmpf/1753#-NrqJNYl7X2A5RF0j5zY:-Ns9kobH8Hnk0nVIK9SU:bdpxkxt) ([raw file](https://github.com/openmpf/openmpf/blob/d152222ae1f2eaceb698ba4dc495a849de9759cc/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/JobCompleteProcessorImpl.java#L770)):*

I was thinking of the case where it is the same action. For example:

```
MOTION --> FACE WITH FEED-FORWARD REGION --> FACE WITH FEED-FORWARD REGION
```

The purpose would be to determine how many faces we failed to detect between stages 2 and 3. A person may run this kind of job to determine the impact of an external system re-templating the face crops from OpenMPF.

I agree this example is an outlier. I'm playing devil's advocate. 

How hard would it be to support this kind of pipeline? If we choose not to, I would feel more comfortable generating a warning rather than silently doing the wrong thing.
</blockquote></details>

I initially misunderstood when I read `distinct()`. Under the hood there's just an action name --> time map:

```
    public void addProcessingTime(Action action, long time) {
        _processingTimes.merge(action.name(), time, BatchJobImpl::mergeTimes);
    }
```

That merge is invoked in the following situations:
- Multiple video segments for the same action
- Multiple feed-forward tracks for the same feed-forward action
- Action appears in multiple stages of the pipeline

It's fine to have one entry in the JSON output object for each action. That one entry may represent the same action executed in more than one stage.

Issue #1718, opened 2023-11-07
Support a `QUALITY_SELECTION_PROP` to specify how the WFM should choose an exemplar


jrobble:
For example, `QUALITY_SELECTION_PROP` can be set to `DESCRIPTOR_MAGNITUDE` for components that generate detections with a `DESCRIPTOR_MAGNITUDE` detection property. The WFM will then use the value of that property to select the exemplar detection from a track. By default, `QUALITY_SELECTION_PROP` is set to `CONFIDENCE`. The WFM will first check if a detection has a `CONFIDENCE` entry in the detection properties map, and if not, use the `confidence` field. (This is similar to how `MARKUP_LABELS_NUMERIC_PROP_TO_SHOW` works).

`FEED_FORWARD_TOP_CONFIDENCE_COUNT` should be renamed to `FEED_FORWARD_TOP_QUALITY_COUNT`. Additionally, `ARTIFACT_EXTRACTION_POLICY_TOP_CONFIDENCE_COUNT` should be renamed to `ARTIFACT_EXTRACTION_POLICY_TOP_QUALITY_COUNT`. Both of those properties should use the `QUALITY_SELECTION_PROP` to determine which detections to use.

Additionally, `CONFIDENCE_THRESHOLD`, and the associated system property, should be removed from the WFM. In general, setting one confidence threshold value for different components doesn't make a lot of sense. Each algorithm may have a different range or tolerance. This is why the system property is almost always set to a value that disables it. In most deployments, the only time the WFM performs confidence filtering is when a component sets `CONFIDENCE_THRESHOLD` to a value in its descriptor file, or it's set as a job property for that component.

Moving forward, a `QUALITY_THRESHOLD` property should be used instead of `CONFIDENCE_THRESHOLD`. That will be used by the WFM to filter out detections when the value for the `QUALITY_SELECTION_PROP` does not meet that threshold.

Update each component to either:

- Remove `CONFIDENCE_THRESHOLD` from the descriptor if set to a value that effectively disables it.
- Leave `CONFIDENCE_THRESHOLD` in the descriptor if the component itself implements support for it.
- Change `CONFIDENCE_THRESHOLD` to `QUALITY_SELECTION_PROP` and `QUALITY_THRESHOLD` in the descriptor if the component is currently using `CONFIDENCE_THRESHOLD` and relying on the WFM to filter out low-confidence detections.

PR #1760, opened 2024-01-30
Add ability to use a detection property for quality selection

clnowacki:
**Issues:**
- https://github.com/openmpf/openmpf/issues/1718

**Related PRs**
- https://github.com/openmpf/openmpf-components/pull/350
- https://github.com/openmpf/openmpf-docker/pull/209

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1760)


clnowacki (commented):
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1760)* status: 16 of 42 files reviewed, 9 unresolved discussions (waiting on @brosenberg42 and @jrobble)

___
*[`trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/DetectionResponseProcessor.java` line 207 at r3](https://reviewable.io/reviews/openmpf/openmpf/1760#-NraDg3s2PbHHTXxPTZZ:-NrgN1ue-rEi-v6O6qmX:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/24adeeda9852d2271ed340e45ba7e44a78f0ebdb/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/DetectionResponseProcessor.java#L207)):*

We had at one point discussed how to handle tracks that didn't have the quality property at the track level, but did have it at the detection level. I believe we decided that if a track is missing the quality property we should check if any detection has the quality property and use the maximum detection value when considering the threshold. When I gave you the `QualityFilter` code, I forgot about that.

I think we can get that behavior using the code below. Originally, we said that the track would be accepted if the maximum detection value is over the threshold. Checking if detections is empty provides the same behavior because the detections below the threshold get `.filter()`ed out. 

```java
if (!qualityFilter.meetsThreshold(objectTrack.getConfidence(), trackProperties)
        && ("CONFIDENCE".equalsIgnoreCase(qualitySelectionProp)
                || trackProperties.containsKey(qualitySelectionProp))) {
    continue;
}
var detections = objectTrack.getFrameLocationsList()
        .stream()
        .map(flm -> toDetection(flm, frameTimeInfo))
        .filter(d -> qualityFilter.meetsThreshold(d.getConfidence(), d.getDetectionProperties()))
        .collect(ImmutableSortedSet.toImmutableSortedSet(Comparator.naturalOrder()));
if (!detections.isEmpty()) {
    var track = new Track(...);
    _inProgressJobs.addTrack(track);
}
```
</blockquote></details>

Done.
___
*[`trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/DetectionResponseProcessor.java` line 230 at r3](https://reviewable.io/reviews/openmpf/openmpf/1760#-Nr_Yj5f0faQfo-qod8P:-NrgN2fmEJ6tttbS0fLv:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/24adeeda9852d2271ed340e45ba7e44a78f0ebdb/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/detection/DetectionResponseProcessor.java#L230)):*

There are a few places in this file where `(float)` was added before `objectTrack.getConfidence()`, but they are not necessary because `objectTrack.getConfidence()` is already a float.
</blockquote></details>

Done.
___
*[`trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/persistent/SystemPropertiesSnapshot.java` line 78 at r1](https://reviewable.io/reviews/openmpf/openmpf/1760#-Npt45Am46Lwoql_rmag:-NrgNKlH4RbWdoaiEaTp:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/9bd8b88837434dd99af6bfe3981e7c2cc0b6a69a/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/data/entities/persistent/SystemPropertiesSnapshot.java#L78)):*

This is never used.
</blockquote></details>

Done.
___
*[`trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/PropertiesUtil.java` line 462 at r1](https://reviewable.io/reviews/openmpf/openmpf/1760#-Npt69J96Ww3v4okwlKb:-NrgN_4eEACVkliDmBvO:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/9bd8b88837434dd99af6bfe3981e7c2cc0b6a69a/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/PropertiesUtil.java#L462)):*

This is never used.
</blockquote></details>

Done.
___
*[`trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/TopQualitySelectionUtil.java` line 54 at r1](https://reviewable.io/reviews/openmpf/openmpf/1760#-NpuWG8-BovRT_uL5lSS:-NrgQ46K3adaSszO0eYk:b-ai0kmw) ([raw file](https://github.com/openmpf/openmpf/blob/9bd8b88837434dd99af6bfe3981e7c2cc0b6a69a/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/TopQualitySelectionUtil.java#L54)):*

When `FEED_FORWARD_TOP_QUALITY_COUNT` is <= 0, that means we should not filter by quality, so all detections should be passed to the next stage.
</blockquote></details>

True, but if `FEED_FORWARD_TOP_QUALITY_COUNT` is <=0 in VideoMediaSegmenter, then this function doesn't get called. Do we really need to check for that case at all?
___
*[`trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/TopQualitySelectionUtil.java` line 83 at r1](https://reviewable.io/reviews/openmpf/openmpf/1760#-NpuWdr5239YS7pbj7VP:-NrgQcXK2_gR9pjvPdKk:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/9bd8b88837434dd99af6bfe3981e7c2cc0b6a69a/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/TopQualitySelectionUtil.java#L83)):*

Please change `qualityProperty.toLowerCase().equals("confidence")` to `qualityProperty.equalsIgnoreCase("confidence")`
</blockquote></details>

Done.
___
*[`trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/TopQualitySelectionUtil.java` line 86 at r3](https://reviewable.io/reviews/openmpf/openmpf/1760#-Nr_VXOaAzX6CVGqv3td:-NrgS6mFEf_mBWIndK-A:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/24adeeda9852d2271ed340e45ba7e44a78f0ebdb/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/TopQualitySelectionUtil.java#L86)):*

We should check if `det.getDetectionProperties().get(qualityProperty)` is null and if it is, return `Double.NEGATIVE_INFINITY`. When the parameter to `Double.parseDouble` is null, a `NullPointerException` is thrown.
</blockquote></details>

Done.
___
*[`trunk/workflow-manager/src/main/resources/workflow-properties.json` line 117 at r3](https://reviewable.io/reviews/openmpf/openmpf/1760#-Nr_kf6pAMb7ecy3EDkP:-NrgSXZhAV5ndkUySi7d:b-jxiip5) ([raw file](https://github.com/openmpf/openmpf/blob/24adeeda9852d2271ed340e45ba7e44a78f0ebdb/trunk/workflow-manager/src/main/resources/workflow-properties.json#L117)):*

The type should be `"DOUBLE"`
</blockquote></details>

But in mpf.properties, the default value is '-Infinity'. Doesn't that mean it has to be a string?
___
*[`trunk/workflow-manager/src/main/resources/properties/mpf.properties` line 157 at r3](https://reviewable.io/reviews/openmpf/openmpf/1760#-Nr_aHNL9Ve3pKJqljNP:-NrgTk8g8abt-x8mcRP2:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/24adeeda9852d2271ed340e45ba7e44a78f0ebdb/trunk/workflow-manager/src/main/resources/properties/mpf.properties#L157)):*

Instead of "must represent a weak ordering", I think we should say "must be integer or floating point values".
</blockquote></details>

Done.

brosenberg42 (changes requested):
> Reviewed 5 of 39 files at r1, 7 of 10 files at r2, 5 of 5 files at r3, 5 of 5 files at r4, 5 of 5 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1760)* status: all files reviewed, 2 unresolved discussions (waiting on @clnowacki and @jrobble)

___
*[`trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/TopQualitySelectionUtil.java` line 54 at r1](https://reviewable.io/reviews/openmpf/openmpf/1760#-NpuWG8-BovRT_uL5lSS:-NrkPS5C56Rf3hTHu-jU:b-dfse59) ([raw file](https://github.com/openmpf/openmpf/blob/9bd8b88837434dd99af6bfe3981e7c2cc0b6a69a/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/TopQualitySelectionUtil.java#L54)):*
<details><summary><i>Previously, clnowacki wrote…</i></summary><blockquote>

True, but if `FEED_FORWARD_TOP_QUALITY_COUNT` is <=0 in VideoMediaSegmenter, then this function doesn't get called. Do we really need to check for that case at all?
</blockquote></details>

I think it would be a good idea because a public utility function should not be dependent on an implementation detail of its callers. Previously, TopConfidenceUtil did return the whole collection when `topConfidenceCount <= 0`. In that case, the check in VideoMediaSegmenter is not needed. It is just an optimization.
___
*[`trunk/workflow-manager/src/main/resources/workflow-properties.json` line 117 at r3](https://reviewable.io/reviews/openmpf/openmpf/1760#-Nr_kf6pAMb7ecy3EDkP:-NrjosI1B7yWxUgffGEe:bi3w52s) ([raw file](https://github.com/openmpf/openmpf/blob/24adeeda9852d2271ed340e45ba7e44a78f0ebdb/trunk/workflow-manager/src/main/resources/workflow-properties.json#L117)):*
<details><summary><i>Previously, clnowacki wrote…</i></summary><blockquote>

But in mpf.properties, the default value is '-Infinity'. Doesn't that mean it has to be a string?
</blockquote></details>

"-Infinity" is considered a valid double. `PipelineValidator` calls `Double.parseDouble` to validate doubles. `Double.parseDouble("-Infinity")` returns `Double.NEGATIVE_INFINITY`

clnowacki (commented):
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1760)* status: 41 of 42 files reviewed, 2 unresolved discussions (waiting on @brosenberg42 and @jrobble)

___
*[`trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/TopQualitySelectionUtil.java` line 54 at r1](https://reviewable.io/reviews/openmpf/openmpf/1760#-NpuWG8-BovRT_uL5lSS:-NrkpLfb2bwM9U8aQoJ4:bovvf7q) ([raw file](https://github.com/openmpf/openmpf/blob/9bd8b88837434dd99af6bfe3981e7c2cc0b6a69a/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/TopQualitySelectionUtil.java#L54)):*

I think it would be a good idea because a public utility function should not be dependent on an implementation detail of its callers. Previously, TopConfidenceUtil did return the whole collection when `topConfidenceCount <= 0`. In that case, the check in VideoMediaSegmenter is not needed. It is just an optimization.
</blockquote></details>

OK, and I understand about not changing the previous behavior. But isn't it counter-intuitive for a utility to return the opposite of what is requested? How does returning the entire collection when the request was for 0 detections make sense, outside of the context of the definition of `FEED_FORWARD_TOP_QUALITY_COUNT`?
___
*[`trunk/workflow-manager/src/main/resources/workflow-properties.json` line 117 at r3](https://reviewable.io/reviews/openmpf/openmpf/1760#-Nr_kf6pAMb7ecy3EDkP:-Nrl1IZF2dFtZX7DTRZ0:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/24adeeda9852d2271ed340e45ba7e44a78f0ebdb/trunk/workflow-manager/src/main/resources/workflow-properties.json#L117)):*

"-Infinity" is considered a valid double. `PipelineValidator` calls `Double.parseDouble` to validate doubles. `Double.parseDouble("-Infinity")` returns `Double.NEGATIVE_INFINITY`
</blockquote></details>

Done.

brosenberg42 (commented):
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1760)* status: 41 of 42 files reviewed, 1 unresolved discussion (waiting on @clnowacki and @jrobble)

___
*[`trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/TopQualitySelectionUtil.java` line 54 at r1](https://reviewable.io/reviews/openmpf/openmpf/1760#-NpuWG8-BovRT_uL5lSS:-NrlEyMI55CzzZdsvMRM:b-l1780r) ([raw file](https://github.com/openmpf/openmpf/blob/9bd8b88837434dd99af6bfe3981e7c2cc0b6a69a/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/TopQualitySelectionUtil.java#L54)):*
<details><summary><i>Previously, clnowacki wrote…</i></summary><blockquote>

OK, and I understand about not changing the previous behavior. But isn't it counter-intuitive for a utility to return the opposite of what is requested? How does returning the entire collection when the request was for 0 detections make sense, outside of the context of the definition of `FEED_FORWARD_TOP_QUALITY_COUNT`?
</blockquote></details>

My original thinking was that when we are getting the top N values, only positive integers for N make sense. When a certain range of numbers doesn't make sense for a job property, we typically use the invalid values to mean the feature is disabled. We have no way to represent the idea of a property not being set because we decided that all properties have a default value.

That reason does require the context of `FEED_FORWARD_TOP_QUALITY_COUNT`. I think throwing an exception when the value is less than 1 would be more correct because nothing really makes sense for negative numbers.

brosenberg42 (approved):
> Reviewed 1 of 1 files at r6, 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1760)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jrobble)

Issue #1786, opened 2024-03-29
Modify the behavior of Markup when CONFIDENCE is the bounding box label to be displayed.


clnowacki:
Currently, if the property to be used to label the markup bounding box is set to `CONFIDENCE`, the markup code will look first to see if there is a track or detection property with that name. If not, it will fall back to using the `confidence` member of the detection object. This is inconsistent with the way that the Top QualitySelectionUtil deals with a value of `CONFIDENCE` for the quality selection property, which will only look at the `confidence` member of the detection object. To be consistent, we should modify the markup code to only look at the `confidence` member of the detection object.

This change does not appear to require a documentation update as the description of `MARKUP_LABELS_NUMERIC_PROP_TO_SHOW` in the Markup Guide and [Algorithms.json](https://github.com/openmpf/openmpf/blob/1b2f98a9fc5f0a68043ab72d619afb03e54cc079/trunk/workflow-manager/src/main/resources/templates/Algorithms.json#L52) are sufficient.

PR #1789, opened 2024-04-04
Modify the behavior of Markup when CONFIDENCE is the bounding box label to be displayed.

clnowacki:
**Issues:**
- https://github.com/openmpf/openmpf/issues/1786

### DIFF
--- a/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/markup/MarkupSplitter.java
+++ b/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/markup/MarkupSplitter.java
@@ -47,6 +47,7 @@ import javax.inject.Inject;
 import org.apache.camel.CamelContext;
 import org.apache.camel.Message;
 import org.apache.camel.impl.DefaultMessage;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.tika.mime.MimeTypes;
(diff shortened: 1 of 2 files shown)


jrobble (approved):
> Reviewed 2 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1789)* status: all files reviewed, 1 unresolved discussion (waiting on @clnowacki)

___
*[`trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/markup/MarkupSplitter.java` line 534 at r1](https://reviewable.io/reviews/openmpf/openmpf/1789#-NueiDvQ0bH0swNeULbQ:-NueiDvQ0bH0swNeULbR:b-w1ao4p) ([raw file](https://github.com/openmpf/openmpf/blob/1e037b0ca98cf62c705eb81b0f07338359ecca87/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/camel/operations/markup/MarkupSplitter.java#L534)):*
> ```Java
>             if (numericProp.equalsIgnoreCase("CONFIDENCE")) {
>                 numericStr = Float.toString(track.getConfidence());
>                 if (numericStr == null) {
> ```

`track.getConfidence()` will return a `float` primitive. I don't think it's possible for `Float.toString()` to return a `null` when processing that.

Issue #1819, opened 2024-06-07
Workflow Manager splitterThreadPoolProfile does not properly re-use threads.


brosenberg42:
Currently, splitterThreadPoolProfile starts new threads even when there are idle threads. The only time a thread would be used more than once is if the thread pool was full and there were queued tasks. It should be changed to only create new threads when there are no idle threads.

---

When WFM is under heavy load, the threading behavior is what we want. Under lighter loads, it creates new threads rather than re-using them. When the system is under lighter load, everything else is faster, so the user is unlikely to notice the issue.

PR #1820, opened 2024-06-07
Fix splitterThreadPoolProfile

brosenberg42:
**Issues:**

- https://github.com/openmpf/openmpf/issues/1819

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1820)

### DIFF
--- a/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/ActiveMQConfiguration.java
+++ b/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/ActiveMQConfiguration.java
@@ -26,9 +26,12 @@
 
 package org.mitre.mpf.wfm;
 
+import java.util.concurrent.TimeUnit;
+
 import org.apache.activemq.broker.BrokerService;
 import org.apache.activemq.broker.region.policy.PolicyEntry;
 import org.apache.activemq.broker.region.policy.PolicyMap;
+import org.apache.camel.ThreadPoolRejectedPolicy;
 import org.apache.camel.builder.ThreadPoolProfileBuilder;
 import org.apache.camel.spi.ThreadPoolProfile;
 import org.mitre.mpf.wfm.util.PropertiesUtil;
@@ -85,10 +88,15 @@ public class ActiveMQConfiguration {
         // The default thread pool profile for splits with parallelProcessing only allows for 10
         // parallel threads.
         return new ThreadPoolProfileBuilder(SPLITTER_THREAD_POOL_REF)
-            .poolSize(propertiesUtil.getAmqConcurrentConsumers())
+            // Always have at least 1 thread in the pool.
+            .poolSize(1)
             .maxPoolSize(propertiesUtil.getAmqConcurrentConsumers())
-            .allowCoreThreadTimeOut(true)
+            .keepAliveTime(1L, TimeUnit.MINUTES)
+            // A queue is not needed because the of the ThreadPoolRejectedPolicy.CallerRuns policy.
+            .maxQueueSize(0)
+            // When the thread pool is at the maximum size and all threads in the pool are busy,
+            // the task will run on the calling thread rather than on a pool thread.
+            .rejectedPolicy(ThreadPoolRejectedPolicy.CallerRuns)
             .build();
     }
-
 }
--- a/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/ActiveMQConfiguration.java
+++ b/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/ActiveMQConfiguration.java
@@ -104,10 +104,14 @@ public class ActiveMQConfiguration {
             .poolSize(1)
             .maxPoolSize(propertiesUtil.getAmqConcurrentConsumers())
             .keepAliveTime(1L, TimeUnit.MINUTES)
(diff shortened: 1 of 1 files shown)


jrobble (changes requested):
> Reviewed 1 of 1 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1820)* status: all files reviewed, 2 unresolved discussions (waiting on @brosenberg42)

___
*[`trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/ActiveMQConfiguration.java` line 96 at r1](https://reviewable.io/reviews/openmpf/openmpf/1820#-O-2G3dI2ImfC6Rj5_dS:-O-2G3dI2ImfC6Rj5_dT:b-fz6p4g) ([raw file](https://github.com/openmpf/openmpf/blob/b128a49647a8c7b4ae8586fcab755c75f3d4ccda/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/ActiveMQConfiguration.java#L96)):*
> ```Java
>             .keepAliveTime(1L, TimeUnit.MINUTES)
>             // A queue is not needed because the of the ThreadPoolRejectedPolicy.CallerRuns policy.
>             .maxQueueSize(0)
> ```

Please mention in a comment our motivation for doing this. Specifically, if you don't set this then the behavior will prefer creating new threads over re-using existing threads.

Also, I believe you said that it's better for the calling thread to block since that way we won't accept new REST requests / messages until clearing out some of the existing tasks.
___
*[`trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/ActiveMQConfiguration.java` line 99 at r1](https://reviewable.io/reviews/openmpf/openmpf/1820#-O-2GNopEjdXnc_m_wVb:-O-2GNopEjdXnc_m_wVc:b-von8xg) ([raw file](https://github.com/openmpf/openmpf/blob/b128a49647a8c7b4ae8586fcab755c75f3d4ccda/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/ActiveMQConfiguration.java#L99)):*
> ```Java
>             // When the thread pool is at the maximum size and all threads in the pool are busy,
>             // the task will run on the calling thread rather than on a pool thread.
>             .rejectedPolicy(ThreadPoolRejectedPolicy.CallerRuns)
> ```

From what I read [here](https://camel.apache.org/manual/threading-model.html) (not sure if that's the right version though), these are the default settings:

```
.keepAliveTime(1L, TimeUnit.MINUTES)
.allowCoreThreadTimeOut(true)
.rejectedPolicy(ThreadPoolRejectedPolicy.CallerRuns)
```

I noticed that you removed `.allowCoreThreadTimeOut(true)`, but set the other two. Not saying you need to change anything, but are you setting these just to be sure they don't change in a future version?

brosenberg42 (commented):
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1820)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @jrobble)

___
*[`trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/ActiveMQConfiguration.java` line 96 at r1](https://reviewable.io/reviews/openmpf/openmpf/1820#-O-2G3dI2ImfC6Rj5_dS:-O-6RZxA64TkFtgKiAmn:b-51gv9g) ([raw file](https://github.com/openmpf/openmpf/blob/b128a49647a8c7b4ae8586fcab755c75f3d4ccda/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/ActiveMQConfiguration.java#L96)):*

Please mention in a comment our motivation for doing this. Specifically, if you don't set this then the behavior will prefer creating new threads over re-using existing threads.

Also, I believe you said that it's better for the calling thread to block since that way we won't accept new REST requests / messages until clearing out some of the existing tasks.
</blockquote></details>

The calling thread doesn't block. It runs the task.
___
*[`trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/ActiveMQConfiguration.java` line 99 at r1](https://reviewable.io/reviews/openmpf/openmpf/1820#-O-2GNopEjdXnc_m_wVb:-O-5lae68ivXKJ0XvZsC:b4u63n2) ([raw file](https://github.com/openmpf/openmpf/blob/b128a49647a8c7b4ae8586fcab755c75f3d4ccda/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/ActiveMQConfiguration.java#L99)):*

From what I read [here](https://camel.apache.org/manual/threading-model.html) (not sure if that's the right version though), these are the default settings:

```
.keepAliveTime(1L, TimeUnit.MINUTES)
.allowCoreThreadTimeOut(true)
.rejectedPolicy(ThreadPoolRejectedPolicy.CallerRuns)
```

I noticed that you removed `.allowCoreThreadTimeOut(true)`, but set the other two. Not saying you need to change anything, but are you setting these just to be sure they don't change in a future version?
</blockquote></details>

The actual reason is that I know what settings I need so I set them explicitly. You bring up another good reason. Specifically, you aren't sure what the actual settings are even after looking at documentation.

jrobble (approved):
> Reviewed 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1820)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @brosenberg42)

PR #1902, opened 2025-04-23
Feature Logging requirements

kennyEung:
**Issues:**

**Overview:**
Several Changes were made
- Created an event record for logging with pertinent data fields
- Created logger class to serialize log events to stdout/file
- Logs successful user login, authenticated user, unsuccessful login

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1902)


jrobble (changes requested):
> Reviewed 11 of 11 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1902)* status: all files reviewed, 11 unresolved discussions (waiting on @kennyEung)

___
*[`trunk/markup/src/main/resources/log4j2.xml` line 31 at r1](https://reviewable.io/reviews/openmpf/openmpf/1902#-ORwI_AWCRYna7P275j3:-ORwI_AWCRYna7P275j4:bx359x0) ([raw file](https://github.com/openmpf/openmpf/blob/18813c8b1290683fb18a3afbfbb654119570f7ac/trunk/markup/src/main/resources/log4j2.xml#L31)):*
You made this same change in `trunk/node-manager/src/main/resources/log4j2.xml`.

I don't think this change does anything. When you see a log message like:

> 2025-06-04 18:36:13,477 INFO [http-nio-8080-exec-4] o.m.m.w.b.i.JobRequestServiceImpl - [Job 1] Job has started and is running at priority 4.

That job id is not added by the log4j layout. Rather, it's part of the `%msg` itself.

Did you notice any change to the log messages after updating the layout?
___
*[`trunk/workflow-manager/src/main/resources/log4j2.xml` line 55 at r1](https://reviewable.io/reviews/openmpf/openmpf/1902#-ORxoQGjAOav-AkMRVAx:-ORxoQGjAOav-AkMRVAy:basdnc3) ([raw file](https://github.com/openmpf/openmpf/blob/18813c8b1290683fb18a3afbfbb654119570f7ac/trunk/workflow-manager/src/main/resources/log4j2.xml#L55)):*
I'm concerned about possible synchronization issues:

- https://stackoverflow.com/questions/4046825/log4j-have-multiple-appenders-write-to-the-same-file-with-one-that-always-logs/31721650#comment22176431_4047350
- https://coderanch.com/t/691268/java/RollingFileAppender-working-multiple-appenders-file

Using an AsyncAppender may be the solution: 

- https://stackoverflow.com/a/31721650
- https://logging.apache.org/log4j/2.12.x/manual/appenders.html#AsyncAppender

However, I'm not so sure this is a problem. From https://logging.apache.org/log4j/2.12.x/manual/appenders.html#RollingRandomAccessFileAppender

> For example, two web applications in a servlet container can have their own configuration and safely write to the same file if Log4j is in a ClassLoader that is common to both of them.

Please test rollover and ensure it's working correctly using both appenders. It may work correctly a few times and you won't notice dropped log messages until after many attempts.

@brosenberg42, do you have an opinion on using the appenders in this way?
___
*[a discussion](https://reviewable.io/reviews/openmpf/openmpf/1902#-ORxxKWx0jvOej-nfzjF:-ORxxKWx0jvOej-nfzjG:b-45f94u) (no related file):*
Create system property to turn audit logging on and off. This is a static WFM property like `hibernate.show.sql`. Setting it while the WFM is running will not take effect until the WFM is rebooted.
___
*[a discussion](https://reviewable.io/reviews/openmpf/openmpf/1902#-ORxxYkY099pV2EP4I-9:-ORxxYkY099pV2EP4I-A:b-2jx8d7) (no related file):*
Need to log user logout behavior.

Need to log unsuccessful login behavior.

In general, we should log every time a WFM REST endpoint is hit. This includes the ones exposed via Swagger, and the internal REST endpoints hit from the web UI.

One of our requirements is to log whenever an object is accessed and whenever a file is transferred. This includes the JSON output object, which can be accessed via the web UI by clicking "{}".
___
*[a discussion](https://reviewable.io/reviews/openmpf/openmpf/1902#-ORxyw5h3o7elXXJn7ee:-ORxyw5h3o7elXXJn7ef:b-db6aus) (no related file):*
We are currently only logging hits on the homepage. Technically, navigating to any other view using the top menu page is within the same "page". From the perspective of a DDOS attack, loading any of the views generates the "Homepage accessed" log message, which is sufficient.

The two other pages we support are the Swagger page and hawtio page. Please log when those pages are accessed. For hawtio it may only be possible to log when the user gets redirected there. That's a start at least. Look into it.

Also, you need to log when the user accesses `http://somehost:someport/login`. This is before the user attempts to login. Since they don't even need to login to access that page, it is the most likely target for a DDOS attack. 
___
*[a discussion](https://reviewable.io/reviews/openmpf/openmpf/1902#-ORy6CLz9JQMlC2yGu4l:-ORy6CLz9JQMlC2yGu4m:b-h9c2l0) (no related file):*
For posterity I'm leaving a comment that EDT time shows up as:

```
{"time":"2025-06-04T23:22:18.278-04:00","tag":"&B1E7-FFFF&","app":"openmpf","user":"admin","op":"l","res":"a","msg":"User successfully logged in."}
```

This is acceptable.
___
*[`trunk/workflow-manager/src/main/java/org/mitre/mpf/mvc/controller/ServerMediaController.java` line 228 at r1](https://reviewable.io/reviews/openmpf/openmpf/1902#-ORreWNtDF8KjJLI-xw_:-ORreWNtDF8KjJLI-xwa:b-g9iczt) ([raw file](https://github.com/openmpf/openmpf/blob/18813c8b1290683fb18a3afbfbb654119570f7ac/trunk/workflow-manager/src/main/java/org/mitre/mpf/mvc/controller/ServerMediaController.java#L228)):*
> ```Java
>                             LogAuditEventRecord.OpType.EXTRACT, 
>                             LogAuditEventRecord.ResType.ALLOW, 
>                             "Downloaded media file: jobId=" + jobId + ", uri=" + sourceUri);
> ```

Job id is not provided for "Downloaded markup file", but it is for "Downloaded media file". Job id could be helpful, so let's keep it. The job id is part of the the markup URL, so it's not required to call it out separately, but we should anyway for consistency.

Please log the job id in the audit message whenever it's available. This goes for all audit messages, not just the ones in this controller.
___
*[`trunk/workflow-manager/src/main/java/org/mitre/mpf/mvc/controller/MarkupController.java` line 249 at r1](https://reviewable.io/reviews/openmpf/openmpf/1902#-ORxtWwm077s6RtzqPb_:-ORxtWwm077s6RtzqPba:b-c7xxvk) ([raw file](https://github.com/openmpf/openmpf/blob/18813c8b1290683fb18a3afbfbb654119570f7ac/trunk/workflow-manager/src/main/java/org/mitre/mpf/mvc/controller/MarkupController.java#L249)):*
> ```Java
>                 .orElse(null);
>         if (job == null) {
>             log.error("Markup with id " + id + " download failed. Invalid job with id " +
> ```

Log invalid attempts too. This includes when exceptions are thrown. Use ERROR for the ResType. 

This goes for all of the endpoint behaviors in all of the *Controller classes.
___
*[`trunk/workflow-manager/src/main/java/org/mitre/mpf/mvc/controller/LoginController.java` line 140 at r1](https://reviewable.io/reviews/openmpf/openmpf/1902#-ORxsh0KBtChx9s40w_N:-ORxsh0L3TNYAWGTQI4N:btpdizz) ([raw file](https://github.com/openmpf/openmpf/blob/18813c8b1290683fb18a3afbfbb654119570f7ac/trunk/workflow-manager/src/main/java/org/mitre/mpf/mvc/controller/LoginController.java#L140)):*
> ```Java
>         
>         if (authentication != null && authentication.isAuthenticated()) {
>             auditEventLogger.log(LogAuditEventRecord.TagType.SECURITY, LogAuditEventRecord.OpType.LOGIN, LogAuditEventRecord.ResType.ALLOW, "User is already authenticated.");
> ```

Nothing shows up in the logs if I open two tabs, log in using one, and then log in using the other tab.

Logging in using two browsers doesn't log this either.

What actions do I need to take to generate this log message?
___
*[`trunk/workflow-manager/src/main/java/org/mitre/mpf/mvc/controller/LoginController.java` line 148 at r1](https://reviewable.io/reviews/openmpf/openmpf/1902#-ORxsGJu9KN60BNwLF5c:-ORxsGJu9KN60BNwLF5d:b96p3xo) ([raw file](https://github.com/openmpf/openmpf/blob/18813c8b1290683fb18a3afbfbb654119570f7ac/trunk/workflow-manager/src/main/java/org/mitre/mpf/mvc/controller/LoginController.java#L148)):*
> ```Java
> 
>         if (authException instanceof BadCredentialsException) {
>             auditEventLogger.log(LogAuditEventRecord.TagType.SECURITY, LogAuditEventRecord.OpType.LOGIN, LogAuditEventRecord.ResType.DENY, "Failed login attempt: Bad credentials.");
> ```

This does not appear in the logs when I use the UI and have a failed login attempt. Nothing gets logged for that operation.

What actions to a need to take to generate this log message?

Ensure invalid login attempts get logged and mention the username in the log message.
___
*[`trunk/workflow-manager/src/main/java/org/mitre/mpf/mvc/controller/LoginController.java` line 171 at r1](https://reviewable.io/reviews/openmpf/openmpf/1902#-ORrcqQWBAs1iXol0UqQ:-ORrcqQWBAs1iXol0UqR:b-axs56q) ([raw file](https://github.com/openmpf/openmpf/blob/18813c8b1290683fb18a3afbfbb654119570f7ac/trunk/workflow-manager/src/main/java/org/mitre/mpf/mvc/controller/LoginController.java#L171)):*
> ```Java
>         if (accessDeniedException != null) {
>             log.error(
>                 "A user successfully authenticated with an OIDC provider, but was not" +
> ```

Create a string with the log line and use that for both the error and log line. This will reduce redundancy.
___
*[`trunk/workflow-manager/src/main/java/org/mitre/mpf/mvc/security/LocalSecurityConfig.java` line 78 at r1](https://reviewable.io/reviews/openmpf/openmpf/1902#-ORrgOatAAKuD0YXOl7F:-ORrgOatAAKuD0YXOl7G:boddxwl) ([raw file](https://github.com/openmpf/openmpf/blob/18813c8b1290683fb18a3afbfbb654119570f7ac/trunk/workflow-manager/src/main/java/org/mitre/mpf/mvc/security/LocalSecurityConfig.java#L78)):*
> ```Java
>     public SecurityFilterChain formSecurityFilterChain(
>             HttpSecurity http,
>             CustomAccessDeniedHandler customAccessDeniedHandler,
> ```

Do we need to log in the CustomAccessDeniedHandler to check for invalid user logins?

That class checks for an `INVALID_CSRF_TOKEN`. Logging that would be useful to detect CSRF attacks.

clnowacki (commented):
> Reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1902)* status: 6 of 16 files reviewed, 11 unresolved discussions (waiting on @jrobble and @kennyEung)

___
*[a discussion](https://reviewable.io/reviews/openmpf/openmpf/1902#-ORxxYkY099pV2EP4I-9:-OTmo_JADyFXyDGTJ7Hg:b-aggnpy) (no related file):*

Need to log user logout behavior.

Need to log unsuccessful login behavior.

In general, we should log every time a WFM REST endpoint is hit. This includes the ones exposed via Swagger, and the internal REST endpoints hit from the web UI.

One of our requirements is to log whenever an object is accessed and whenever a file is transferred. This includes the JSON output object, which can be accessed via the web UI by clicking "{}".
</blockquote></details>

Kenny - The first two of these are done, and the REST endpoint logging is a separate task. All that is left to do here is to add logging when accessing the JSON output object. 
___
*[a discussion](https://reviewable.io/reviews/openmpf/openmpf/1902#-ORxyw5h3o7elXXJn7ee:-OTmpRtb7gipD7IVMdjH:b-r6r781) (no related file):*
<details><summary><i>Previously, kennyEung wrote…</i></summary><blockquote>

Done.
</blockquote></details>

I don't see logging when I navigate to the Swagger page from the home page. Is that supposed to be done, or is that waiting for the REST API logging?

clnowacki (commented):
> Reviewed 1 of 11 files at r1, 1 of 3 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1902)* status: 5 of 17 files reviewed, 11 unresolved discussions (waiting on @jrobble and @kennyEung)

___
*[`trunk/workflow-manager/src/main/java/org/mitre/mpf/mvc/controller/LoginController.java` line 148 at r1](https://reviewable.io/reviews/openmpf/openmpf/1902#-ORxsGJu9KN60BNwLF5c:-OViDlpZ7TIBMHBb8p43:bonucy8) ([raw file](https://github.com/openmpf/openmpf/blob/18813c8b1290683fb18a3afbfbb654119570f7ac/trunk/workflow-manager/src/main/java/org/mitre/mpf/mvc/controller/LoginController.java#L148)):*
<details><summary><i>Previously, kennyEung wrote…</i></summary><blockquote>

A simple way to generate this log message is on the Login UI, logging in with empty credentials (empty user, empty pass). It should generate: 

</blockquote></details>

As Jeff suggested, please mention the username in the log message when authentication fails.
___
*[`trunk/workflow-manager/src/main/java/org/mitre/mpf/mvc/controller/MarkupController.java` line 249 at r1](https://reviewable.io/reviews/openmpf/openmpf/1902#-ORxtWwm077s6RtzqPb_:-OViELbLChvu__3GTZhd:b7eybgo) ([raw file](https://github.com/openmpf/openmpf/blob/18813c8b1290683fb18a3afbfbb654119570f7ac/trunk/workflow-manager/src/main/java/org/mitre/mpf/mvc/controller/MarkupController.java#L249)):*

Log invalid attempts too. This includes when exceptions are thrown. Use ERROR for the ResType. 

This goes for all of the endpoint behaviors in all of the *Controller classes.
</blockquote></details>

Has this comment been addressed? I don't see anything in the code, but maybe you discussed this off-line with Jeff?
___
*[`trunk/workflow-manager/src/main/java/org/mitre/mpf/mvc/controller/ServerMediaController.java` line 228 at r1](https://reviewable.io/reviews/openmpf/openmpf/1902#-ORreWNtDF8KjJLI-xw_:-OViH4TFCoGrhnGsfq5m:bdf20fo) ([raw file](https://github.com/openmpf/openmpf/blob/18813c8b1290683fb18a3afbfbb654119570f7ac/trunk/workflow-manager/src/main/java/org/mitre/mpf/mvc/controller/ServerMediaController.java#L228)):*

Job id is not provided for "Downloaded markup file", but it is for "Downloaded media file". Job id could be helpful, so let's keep it. The job id is part of the the markup URL, so it's not required to call it out separately, but we should anyway for consistency.

Please log the job id in the audit message whenever it's available. This goes for all audit messages, not just the ones in this controller.
</blockquote></details>

Please address this comment.

clnowacki (approved):
> Reviewed 1 of 11 files at r1, 1 of 4 files at r2, 3 of 3 files at r4, 1 of 4 files at r5, 1 of 2 files at r6, 2 of 3 files at r7, 1 of 3 files at r8, 8 of 8 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1902)* status: all files reviewed, 11 unresolved discussions (waiting on @jrobble)

PR #53, opened 2017-05-03
Add http properties to mpf-system-tests.properties. Also testing docu…

brosenberg42:
…mentation for #23

---
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/53)

PR #22, opened 2017-02-03
Test commit.

mistr-stoker:
This is a test.

---
This change is [Reviewable](https://reviewable.io/reviews/openmpf/openmpf/22)

Issue #2, opened 2017-01-27
Separate out each component into a subrepo


lewisloren:
Components should not be packaged in the same .tar.gz as the MPF core. To date we've been delivering source code that includes the MPF core source code and the source code for all of the components. Moving forward we should provide the source code for components separately from the MPF core source code. This provides for greater flexibility and allows us to keep the MPF core source code clean and ready to be shared.

jrobble:
This task is no longer valid. Components will now be placed into the openmpf-components repo or the openmpf-contrib-components repo.

PR #1906, opened 2025-05-14
Feat/ff all tracks

jrobble:
**Issues:**

- https://github.com/openmpf/openmpf/issues/1 < Example
- https://github.com/openmpf/openmpf/issues/2 < Example

If a related issue doesn't exist, then create one first and assign it to yourself.

**Related PRs:**

- https://github.com/openmpf/openmpf/pull/22 < Example
- https://github.com/openmpf/openmpf/pull/53 < Example


brosenberg42:
*[`trunk/workflow-manager/src/main/resources/workflow-properties.json` line 14 at r1](https://reviewable.io/reviews/openmpf/openmpf/1906#-OQJmQuz5GNSkdFQ2dXb:-OQJmQuz5GNSkdFQ2dXc:bjze2nq) ([raw file](https://github.com/openmpf/openmpf/blob/3413018c169916c4ebaaa0b9517284fd675eedc3/trunk/workflow-manager/src/main/resources/workflow-properties.json#L14)):*
> ```JSON
>         "type": "BOOLEAN",
>         "defaultValue": "false",
>         "mediaTypes": ["VIDEO", "IMAGE", "AUDIO", "UNKNOWN"]
> ```

This should only be `["VIDEO"]`.

jrobble (commented):
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1906)* status: 0 of 21 files reviewed, 1 unresolved discussion (waiting on @brosenberg42 and @jrobble)

___
*[`trunk/workflow-manager/src/main/resources/workflow-properties.json` line 14 at r1](https://reviewable.io/reviews/openmpf/openmpf/1906#-OQJmQuz5GNSkdFQ2dXb:-O_TQOjs6iQTGNWfr2Mz:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/3413018c169916c4ebaaa0b9517284fd675eedc3/trunk/workflow-manager/src/main/resources/workflow-properties.json#L14)):*

This should only be `["VIDEO"]`.
</blockquote></details>

Done.

jrobble:
*[`trunk/detection/executor/cpp/batch/PythonComponentHandle.cpp` line 573 at r2](https://reviewable.io/reviews/openmpf/openmpf/1906#-O_weKPr2fmjMpGWwCw8:-Oa6yUWW2k1h8zKzot4q:b-wyficm) ([raw file](https://github.com/openmpf/openmpf/blob/78659849767866b7f55803225368a5751f2caa84/trunk/detection/executor/cpp/batch/PythonComponentHandle.cpp#L573)):*

This is logically wrong. `Supports()` can return true even when calling the actual method will throw `MPF_UNSUPPORTED_DATA_TYPE`. One way to address the issue is to change the parameter to `PythonComponentHandle::Supports` and `CppComponentHandle::Supports` to be `JobContext`.
</blockquote></details>

`Supports()` is used here in `mpf_cli_job_runner.py`:

```python
class JobRunner(contextlib.AbstractContextManager):
    def __init__(self,
                 cmd_line_args: argparse.Namespace,
                 env_props: Mapping[str, str],
                 job_stdin: TextIO,
                 component,
                 descriptor):
        with contextlib.ExitStack() as exit_stack:
            self._output_dest = cmd_line_args.output
            exit_stack.push(self._output_dest)

            self._component_handle = component
            self._sdk_module = component.sdk_module

            self._mime_type = self._get_mime_type(
                cmd_line_args.media_type, cmd_line_args.media_path, cmd_line_args.media_metadata)

            self._media_type = self._get_media_type(
                cmd_line_args.media_type, self._mime_type)

            if not self._component_handle.supports(self._media_type):
                raise RuntimeError(
                    f'The component does not support {self._media_type.name.lower()} jobs.')
```

and here in detection executor `main.cpp`:

```c++
    while (true) {
        logger.Info("Waiting for next job.");
        auto job_context = job_receiver.GetJob();
        if (!component.Supports(job_context.job_type)) {
            job_receiver.ReportUnsupportedDataType(job_context);
            continue;
        }
```

I can see how to get the job context from the latter code, but not the former.

Once I have the job context what do I do with it?

jrobble (commented):
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1906)* status: 15 of 23 files reviewed, 8 unresolved discussions (waiting on @brosenberg42 and @jrobble)

___
*[`trunk/detection/executor/cpp/batch/CppComponentHandle.cpp` line 63 at r2](https://reviewable.io/reviews/openmpf/openmpf/1906#-Oa0FLiH31Wdumf-hNYP:-Oa6FJor2-e72ABAd_Xi:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/78659849767866b7f55803225368a5751f2caa84/trunk/detection/executor/cpp/batch/CppComponentHandle.cpp#L63)):*

We could throw an exception here instead of calling in to the component. It would allow us to remove the method from the C++ component interface.
</blockquote></details>

Done.
___
*[`trunk/detection/executor/cpp/batch/ProtobufRequestUtil.cpp` line 118 at r2](https://reviewable.io/reviews/openmpf/openmpf/1906#-O_wU9KE3SOkBcn7CMHh:-Oa6gaGmDw-FW-ZCgt9e:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/78659849767866b7f55803225368a5751f2caa84/trunk/detection/executor/cpp/batch/ProtobufRequestUtil.cpp#L118)):*

Please create a function named `ConvertFeedForwardTrack` and call it from here and from `CreateVideoJob`. It would convert the protobuf track to the C++ track similar to `ConvertFeedForwardLocation`
</blockquote></details>

Done.
___
*[`trunk/detection/executor/cpp/batch/ProtobufRequestUtil.cpp` line 140 at r2](https://reviewable.io/reviews/openmpf/openmpf/1906#-Oa0jWlT877rpArJZize:-Oa6lV5b0lHXFUvbusq7:b-pv2d88) ([raw file](https://github.com/openmpf/openmpf/blob/78659849767866b7f55803225368a5751f2caa84/trunk/detection/executor/cpp/batch/ProtobufRequestUtil.cpp#L140)):*

Shouldn't this job type always have feed forward tracks? Is there a use case where it makes sense to create a `MPFAllVideoTracksJob` with no tracks?
</blockquote></details>

See this discussion: https://reviewable.io/reviews/openmpf/openmpf-python-component-sdk/91#-Oa0icuW7rcYdA2ZnFnn
___
*[`trunk/detection/executor/cpp/batch/PythonComponentHandle.h` line 60 at r2](https://reviewable.io/reviews/openmpf/openmpf/1906#-Oa0c2ID9is8CPkKlXPZ:-Oa7205bBXu1jNy4ttAl:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/78659849767866b7f55803225368a5751f2caa84/trunk/detection/executor/cpp/batch/PythonComponentHandle.h#L60)):*

Please add a test to `test_batch_executor.cpp` for this method. There are unit tests for all of the other job types in that file.
</blockquote></details>

Done.
___
*[`trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/segmenting/DetectionRequest.java` line 38 at r2](https://reviewable.io/reviews/openmpf/openmpf/1906#-O_wMqiIB4Euw9m7-yRW:-Oa6hGi625svxT2bBEbv:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/78659849767866b7f55803225368a5751f2caa84/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/segmenting/DetectionRequest.java#L38)):*

Normally, Optional is not used with collections because, in most cases, you can just use an empty collection. In this class, when there are no headers, we use `Map.of()`. We should handle `feedForwardTracks` similarly and use `List.of()` when there are no tracks.
</blockquote></details>

Done.
___
*[`trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/segmenting/VideoMediaSegmenter.java` line 162 at r2](https://reviewable.io/reviews/openmpf/openmpf/1906#-Oa0L2yS1sleN4Vqlskm:-Oa6urRA8UqzzTRbRUWF:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/78659849767866b7f55803225368a5751f2caa84/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/segmenting/VideoMediaSegmenter.java#L162)):*

The A at the beginning of the variable name should be lower case.
</blockquote></details>

Done.
___
*[`trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/service/TaskMergingManager.java` line 87 at r2](https://reviewable.io/reviews/openmpf/openmpf/1906#-O_wOq8Z8mTrwUP6IxPh:-Oa6lIDlBkU3X42v1gwt:b-896fix) ([raw file](https://github.com/openmpf/openmpf/blob/78659849767866b7f55803225368a5751f2caa84/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/service/TaskMergingManager.java#L87)):*

In a comment on DetectionRequest.java, I suggested changing the `Optional<List<Track>>` field to `List<Track>`. If we make that change, we might need to handle an empty list here.
</blockquote></details>

Done.

jrobble:
*[`trunk/detection/executor/cpp/batch/PythonComponentHandle.cpp` line 573 at r2](https://reviewable.io/reviews/openmpf/openmpf/1906#-O_weKPr2fmjMpGWwCw8:-OcJS22u1JzSNnvuQWJ4:b-e9ldjx) ([raw file](https://github.com/openmpf/openmpf/blob/78659849767866b7f55803225368a5751f2caa84/trunk/detection/executor/cpp/batch/PythonComponentHandle.cpp#L573)):*

> Overall, I'm lost. What are we trying to achieve?

If you try to run an all tracks job on a component that only supports regular video jobs, `Supports` will return true. We didn't add a new data type for multi-track jobs, so we need more than just the data type to determine whether or not the job type is supported.  `mpf_cpp_runner.CppComponentHandle` is a Python class. It is meant to be the CLI runner's version of the batch executor's `CppComponentHandle.cpp`. 

This is how I thought you would implement Supports with the change:
```cpp
bool Supports(const JobContext& job_context) const {
    if (std::holds_alternative<MPFVideoJob>(job_context.job)) {
        return component_.get_detections_from_video_method.has_value();
    }
    else if (std::holds_alternative<MPFAllVideoTracksJob>(job_context.job)) {
        return component_.get_detections_from_all_video_tracks_method.has_value();
    }
    else if (std::holds_alternative<MPFImageJob>(job_context.job)) {
        return component_.get_detections_from_image_method.has_value();
    }
    else if (std::holds_alternative<MPFAudioJob>(job_context.job)) {
        return component_.get_detections_from_audio_method.has_value();
    }
    else {
        return component_.get_detections_from_generic_method.has_value();
    }
}
```
</blockquote></details>

I implemented this and pushed the changes. There is an issue with this new code:

```c++
while (true) {
        logger.Info("Waiting for next job.");
        auto job_context = job_receiver.GetJob();
        if (!component.Supports(job_context)) {
            job_receiver.ReportUnsupportedDataType(job_context);
            continue;
        }
```

We can no longer assume that `Supports()` fails due to a data type issue, so we can't call `ReportUnsupportedDataType()`. I think we need to generate an exception inside of `Supports()` with the error message. If that's the case, should we rename the call to something like "validates"?

More generally, in the future, I don't see how it's possible to determine if C++ components support the all tracks method since the C++ components have to completely implement the API. We can't check if a method is missing from the implementation with `*_method.has_value()`. 

Is that a problem for C++ implementations or is the default behavior of generating this exception acceptable?

```c++
throw MPFDetectionException(MPFDetectionError::MPF_UNSUPPORTED_DATA_TYPE);
```

Looking at that, again it assumes an unsupported data type when it may just be that the all tracks method isn't implemented. Seems like we would need a more generic NOT_IMPLEMENTED exception.

jrobble (commented):
> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1906)* status: 17 of 27 files reviewed, all discussions resolved (waiting on @brosenberg42 and @jrobble)

___
*[`trunk/detection/executor/cpp/batch/PythonComponentHandle.cpp` line 573 at r2](https://reviewable.io/reviews/openmpf/openmpf/1906#-O_weKPr2fmjMpGWwCw8:-OcOFM5H2gBuxPiklsvx:b-otk2bz) ([raw file](https://github.com/openmpf/openmpf/blob/78659849767866b7f55803225368a5751f2caa84/trunk/detection/executor/cpp/batch/PythonComponentHandle.cpp#L573)):*

I was considering the case to be an unsupported **job** type. I see your point about it not really being an unsupported **data** type. In the code you posted, you can see we handle unsupported data type differently than other errors. I originally thought we should handle the issue in the same way. I see that the error messages are going to be confusing. I think you can just revert that change, sorry for the confusion.

As far as future changes go, I think we should remove `Supports` and rely on `GetDetections` throwing `MPFDetectionException(MPFDetectionError::MPF_UNSUPPORTED_DATA_TYPE)` to get the existing behavior. It can use a different error code if the problem is that it does not support multi-track jobs. To do that, we would also need to refactor the component executor and update the C++ components
</blockquote></details>

I backed those out.
___
*[`trunk/detection/executor/cpp/batch/ProtobufResponseUtil.cpp` line 46 at r6](https://reviewable.io/reviews/openmpf/openmpf/1906#-OcOFUP35fMnsfTBKpoY:-OcOFUP42b4SIsnT4rWK:bool54p) ([raw file](https://github.com/openmpf/openmpf/blob/a15ca493b1fba8619f564e0486dc0240e28236cd/trunk/detection/executor/cpp/batch/ProtobufResponseUtil.cpp#L46)):*
> ```cpp
>             case MPFDetectionDataType::VIDEO: {
>                 auto video_response = detection_response.mutable_video_response();
>                 if (std::holds_alternative<MPFVideoJob>(context.job)) {
> ```

I ran an EAST pipeline with feed forward all tracks, knowing that EAST doesn't support that method. I did this to test to make sure we get a data type not supported error. I was getting a DEAD LETTER error until I made this change.

Issue #1719, opened 2023-11-07
Remove `OUTPUT_LAST_TASK_ONLY` and support `IS_ANNOTATOR` and `SUPPRESS_TRACKS` WFM properties


jrobble:
The `OUTPUT_LAST_TASK_ONLY` property is confusing when used with pipelines that support `TRIGGER`s. Consider this pipeline where the STT stages trigger on specific languages:

```
LANGUAGE DETECTION --> ENGLISH SPEECH-TO-TEXT --> SPANISH SPEECH-TO-TEXT
```

If `OUTPUT_LAST_TASK_ONLY` is set to true, does that mean the user only wants Spanish STT tracks? They probably also want English STT tracks. But what about the situation where the first LDD stage also generates a Russian track? Should that appear in the JSON output object? If so, should that be the only track that appears in the `LANGUAGE DETECTION` section in the JSON output? Should `LANGUAGE DETECTION` also appear in the `TRACKS SUPPRESSED` section of the JSON output because some of the LDD tracks (English and Spanish) are suppressed because they not the last task (i.e. they trigger stages downstream)?

In general, when dealing with stages that have `TRIGGER`s, the last task is no longer predictable. The last task for a track depends on what that track triggers. It gets complicated.

---

Now consider this pipeline where `OUTPUT_LAST_TASK_ONLY` is set to false:

```
LANGUAGE DETECTION --> ENGLISH SPEECH-TO-TEXT --> SPANISH SPEECH-TO-TEXT --> KEYWORD TAGGING
```

In this pipeline, the `KEYWORD TAGGING` stage is configured with `OUTPUT_MERGE_WITH_PREVIOUS_TASK` set to true. This means that the `KEYWORD TAGGING` stage will appear in the `TRACKS MERGED` section of the JSON output. However, again consider the Russian LDD track. Since that is not triggered by either STT stage, that will feed directly into the `KEYWORD TAGGING` stage, so for that track the previous stage is the LDD, whereas for the English and Spanish tracks the previous stage is the respective STT stage. Does this mean that `LANGUAGE DETECTION` should also appear in the `TRACKS MERGED` section of the JSON output? This is confusing.

---

To simplify these situations, let's implement support for the following two properties:

- `IS_ANNOTATOR`: Functionally, this works the same as `OUTPUT_MERGE_WITH_PREVIOUS_TASK`, but is intent is more clear.            
[...]

PR #1913, opened 2025-05-30
Add support for IS_ANNOTATORS and SUPPRESS_TRACKS Properties

jtat1062:
**Issues:**

- https://github.com/openmpf/openmpf/issues/1719


brosenberg42 (changes requested):
A big part of this task was removing the concept of task merging from Workflow Manager and replacing it with the `IS_ANNOTATOR` algorithm property and the `annotators` output field. Please remove `OUTPUT_MERGE_WITH_PREVIOUS_TASK` and any reference to task merging.

> *[Reviewable](https://reviewable.io/reviews/openmpf/openmpf/1913#-:-OUAGe8JA79C4NzfvunK:b-kvx8v8)* status: 0 of 15 files reviewed, 1 unresolved discussion (waiting on @jtat1062)

___
*[`trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/AggregateJobPropertiesUtil.java` line 556 at r1](https://reviewable.io/reviews/openmpf/openmpf/1913#-OUAI8H54TIsZPuTlIZx:-OUAI8H6DYV_RmPk6_0c:b-6uu11f) ([raw file](https://github.com/openmpf/openmpf/blob/c8e2ac5dedc05059e69d839e54f55e130b5f2ae0/trunk/workflow-manager/src/main/java/org/mitre/mpf/wfm/util/AggregateJobPropertiesUtil.java#L556)):*
> ```Java
>     }
> 
>     public boolean isSuppressTrack(Media media, BatchJob job) {
> ```

It does not make sense to set `SUPPRESS_TRACKS` for the entire pipeline. When `SUPPRESS_TRACKS` is set on an action or algorithm, only the tracks for that specific action or algorithm should be suppressed. If someone where to set `SUPPRESS_TRACKS` at the media or job level, that would mean it should suppress output from all stages of the pipeline.

